PR #31, opened 2014-09-27
add spread_ method for data.table

matthieugomez:
`gather_`  works directly on data.table since it uses reshape2::melt. However, it is not the case for `spread_`. So I've written a method for data.table using dcast.data.table


hadley:
There's a lot going on here, and the code isn't that clear to me. I'd consider a revised PR if you refactor the code a bit so it's easier to understand what's happening, and include some unit tests.

PR #38, opened 2014-10-18
add an iris example for gather(), base R and dplyr

jennybc:
new example with a standard dataset
shows gather() with and without the dplyr/pipe syntax
gather() called with explicit arg = value pairs


hadley:
It would be possible (but tricky) to implement it, but I think it's better to be explicit about the new column names. I think that makes it easier for people to understand (and makes the parallel between spread and gather more clear)

PR #59, opened 2015-01-30
respect input class

holstius:
Gathering a `tbl_dt` should produce a `tbl_dt`. Add `gather_.tbl_dt` and
tests. (Also add tests for `spread`.)


hadley:
Thanks!

Issue #44, opened 2014-11-21
unnest multiple columns at once


momeara:
I'd like `unnest` to support unnesting multiple columns at once. For example,

```
x <- data_frame(
  a=c("a:b", "c"), b=c("1:2", "3"), c=c(11,22)) %>%
  transform(
    a = strsplit(a,":"),
    b = strsplit(b,":")) %>%
  unnest(a, b)
```

would produce

```
  a b  c
1 a 1 11
2 b 2 11
3 c 3 22
```

As a real world example where this comes up, the [HGNC](http://www.genenames.org) allows extracting gene family ids and descriptions, but it organizes them like this:
```
       hgnc_id                hgnc_gene_name hgnc_gene_family_ids                         hgnc_gene_family_descriptions
 1: HGNC:10006    Rh-associated glycoprotein  CD\tbloodgroup\tSLC   CD molecules\tBlood group antigens\tSolute carriers
 2: HGNC:10008 Rh blood group, CcEe antigens       CD\tbloodgroup                    CD molecules\tBlood group antigens
 3: HGNC:10009     Rh blood group, D antigen       CD\tbloodgroup                    CD molecules\tBlood group antigens
 4:  HGNC:1001         B-cell CLL/lymphoma 6      ZBTB\tZNF\tBTBD -\tZinc fingers, C2H2-type\tBTB/POZ domain containing
```

I'd like it unnest `hgnc_gene_family_ids` and `hgnc_gene_family_descriptions` simultaneously:
```
       hgnc_id                hgnc_gene_name hgnc_gene_family_ids hgnc_gene_family_descriptions
 1  HGNC:10006    Rh-associated glycoprotein                   CD                  CD molecules
 2  HGNC:10006    Rh-associated glycoprotein           bloodgroup          Blood group antigens
 3  HGNC:10006    Rh-associated glycoprotein                  SLC               Solute carriers
 4  HGNC:10008 Rh blood group, CcEe antigens                   CD                  CD molecules
 5  HGNC:10008 Rh blood group, CcEe antigens           bloodgroup          Blood group antigens
 6  HGNC:10009     Rh blood group, D antigen                   CD                  CD molecules
 7  HGNC:10009     Rh blood group, D antigen           bloodgroup          Blood group antigens
 8   HGNC:1001         B-cell CLL/lymphoma 6                 ZBTB                             -
 9   HGNC:1001         B-cell CLL/lymphoma 6                  ZNF       Zinc fingers, C2H2-type
[...]

PR #69, opened 2015-03-25
Add nest_string()/unnest_string() functions

aaronwolen:
This PR slightly modifies `unnest()` so the transform step from the example isn't necessary and can just be written as:

```
df <- data.frame(
  x = 1:3,
  y = c("a", "d,e,f", "g,h"),
  stringsAsFactors = FALSE
)
unnest(df, y)
```

I'm often doing these kind of unnesting operations and just wanted to save myself some typing by letting `unnest()` handle the string splitting. 

This also adds a `nest()` function so it's possible to round trip:

```
df %>% unnest(y) %>% nest(y)
```


hadley:
Hmmmm, that's only one type of nesting - the other is when you have something like:

```R
dplyr::data_frame(x = c(1, 2), y = list(1:3, 9:10))
```

So maybe a more specific name like `unnest_string()` ?

aaronwolen:
Ah, good call. `unnest_string()` makes sense and fits with the precedent set by `extract_numeric()`. 

But which type of nesting do you think is more common? If it's string nesting then maybe `unnest_list()` makes more sense?

aaronwolen:
Thinking about it for the past 15 seconds, `unnest_list()` sounds more like a `base::unsplit()` replacement than a data.frame function. `unnest_string()` is probably the way to go.

Are you envisioning `nest_string()`/`unnest_string()` and `nest()`/`unnest()` for list columns? Or is that nest overload?

hadley:
To me, `unnest()` seems like the atomic operation (in the sense you combine it with other atoms to do something useful), so I'd like to keep it the primary verb.

R/nest-string.R
```diff
@@ -0,0 +1,93 @@
...
+#' @export
+#' @examples
+#'
+#' df <- data.frame(
+#'   x = 1:3,
+#'   y = c("a", "d,e,f", "g,h"),
+#'   stringsAsFactors = FALSE
+#' )
+#' unnest_string(df, y)
+unnest_string <- function(data, col = NULL, sep = "[^[:alnum:]]+", ...) {
```
hadley: Doesn't this always need a col?

R/nest-string.R
```diff
@@ -0,0 +1,93 @@
...
+#' @export
+#' @examples
+#'
+#' df <- data.frame(
+#'   x = c(1, 2, 2, 2, 3, 3),
+#'   y = c("a", "d", "e", "f", "g", "h"),
+#'   stringsAsFactors = FALSE
+#' )
+#' nest_string(df, y)
+nest_string <- function(data, col = NULL, sep = ",") {
```
hadley: I'm not sure I want `nest_string()` - I think `unnest_string()` is useful, but the reverse seems a little ill-specified to me (esp how the grouping work)

hadley:
Could you PTAL at the build failure?

aaronwolen:
Looks like the build failure was introduced in #59. I added a fix here (9c6fcc4) but can submit a separate PR If you prefer.

aaronwolen:
Thanks for the comments. I see your point about `nest_string()` and removed it to wrap up this PR.

Any interest in extending `unnest()` (and `unnest_string()`) to accept multiple columns as suggested in #44?

DESCRIPTION
```diff
@@ -17,6 +17,8 @@ Imports:
     dplyr (>= 0.2),
     stringi,
     lazyeval
+Suggets:
```
hadley: Was this deliberate?

hadley:
Would you mind including a couple of unit tests too please?

DESCRIPTION
aaronwolen: I meant to add data.table to the suggested packages to address this [build failure](https://travis-ci.org/hadley/tidyr/builds/48924616), but listing it under a redundant/misspelled key was just a silly mistake (insert embarrassed emoji here).

hadley: Ah I just discovered that too and fixed it (in a slightly different way). Could you please merge/rebase?

aaronwolen:
Sure, I'll rebase and add a few tests.

hadley:
I wonder if this should be `unnest_separate()` since it's similar to separate?

aaronwolen:
Yeah, I think that makes sense. Should I update the PR?

hadley:
Yes, that would be great!  I think maybe it should have `col` and `convert` arguments to be as similar as possible to `separate()`.  And maybe it should be called `separate_rows()`? Or something like that. 

It would also be useful give this function, `extract()`, and `separate()` a common `@family`, maybe `@family string splitting` ?

aaronwolen:
From a readability perspective I definitely prefer `separate_rows()` to `unnest_separate()`. 

However, I think an argument can be made that this function is conceptually more similar to `unnest()`, which produces new rows, than it is to `separate()`, which produces a new column. If you buy that argument, `unnest_*something*()` might be more inline with user expectations, and if you don't... well, then  `separate_rows()` it is!

hadley:
It's half-way between both, but I think people are more likely to look for it after discovering that `separate()` doesn't do what they want. I think if you understand `unnest()` you might not need `unnest_string()`.

R/separate-rows.R
```diff
@@ -57,3 +57,14 @@ separate_rows_.tbl_df <- function(data, cols, sep = "[^[:alnum:].]+",
                                   convert = FALSE) {
   as_data_frame(NextMethod())
 }
+
+#' @export
+separate_rows_.grouped_df <- function(data, cols, sep = "[^[:alnum:].]+",
+                                  convert = FALSE) {
+
```
aaronwolen: Should `separate_rows()` prevent modification of grouped variables, @hadley?

hadley: It shouldn't - you can copy the approach from `separate_`:

```R
#' @export
separate_.grouped_df <- function(data, col, into, sep = "[^[:alnum:]]+",
                                 remove = TRUE, convert = FALSE,
                                 extra = "warn", fill = "warn", ...) {
  regroup(NextMethod(), data, if (remove) col)
}
```

aaronwolen:
Cool, this should be ready for review.

hadley:
Thanks!

PR #84, opened 2015-05-26
Avoid warning thrown by split.data.frame()

aaronwolen:
`nested` should be split as a list, not a data.frame. Otherwise `split.data.frame()` throws a warning when the number of rows in `nested` is not a multiple of the length of `types`.

Example:

```
unnest(data_frame(x = 1:3, y = list(1,2:3,4), z = list(5,6:7,8)))
## Warning message:
In split.default(x = seq_len(nrow(x)), f = f, drop = drop, ...) :
  data length is not a multiple of split variable
```

### DIFF
--- a/R/unnest.R
+++ b/R/unnest.R
@@ -74,7 +74,7 @@ unnest_.data.frame <- function(data, unnest_cols) {
   }
 
   types <- vapply(nested, list_col_type, character(1))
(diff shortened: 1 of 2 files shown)


hadley:
Would you mind adding a unit test too please?

hadley:
Thanks!

Issue #65, opened 2015-03-09
spread_.data.table


stefanfritsch:
Hi,

there is currently no spread for data.table which is a bit inconvenient at times. It also could take advantage of the native dcast method.

```R
spread_.data.table <- function(data, key_col, value_col, fill=NA, convert=FALSE, drop=TRUE)
{
  if( !require(data.table) )
    NextMethod()
    
  orig.cols <- setdiff(colnames(data), key_col)
  
  lhs <- paste0(orig.cols,collapse="+")
  
  Output <- dcast.data.table(data=data,
                             formula=paste0(lhs,"~",key_col),
                             value.var=value_col,
                             fill=fill,
                             drop=drop)
  
  if( convert==TRUE )
  {
    # There is the possibility for duplicated colnames
    # We err on the side of caution and just replace unambiguously new columns.
    new.cols <- setdiff(colnames(Output),orig.cols)
    
    Output[
      ,`:=`(
        (new.cols),
        lapply(.SD,
               function(col)
                 type.convert(col,as.is=TRUE))
      ), 
      .SDcols=new.cols
    ]
  }
  
  Output
}
```

hadley:
I'd consider a PR if you included unit tests.

PR #124, opened 2015-10-14
Data.table methods

stefanfritsch:
Adds data.table methods for spread and gather as per #65.

This currently needs `Imports: data.table` because otherwise the data.table brackets don't work even if I use dplyr functions(?!). It should be enough to export .datatable.aware = TRUE, but I have no idea how I get roxygen2 to do export a variable. `#' @export` doesn't work.


hadley:
I'd prefer to hold off on these until the tidyr interface is a bit more stable.

PR #132, opened 2015-10-30
Documentation improvements

krlmlr:
- roxygen5
- SE functions are internal
- All NSE functions link to their SE counterparts from the "See also" section

Maybe it's better to combine help pages of NSE and SE functions via `@rdname`.


hadley:
Thanks!

PR #135, opened 2015-10-30
Change order in unite

### LINKED ISSUE #89: When creating a variable with the same name as one of the original columns used in unite, the new column is removed (with remove = TRUE).
For example:
```r
unite(mtcars, vs, vs, am)
```
removes both the *vs* and *am* column, while not creating a new column *vs*.
If this is intended functionality, perhaps a warning message would be appropriate.

krlmlr:
First remove old columns, then add new. With test.

Fixes #89. CC @theunis


hadley:
Thanks!

Issue #144, opened 2015-12-28
Nesting data frames


lionel-:
It can be useful to nest data frames so that contextual information is available inside the nests. Here's an  implementation sketch for a `nest_df()` function:

First a C++ util borrowed from purrr to turn a grouped data frame into a list of data frames:
```C++
#include <Rcpp.h>
#include <dplyr.h>
using namespace Rcpp;

// [[Rcpp::export]]
SEXP extract_slices(const List data, const CharacterVector kept_cols) {
  ListOf<IntegerVector> indices(data.attr("indices"));
  int n_slices = indices.size();

  if (n_slices == 0) {
    return data[kept_cols];
  }

  CharacterVector classes = CharacterVector::create("tbl_df", "data.frame");
  dplyr::DataFrameSubsetVisitors visitors(data);

  List out = no_init(n_slices);
  for (int i = 0; i < n_slices; ++i) {
    List out_i = visitors.subset(indices[i], classes);
    out[i] = out_i[kept_cols];
  }

  return out;
}
```

The R functions:
```r
nest_df <- function(data, key, ...) {
  nest_cols <- unname(dplyr::select_vars(names(data), ...))
  nest_df_(data, key, nest_cols)
}

nest_df_ <- function(data, key, nest_cols) {
  group_cols <- setdiff(names(data), nest_cols)

  dfs <- data %>%
    dplyr::group_by_(.dots = group_cols) %>%
    extract_slices(nest_cols)

  out <- data %>%
    dplyr::select_(.dots = group_cols) %>%
    dplyr::distinct()

  out[[key]] <- dfs
  dplyr::tbl_df(out)
}
```

And then:
```r
df <- data_frame(
  a = c(rep(1, 3), rep(2, 3)),
  b = rep(letters[1:3], each = 2),
  c = 1:6,
  d = 6:1
)
df
#> Source: local data frame [6 x 4]
#>
#>       a     b     c     d
#>   (dbl) (chr) (int) (int)
#> 1     1     a     1     6
#> 2     1     a     2     5
#> 3     1     b     3     4
#> 4     2     b     4     3
#> 5     2     c     5     2
#> 6     2     c     6     1

df %>% nest_df("key", c, d)
#> Source: local data frame [4 x 3]
#>
#>       a     b            key
#>   (dbl) (chr)          (chr)
#> 1     1     a <tbl_df [2,2]>
#> 2     1     b <tbl_df [1,2]>
#> 3     2     b <tbl_df [1,2]>
[...]

PR #147, opened 2015-12-30
Add key argument to nest()

lionel-:
With the slightly faster visitors implementation. See #144


src/nest.cpp
```diff
@@ -0,0 +1,24 @@
+#include <Rcpp.h>
+#include <dplyr.h>
+using namespace Rcpp;
+
+// [[Rcpp::export]]
+SEXP extract_groups(const List data, const CharacterVector nested_cols) {
+  ListOf<IntegerVector> indices(data.attr("indices"));
```
hadley: Is this faster than the R equivalent using the indices attribute? The reason I didn't take this approach original is that sometimes indices are lazily computed - i.e. after you `filter()` a grouped data frame, the indices are dropped, and only recomputed when needed (IIRC)

lionel-: ow, I didn't know that. I always assumed indices were kept in a working state.

I don't see that behaviour with `filter()` but it may be safer to use your original implementation.

hadley: I can't remember exactly when they get lost, but `do.grouped_df()` has:

```R
  # Force computation of indices
  if (is.null(attr(.data, "indices"))) {
    .data <- grouped_df_impl(.data, attr(.data, "vars"),
      attr(.data, "drop") %||% TRUE)
  }
```

dplyr should probably have an exported method to get the indices.

@romainfrancois can you remember when the grouped indices get dropped?

lionel-:
@hadley I think it would be more consistent to also have select semantics for grouped dfs.

This would let the users specify nested columns, so that the following would result in 2-column data frames:
```r
mtcars %>% group_by(cyl) %>% nest("data_col", disp, drat)
```

This would result in zero-columns dfs:
```r
mtcars %>% group_by(cyl) %>% nest("data_col")
```
And this would select everything besides "cyl":
```r
mtcars %>% group_by(cyl) %>% nest("data_col", everything())
```

hadley:
@lionel- I switch from select semantics to mutate semantics because:

1. I think most of the time you'll be unnesting a single variable

lionel-:
@hadley it's about `nest()` here ;)

hadley:
@lionel- oh oops. Yes, it definitely should have select semantics, rather than ignoring as it does now.

hadley:
So the complete set of changes to nest should be:

* add `key` argument that allows you to choose what the new column is
* add `.id` argument that allows you to create a variable from the list names
* use `...` to restrict nesting variables

Do you want to have a go at any of these? Otherwise, I'm happy to do them.

PR #151, opened 2016-01-03
Cleanup for separate() and spread()

krlmlr:
- Warn when using defunct ... argument to separate()
- Declare methods for "grouped_df" class


hadley:
Thanks!

PR #156, opened 2016-01-19
add default period = 1 and conserve attributes

matthieugomez:


hadley:
A pull request without any unit tests or motivation is not very useful.

PR #173, opened 2016-03-14
Improve gather_.data.frame when na.rm = TRUE but df really haven't NAs

artemklevtsov:
Hi.

It should improve performance when a data.frame no contains NAs.


hadley:
Thanks - could you please also add a bullet point to NEWS?

hadley:
I think you forgot to update this PR

PR #176, opened 2016-03-24
Allow replace_na replacements to be given in the ... args

dgrtwo:
Allow replace_na replacements to be given as

```r
replace_na(data, x = 0, y = "unknown")
```

along with allowing the original behavior of

```r
replace_na(data, list(x = 0, y = "unknown"))
```

Also allows NSE evaluation to replace one NAs in one column with another column:

```r
df <- data.frame(x = c(1, 2, NA), y = c(3, NA, 4))
replace_na(df, x = y)
```
It's backwards compatible, ... wasn't being used. Includes additional test cases and examples.

### DIFF
--- a/R/replace_na.R
+++ b/R/replace_na.R
@@ -1,13 +1,34 @@
 #' Replace missing values
 #'
+#' Replace missing values in columns, either with a specific replacement
+#' value or with a value (or expression) based on other columns in the
+#' data.
+#'
 #' @param data A data frame.
 #' @param replace A named list given the value to replace \code{NA} with
 #'   for each column.
-#' @param ... Additional arguments for methods. Currently unused.
+#' @param ... Named arguments providing values to replace \code{NA} with
+#' for each column, or an expression based on other columns
+#'
+#' Note that if you are replacing a column named \code{replace}
+#' you need to use replace = list(replace = 0) argument rather than a named
+#' argument. This is to keep \code{replace_na} backwards-compatible.
+#'
 #' @examples
 #' library(dplyr)
 #' df <- data_frame(x = c(1, 2, NA), y = c("a", NA, "b"))
+#'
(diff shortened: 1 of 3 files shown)


hadley:
Hmmm, in principle I like the idea of supporting dots, but I don't like having both list and `...` syntaxes in the same function.  And it starts to overlap with `dplyr::coalesce()`

dgrtwo:
I hadn't seen coalesce and I certainly see how that replaces the need to refer to other columns; e.g. with `mutate(x = coalesce(x, y))` rather than `replace_na(x = y)`.

As for list vs `...`, I guess I'm not clear why the argument is a list in the first place. `replace(x = 0)` feels more natural (like a mutate) to me than `replace(list(x = 0))` and it's not missing any expressivity (Are there other similar dplyr/tidyr functions that take a list like this?). But I completely understand if the change is not worth either reverse incompatibility or the behavior of allowing both syntaxes.

hadley:
Because of `expand()`, I think.

hadley:
This just feels like too small a gain for the amount of work - but lets see how it plays out once `coalesce()` is in circulation. It may be that there's some missing verb that would let you do `mutate_each()` with a `coalesce()` that has different arguments for each column.

dandermotj:
I would add, it would be nice to be able to specify ranges of columns using `x:y` like in `unite` and dplyr.
```r
replace_na(data, x:y = 0, a:b = "unkown") 
```
The first time I used `replace_na` I automatically tried to select columns this way and was confused when it didn't work first time!

dgrtwo:
@dandermotj  That would be a syntax error; you'd need backticks or quotes around `a:b`, and that point I'm not sure it's clear enough to be worth it.

dandermotj:
You're correct, don't know why I thought that would work..

PR #180, opened 2016-04-10
added full_set() with examples

eliasdabbas:
Similar to `full_seq()` I think we can have `full_set()` which does the same but for character vectors. Two examples added for `expand()` and `complete()`


hadley:
I'm not sure why you wouldn't just use `unique()`?

eliasdabbas:
This is not to find unique values, it is to complete the data by adding implicitly missing ones, but for character vectors. 
I expanded on the example you have in the documentation, here's how it might work: 

`full_set <- function(x, vec) {
   if (any(!x %in% vec)) {
     stop("x has to be a subset of vec")
   }
  return(vec)
}

resources <- frame_data(
  ~year, ~metric, ~value,
  1999, "coal", 100,
  2001, "coal", 50,
  2001, "steel", 200
)

all_metrics <- c("aluminium", "coal", "steel") 
resources %>% complete(metric = full_set(metric, all_metrics), year)
`

hadley:
In that case, why wouldn't you do `resources %>% complete(metric = all_metrics, year)` ?

eliasdabbas:
Didn't think it would work. Thanks!

PR #187, opened 2016-05-16
Fixes #170, regarding order of separate columns when fill = "left".

### LINKED ISSUE #170: separate with fill = "left" reverses column order
When `fill = "left"`, not only do NA values get filled in from the left, but the columns are filled in reverse order. Example:

``` r
library(tidyr)

d <- data.frame(x = c("1-2-3", "2-3", "3"))

# I understand this behavior
separate(d, x, c("a", "b", "c"))
#> Warning: Too few values at 2 locations: 2, 3
#>   a    b    c
#> 1 1    2    3
#> 2 2    3 <NA>
#> 3 3 <NA> <NA>

# here, surprised that 3 and 2 are reversed:
separate(d, x, c("a", "b", "c"), fill = "left")
#>      a    b c
#> 1    1    2 3
#> 2 <NA>    3 2
#> 3 <NA> <NA> 3
```

I think this is undocumented since the docs for fill only say that NAs are filled from the left, not that the values will be reversed. I also think the most common use case would be splitting a column where there is an optional first value (like either a `d/m/y` date or an `m/y` date) and can't think of a use case where reversing the order *only* when there aren't enough values would be helpful.

It seems like changing it would be a [straightforward change here](https://github.com/hadley/tidyr/blob/master/src/simplifyPieces.cpp#L31), but I don't know if this is desired behavior. I'm happy to PR it myself.

dgrtwo:
Fixes #170, regarding order of separate columns when fill = "left".

Added test case for this situation

### DIFF
--- a/src/simplifyPieces.cpp
+++ b/src/simplifyPieces.cpp
@@ -27,9 +27,13 @@ List simplifyPieces(ListOf<CharacterVector> pieces, int p,
     } else if (x.size() < p) { // too small
       tooSml.push_back(i + 1);
 
+      int gap = p - x.size();
       for (int j = 0; j < p; ++j) {
-        int idx = fillLeft ? p - j - 1 : j;
-        out[idx][i] = (j < x.size()) ? x[j] : NA_STRING;
+        if (fillLeft) {
+          out[j][i] = (j >= gap) ? x[j - gap] : NA_STRING;
+        } else {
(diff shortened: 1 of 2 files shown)


hadley:
Can you please add a bullet to NEWS to?  (And it looks like you need a rebase)

PR #188, opened 2016-05-17
Use tibble:::as_data_frame.matrix()

### LINKED ISSUE #131: Use as_data_frame from dplyr
instead of matrixToDataframe

krlmlr:
- remove redundant matrixToDataFrame()
- explicitly import tibble

Fixes #131.


hadley:
Maybe import `data_frame()` too? It's used a lot in the tests.

hadley:
And then we can drop the dplyr dependency?

krlmlr:
Not quite, tidyr uses quite a lot from dplyr.

hadley:
Oh yeah, not sure what I was thinking.

krlmlr:
On that note, perhaps spread and gather for SQL sources (hadley/dplyr#1753) could live here -- we have the dependency anyway, and everything can be expressed with high-level operations.

hadley:
Yeah, that makes sense

krlmlr:
PTAL

hadley:
Thanks!

Issue #194, opened 2016-05-27
Consider adding a drop_na verb


janschulz:
[forwarded from https://github.com/hadley/dplyr/issues/1797 as per @hadley: "This feels more like a tidyr verb to me."]

pandas has the [pandas.DataFrame.dropna()](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.dropna.html) which lets you drop rows which have NAs in it but also let you specify subset of columns (-> `subset`) to be looked at for NAs.

A similar dplyr verb would be 

```R
drop_na <- function(data, ...){
    if (missing(...)){
        f = complete.cases(data)
    } else {
        f <- complete.cases(select_(data, .dots = lazyeval::lazy_dots(...)))
    }
    filter(data, f)
}
```

Examples: 
```R
> df <- data.frame(a=c(1,2,3,4,NA), b=c(NA,1,2,3,4), ac=c(1,2,NA,3,4))
> df %>% drop_na(a,b)
  a b ac
1 2 1  2
2 3 2 NA
3 4 3  3
> df %>% drop_na(starts_with("a"))
  a  b ac
1 1 NA  1
2 2  1  2
3 4  3  3
> df %>% drop_na()
  a b ac
1 2 1  2
2 4 3  3
```

I can clean the above up and submit a PR if such a verb would be considered.

hadley:
Yeah, I'd be happy to review a PR.

PR #201, opened 2016-06-13
Add drop_na verb

janschulz:
This is modeled after pandas `df.dropna()` method and follows the same pattern  as `distinct()`:
    
* if no argument is given, consider all variables while dropping rows
* if arguments are given, use them to select variables and then only consider
  these variables while dropping rows.
    
Also included is a NSE version (`drop_na_()`).

Closes: #194


R/drop_na.r
```diff
@@ -0,0 +1,36 @@
...
+#'    variables are considered while dropping rows
+#' @keywords internal
+#' @export
+drop_na_ <- function(data, relevant_cols){
+  if (is.null(relevant_cols) || length(relevant_cols) == 0) {
+    f = stats::complete.cases(data)
+  } else {
+    f <- stats::complete.cases(data[relevant_cols])
+  }
+  dplyr::filter(data, f)
```
hadley: Not really any point in using filter here - just subset directly

R/drop_na.r
```diff
@@ -0,0 +1,36 @@
...
+
+#' Standard-evaluation version of \code{drop_na}.
+#'
+#'
+#' @param data A data frame.
+#' @param relevant_cols Character vector of column names. If empty or \code{NULL}, all
+#'    variables are considered while dropping rows
+#' @keywords internal
+#' @export
+drop_na_ <- function(data, relevant_cols){
```
hadley: I think this should be generic to match the other tidyr verbs

R/drop_na.r
```diff
@@ -0,0 +1,36 @@
...
+drop_na <- function(data, ...) {
+  relevant_cols <- unname(dplyr::select_vars(colnames(data), ...))
+  drop_na_(data, relevant_cols)
+}
+
+#' Standard-evaluation version of \code{drop_na}.
+#'
+#'
+#' @param data A data frame.
+#' @param relevant_cols Character vector of column names. If empty or \code{NULL}, all
```
hadley: Why check for NULL here? I think it would be better to check that `relevant_cols` is a character vector. (And I think a better name would simple be `vars`)

R/drop_na.r
janschulz: I saw that, but as far as I underatand it, it should not behave different for grouped vs ungrouped data nor for `tbl_df`. Or should this be implemented just in case?

R/drop_na.r
janschulz: RE `NULL`: more or less for convenience, so that one can write `drop_na_(data, NULL)` instead of `drop_na_(data, character())`

Re checking that it is a character vector: I didn't add that because e.g. `fill_.data.frame` also doesn't do this check. Should I add a commit which adds it there, too?

R/drop_na.r
hadley: It should be implemented to be consistent - and, e.g., someone might want to create a method that works with remote database tables.

R/drop_na.r
janschulz: done

R/drop_na.r
janschulz: implemented generic and friends

R/drop_na.r
janschulz: removed NULL check (-> not accepted anymore) and check that the input is a character vector.

janschulz:
The travis fail seems to be unrelated:

```
Quitting from lines 238-239 (tidy-data.Rmd) 
Error: processing vignette 'tidy-data.Rmd' failed with diagnostics:
unused argument (envir = asNamespace("tibble"))
Execution halted
```

janschulz:
rebased on top of the new release (News changes)

hadley:
Thanks - I need to do a point release for old-R compatibility, and I'll merge after that.

janschulz:
Rebased on top of master (News entry...) and repushed.

PR #205, opened 2016-06-13
Use get() instead of getS3method() for R < 3.3

### LINKED ISSUE #202: as_data_frame_matrix: "envir" argument in getS3method
It seems like `envir` is not a valid argument in `utils::getS3method` in R version 3.2.4. Couldn't find a specific mention of when this was added to `R.utils`, but unless there is something wrong with my R install I suppose this warrants a version bump for the R in `DESCRIPTION`. 

`getS3method` is only used on one line however, so it might be a better idea to just add an `if` statement for older versions of R (where envir is not an argument) and something like this:

```R
get("as_data_frame.matrix", envir = asNamespace("tibble"))
```

Could submit a PR if I have the right idea here (and figure out which version of `R.utils` added the argument).

-Kristian

krlmlr:
Fixes #202. Tested locally on R 3.2.5 and R 3.3-patched.


krlmlr:
Wanted to add NEWS entry, but release branch doesn't seem to be merged yet.

itsdalmo:
Novice question here; Why not use a `if` statement here which dep. on R version? Is it because of performance or something else? Also, how long will you have to wait before exchanging `get` with `getS3method` and bumping the R version dependency (if ever)?

krlmlr:
I just went the path of least resistance here. The code is hit by about 10 tests, so if it ever breaks, we may think again about updating it.

hadley:
Just merged

krlmlr:
Are you, too, now receiving GitHub notifications when somebody pushes to an active pull request?

hadley:
Thanks!

And yes - it seems to be a new github feature. I haven't seen it announced yet though.

PR #218, opened 2016-07-15
change right_join to full_join in expand docs

srs70187:
In the expand documentation I believe full_join was meant instead of right_join.


hadley:
Why do you think this should be `full_join()`?

hadley:
Oh I had the arguments around the wrong way

srs70187:
Cool, thanks for checking.

PR #228, opened 2016-08-23
Description of `cols` parameter

wibeasley:
From the docs, I wasn't sure if it was supposed to be a list or vector.  To be consistent, I copied the description from [expand()](https://github.com/hadley/tidyr/blob/ea2806a65111cc7237f8724b4b338c9d22f310aa/R/expand.R#L93).


hadley:
Looks like we fixed locally. Thanks!

PR #229, opened 2016-08-25
Add a gather_vars() generic. Fixes #195

### LINKED ISSUE #195: Provide a `gather()` generic?
Similar to how ggvis provides custom methods for dplyr generics, I'd like to provide a `gather()` method in plotly.

This doesn't seem possible currently given that [`gather()` is not a generic and assumes the first argument is a data frame/table.](https://github.com/hadley/tidyr/blob/164b9bffc75a431224845da0b80526f94686c5b3/R/gather.R#L49-L53)

There may be a better way, but it seems the easiest solution would be to export a `gather()` generic, and essentially copy it's current body to a `gather.data.frame` method. Does that sound reasonable? If so, I'd be happy to send a pull request.

cpsievert:


hadley:
Post tidyselect, this fix is no likely no longer correct, but I promise I'll take a look as part of this release.

PR #266, opened 2016-12-18
Fix doc for TB dataset variables.

mkcor:
Just a small edit.

I'm not sure how to update `roxygen2` to version 5.0.1.9000, in order to run `roxygen2::roxygenise()` without 'reverting' commit acacd6a38...

I guess the RStudio package manager does not offer it because it's a dev version.


hadley:
Thanks! I'll re-document locally.

PR #267, opened 2017-01-03
Specify full license text in repository

dhimmel:
Currently, the repository does not include the full text of its MIT License, which is problematic (see https://github.com/tidyverse/tidyr/issues/262). This pull requests specifics the full license text corresponding to `tidyr`'s [CRAN page](https://cran.r-project.org/web/packages/tidyr/index.html).

`LICENSE.txt` is not transmitted to CRAN (as per `.Rbuildignore`) to remain complaint with the [R-exts guidelines](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Licensing). This solution was initially suggested by @gavinsimpson. Also paging @hadley and @cderv as reviewers.


hadley:
I'd prefer `LICENSE.md`

dhimmel:
@hadley markdown will work well for the MIT License, where the only formatting that occurs is converting newlines to spaces in paragraphs.

As a general comment, I think we want to make sure the license on CRAN and GitHub do not diverge. So for licenses that don't render well to markdown, keeping `.txt` may make sense.

hadley:
There are markdown versions of all the major licenses though.

dhimmel:
> There are markdown versions of all the major licenses though.

Agreed. Just a note to remember to check that the markdown license uses the same text (as there are often small textual variations in the common licenses).

Also there is the potential issue that formatting alters the interpretation of the license. This empirical support here is the ALL CAPS disclaimers in many licenses, including the MIT License. I personally think the benefits of nice formatting outweigh this risk. This is less of an issue if you only have a single copy of your license in markdown. But here where we have a text and markdown version, there could be an issue. I'm not too worried, but wanted to bring the argument up.

dhimmel:
I rebased to resolve conflicts that had accrued in `.Rbuildignore`.

@hadley any more suggestions?

The pull request is complete from my end.

hadley:
Thanks!

jimhester:
The only slight issue with this is that while [GitHub's license finder](https://github.com/benbalter/licensee) will detect licenses in LICENSE.md it currently looks at the LICENSE file first and in this case does not find a valid license there, so does not try to fall back to look at LICENSE.md.

dhimmel:
@jimhester, the [`what-we-look-at.md`](https://github.com/benbalter/licensee/blob/91430bd136a24c3a3dc0189be5c40807d31954a6/docs/what-we-look-at.md) is moderately helpful but doesn't describe the licensee order of operations for license detection.

> This is why even with this change there is no automatically detected license for tidyr.

Just confirming that license auto-detection did not work previously and still doesn't work.

@jimhester, I see you created a licensee PR to support the CRAN license guidelines: https://github.com/benbalter/licensee/pull/129 and https://github.com/benbalter/licensee/pull/160. Once this work is done, will the detection work?

jimhester:
> Once this work is done, will the detection work

That is the idea...

PR #268, opened 2017-01-05
Fix wording of error message in test

krlmlr:
@hadley: This fixes compatibility with the dev version of tibble. Can we do a mini-release?

I wonder what's the best strategy for testing against development versions. Is there a way to install all dependencies from GitHub (if available) instead of CRAN?


hadley:
Sure, I can do that

krlmlr:
Thanks.

krlmlr:
@hadley: The tests in 7d6302321 are compatible with tidyverse/tibble@1952a51b, which I intend to release to CRAN as soon as revdep checks pass.

PR #273, opened 2017-01-19
added @examples to 'gather_'

tbrugz:
Added an example to `gather_` that is similar to the one from `gather`


hadley:
`gather_` is going away, but thanks for the idea.

Issue #272, opened 2017-01-17
`nest()` preserves attributes (including class) but `unnest()` does not


petereckley:
Is this intended behaviour? I could not find it in the docs or NEWS. I would advocate for `unnest()` to preserve attributes since this makes it easier to use in subclasses that specialise data.frames.

karimn:
I'm seeing this too. Here's an example:

```r
> nested.data <- tibble(z = 2)
> attr(nested.data, "someattr") <- "xyz"
> test.data <- tibble(x = 1, y = list(nested.data))
> str(test.data)
Classes ‘tbl_df’, ‘tbl’ and 'data.frame':	1 obs. of  2 variables:
 $ x: num 1
 $ y:List of 1
  ..$ :Classes ‘tbl_df’, ‘tbl’ and 'data.frame':	1 obs. of  1 variable:
  .. ..$ z: num 2
  .. ..- attr(*, "someattr")= chr "xyz"
> test.data %>% unnest(y) %>% str
Classes ‘tbl_df’, ‘tbl’ and 'data.frame':	1 obs. of  2 variables:
 $ x: num 1
 $ z: num 2
```

cb4ds:
I am working on this and ran into the following question regarding whether this is a defect or by design:

A nested column might contain multiple data frames with multiple sets of attributes (potentially with the same or different attribute names).  In my opinion, if the nested column has a single data frame it makes very good sense that it would be useful to preserve the attributes.  However the question of what to do with multiple data frames nested as a list is a problem.  One solution is to drop attributes on multiple dataframes being unnested - however this leads to the undesirable situation where the function behaves differently based on one vs. multiples in content.  This is unintuitive and may lead to more questions than it is worth.  

If it is still desirable to go ahead with the fix and keep attributes on unnest - what is the preferred way to handle multiple data frames being unnested from a list?

Using the example above:
`
> nested.data <- tibble(z = 2)
> attr(nested.data, "someattr") <- "xyz"
> nested.data.2 <- tibble(z = 4)
> attr(nested.data.2, "anotherattr") <- c("ABC", 123)
> test.data <- tibble(x = 1, y = list(nested.data, nested.data.2))
`

petereckley:
[...]

PR #275, opened 2017-01-31
#272 (unnest preserves top-level attributes) fix and test

cb4ds:
Issue #272 Resolved - unnest now preserves all top-level attributes EXCEPT (dim, dim names, names, row.names).

Notes on changes:
- mapply is used to assign additional attributes to the output object without unassigning all attributes
- the class attribute is handled separately and is set as a superset of the classes of the original object and the result object.


hadley:
I'd rather think about this a little more before implementing the code. But I do really appreciate the effort you put into making the PR.

Issue #193, opened 2016-05-25
spread with drop=FALSE and additional columns


JerryWho:
Let's say I have the following data.frame:

``` R
df <- data.frame(year=rep(1:5,3), blood=rep(c("A", "B", "0"), 5), count=1:15)
df$blood <- factor(df$blood, levels=c("A", "B", "AB", "0"))

df
       year blood count
1     1     A     1
2     2     B     2
3     3     0     3
4     4     A     4
5     5     B     5
6     1     0     6
7     2     A     7
8     3     B     8
9     4     0     9
10    5     A    10
11    1     B    11
12    2     0    12
13    3     A    13
14    4     B    14
15    5     0    15
```

As you can see the value "AB" doesn't appear in the data.frame.

So I have to use `drop=FALSE` when spreading the data:

``` R
 spread(df, blood, count, drop=FALSE)
  year  A  B AB  0
1    1  1 11 NA  6
2    2  7  2 NA 12
3    3 13  8 NA  3
4    4  4 14 NA  9
5    5 10  5 NA 15
```

So let's say there's an additional column `year2`:

``` R
df$year2 <- df$year * 2
spread(df, blood, count, drop=FALSE)

   year year2  A  B AB  0
1     1     2  1 11 NA  6
2     1     4 NA NA NA NA
3     1     6 NA NA NA NA
4     1     8 NA NA NA NA
5     1    10 NA NA NA NA
6     2     2 NA NA NA NA
7     2     4  7  2 NA 12
8     2     6 NA NA NA NA
9     2     8 NA NA NA NA
.....
```

We get for every combination of year and year2 a line in the resulting data.frame. That's no what I've expected to get.

So I tried to add the functionality in https://github.com/JerryWho/tidyr/tree/missingvalues.
Now I get

``` R
 spread(df, blood, count, drop=FALSE, other_col_as_one = TRUE)
  year year2  A  B AB  0
1    1     2  1 11 NA  6
2    2     4  7  2 NA 12
3    3     6 13  8 NA  3
4    4     8  4 14 NA  9
5    5    10 10  5 NA 15
```

Can I achieve this result with the orignal version of tidyr? Otherwise you can maybe add this functionality into the official version.

hadley:
I'd rather make `drop = FALSE` simply do the right thing

jennybc:
Isn't this multi-spread (#149)? Maybe that's not exactly true, but it has an awful lot of overlap with the middle bit of those examples.

```r
[...]

PR #285, opened 2017-03-25
Complete Empty Factors

### LINKED ISSUE #270: complete() does not use factor levels if no data
complete() has behavior advertised as follows: 

> For factors, the full set of levels (not just those that appear in the data) are used.

However, this seems to only be the case so long as you have at least one record.  A few workarounds are less than optimal in my opinion, and it would be quite helpful for this functionality to work with an empty data.frame
 
```
options <- factor(c('a','b','c'))

df <- data.frame(column=character(), column2=character()) 
levels(df$column) <- options; levels(df$column2) <- options;

## Does not work - though I would expect it to
df %>% complete(column, column2)

## Works - nice if you have options readily available
df %>% complete(column=options, column2=options)

## Works - based on factors, and a bit tedious
df %>% complete(column=factor(levels(column),levels=levels(column)),column2=factor(levels(column),levels=levels(column)))



## With data, things work fine:
df <- data.frame(column=c('a'),column2=c('b'))
levels(df$column) <- options; levels(df$column2) <- options;

## Does not work above - works here
df %>% complete(column, column2)
```

### LINKED ISSUE #271: complete() docs suggest non-working code
From `?complete`:

>  For continuous variables, you may need to fill in values that don't appear in the data: to do so use expressions like `year = 2010:2020` or `year = full_seq(year)`.

But the latter should be `year = full_seq(year, 1)` or will raise an error.

### LINKED ISSUE #276: ?complete() documentation refers unexpectedly to expand() re: nesting
The documentation for `complete()` says:

>To find all unique combinations of x, y and z, including those not found in the data, supply each variable as a separate argument. To find only the combinations that occur in the data, use nest: **expand(df, nesting(x, y, z))**.
>You can combine the two forms. For example, **expand(df, nesting(school_id, student_id), date)** would produce a row for every student for each date.

I was confused to see `expand(...)` and not `complete(...)` in the example.  This is copy-pasted text from `?expand`; could the "expand" just be replaced by "complete"?

colearendt:
Close #270, close #271, close #276 

Address documentation issues.

One documentation issue involved code that did not execute (#271).  Another was not a documentation issue, per say, it involved complete inhering an expand parameter (#276).

Decision on behavior for factors is yet to be decided.  This PR will change behavior so that empty factors are not dropped (by expand, crossing, complete), but "completed."  nesting will continue to drop empty factors per the original behavior.  

This required a parameter on the "drop_empty" function to alter behavior with factors.  factor=FALSE is the legacy behavior.

Added tests with more clarity on different argument types.  Also added explicit calls to the dplyr package in some of the relevant tests.


hadley:
It's been a long time, so I wanted to check that you're still interested in this PR before I start the review process.

colearendt:
Thanks for the note!  Yes, I definitely am.  I'm not sure if there is a better way to implement, but I think the functionality is useful for improving the consistency of how factors are treated.

hadley:
BTW in the future, it's easier for me to review if you do one PR per issue; before I closely read the linked issues I'd assumed that they were all related.

R/expand.R
```diff
@@ -150,8 +150,8 @@ cross_df <- function(x, y) {
   y_idx <- rep(seq_len(nrow(y)), nrow(x))
   dplyr::bind_cols(x[x_idx, , drop = FALSE], y[y_idx, , drop = FALSE])
 }
-drop_empty <- function(x) {
-  empty <- map_lgl(x, function(x) length(x) == 0)
+drop_empty <- function(x, factor=TRUE) {
```
hadley: Can you please follow http://style.tidyverse.org/syntax.html#spacing and put space around `=` and` before `,`?

tests/testthat/test-complete.R
```diff
@@ -13,3 +13,11 @@ test_that("preserves grouping", {
   expect_s3_class(out, "grouped_df")
   expect_equal(dplyr::groups(out), dplyr::groups(df))
 })
+
+test_that('expands empty factors', {
+  emptyfactor <- factor(character(),levels=c('a','b','c'))
```
hadley: You can just write `factor(levels = c('a', 'b', 'c'))`

(But please use `"` for strings)

R/expand.R
hadley: Also I don't _think_ this actually needs to be an option, but it does need a comment mentioning that we need to keep it consistent with `ulevels()`

tests/testthat/test-expand.R
```diff
@@ -46,11 +46,26 @@ test_that("preserves ordered factors", {
...
   expect_equal(expand(tb, x, y = numeric()), tb)
   expect_equal(nesting(x = tb$x, y = numeric()), tb)
   expect_equal(crossing(x = tb$x, y = numeric()), tb)
+
+  expect_equal(expand(tb, x, y = character()), tb)
+  expect_equal(nesting(x = tb$x, y = character()), tb)
+  expect_equal(crossing(x = tb$x, y = character()), tb)
+})
+
+test_that('zero length factor inputs are completed by expand & crossing, dropped by nesting', {
```
hadley: Why does nesting drop?

R/expand.R
colearendt: Yes, of course.  The option there is due to the nesting discussion below (nesting has different behavior).

tests/testthat/test-expand.R
colearendt: This is a fantastic question and one that I should have noted.  Per the docs:

> nesting() is the complement to crossing(): it only keeps combinations of all variables that appear in the data

As a result, I was not sure if an empty factor should be considered a "combination appearing in the data."  If so, I am happy to work on an implementation.  I can see where this might come across as a lack of consistency - however, this is a context where `nesting` / `crossing` are designed to behave differently. 

For instance, non-empty factors are currently handled differently by `nesting` and `crossing.` My aim was to maintain that dichotomy:

```r
tb <- tibble::tibble(x=1:2)                       
emptyfactor <- factor(levels = c("a","b"))        
                                                  
simplefactor <- factor(c("a"),levels = c("a","b"))
                                                  
## nesting factor behavior                        
nesting(x = tb$x, y = simplefactor)               
#> # A tibble: 2 x 2
#>       x      y
#>   <int> <fctr>
#> 1     1      a
#> 2     2      a
                                                  
## crossing factor behavior                       
crossing(x = tb$x, y = simplefactor)              
#> # A tibble: 4 x 2
#>       x      y
#>   <int> <fctr>
#> 1     1      a
#> 2     1      b
#> 3     2      a
#> 4     2      b
```

hadley: Yeah, I don't see how you can "nest" the empty levels of a factor because they don't appear in the data. I think this is the issue at the heart of #193

PR #288, opened 2017-04-02
Added function 'binarize'

MaximeWack:
It turns ragged column(s) into sparse columns (one logical column for each of the different values).

For example, `binarize(df, starts_with("col"), prefix = "letter_")`

where df is

|id|col1|col2|col3|
|--|----|----|----|
|1|A|B|C|
|2|B|C|NA|
|3|D|NA|NA|
|4|B|D|NA|

gives

|id|letter_A|letter_B|letter_C|letter_D|
|--|----|----|----|----|
|1|T|T|F|F|
|2|F|T|T|F|
|3|F|F|F|T|
|4|F|T|F|T|

I found this use case many times in medical surveys where sometimes disease history is (badly) recorded using multiple drop-down lists or sets of checkboxes. IIRC, google surveys also treats sets of checkboxes this way, with one column containing semi-colon separated values. This can be dealt with using a call to `separate` then a call to `binarize`.

I've pondered a (too) long time whether to submit this function to **forcats** or **tidyr**, but it seems a better fit here, given the examples and the use pattern.
Also, I'm not set on the name, maybe something like `sparse`, or `scatter` would be better. If this calls for an `unsparse` or `rag` function that would do the opposite operation, I'd be happy to provide!


hadley:
I think I'd like to tackle this problem a different way. Would you mind filing an issue so I don't forget about it? Thanks!

PR #295, opened 2017-04-20
Add explicit tests for grouped unnests

krlmlr:
`unnest.grouped_df()` method already implemented.

Do we need a `rowwise_df` method? There seem to be no other methods for `rowwise_df`.


hadley:
Thanks! 

No need for `rowwise_df` method since we're moving away from rowwise objects

PR #301, opened 2017-05-12
Port tidyr to tidyeval

lionel-:
Depends on https://github.com/tidyverse/dplyr/pull/2767

- Arguments selecting existing columns are taken with `dplyr::select_var()` and support unquoting of column names and positions.

- Arguments specifying new columns are taken with `rlang::quo_name()` and support unquoting of symbols and strings.

- Dispatch has been a bit simplified with a `reconstruct_tibble()` function used in a single `data.frame` method. Dispatch of tidyeval methods with named arguments is a bit difficult because `enquo()` does not work well with `NextMethod()` :/

- Various cleanups: use rlang and purrr API, use `tibble()` instead of `data_frame()`, etc


lionel-:
TODO?

* remove warnings about bare symbols for non-existing objects being now discouraged in the tidyverse. They have "name semantics", like the LHS of arguments.

* Take `into` argument of separate with `vars()` and document it to have name semantics.

  ```r
  df %>% separate(x, vars(a, b))
  ```

hadley:
Can you update to latest dplyr, rlang etc and fix the merge conflict?  Should we use tidyselect here too?

I think we should plan to do a tidyr release that just contains these changes, and we'll call it tidyr 0.7.0 (for these popular packages I think using an odd-even release schedule where odd numbers are API changes is probably a good idea).

PR #302, opened 2017-05-13
Don't test error message that is created by tibble

krlmlr:
@hadley: Could you please submit a point release, so that we can update tibble? I've asked @jimhester to submit readr, too.


hadley:
I'll do a point release tomorrow.

Issue #300, opened 2017-05-02
separate_rows() drops <list> columns in tibbles


burchill:
I noticed the behavior where using `tidyr::separate_rows()` drops `<list>` columns in tibbles.

Below is a small reproducible example:

```r
library(dplyr)
library(tidyr)
# Package versions:
# dplyr_0.5.0     tidyr_0.6.1 

# Make a data frame and turn it into a tibble
d <- data.frame(x=c(1,2,3), y=c("a,A,alpha","b","c")) %>% 
  dplyr::tbl_df()
# One list column will have a model, just to demonstrate
model_data <- data.frame(
  x=c(rnorm(100,-1,1),rnorm(100,100,1)),
  y=c(rnorm(100,-1,1),rnorm(100,100,1))
  )
# Add in the model 
d$mod <- list(lm(y~x,data=model_data))
# Add in another test list column
d$reg_list <- list("test")

print(d)
# A tibble: 3 × 5
#      x         y      mod     mod2  reg_list
#  <dbl>    <fctr>   <list>   <list>    <list>
#1     1 a,A,alpha <S3: lm> <S3: lm> <chr [1]>
#2     2         b <S3: lm> <S3: lm> <chr [1]>
#3     3         c <S3: lm> <S3: lm> <chr [1]>

# Now the `mod` and `reg_list` columns disappear
d %>% 
  tidyr::separate_rows(y,sep=",") %>%
  print()
# A tibble: 5 × 2
#      x     y
#  <dbl> <chr>
#1     1     a
#2     1     A
#3     1 alpha
#4     2     b
#5     3     c
```

hadley:
Thanks for providing some code that illustrates the problem — it's a great start 😄 However, your code is currently a bit too complicated: there's quite a bit of extraneous stuff in there that doesn't seem directly related to the problem. Can you please try and simplify your example some more? The more minimal you can make the reprex, the faster I can identify the problem and fix it.

burchill:
Is this better?

```r
library(dplyr)
library(tidyr)
# Package versions:
# dplyr_0.5.0     tidyr_0.6.1 

d <- data.frame(y=c("a,b")) %>% 
  dplyr::tbl_df()
d$reg_list <- list("c")
d
# A tibble: 1 × 2
#           y  reg_list
#   <fctr>    <list>
# 1    a,b <chr [1]>
tidyr::separate_rows(d,y,sep=",")
# A tibble: 2 × 1
#       y
#   <chr>
[...]

PR #304, opened 2017-05-17
Add drop argument to separate_rows()

aaronwolen:
This addresses #300 by exposing `unnest()`'s `.drop` argument. Note, I used `drop` rather than `.drop` to be consistent with `separate()`, which hasn't adopted the `.argument` argument style.


hadley:
Thanks! Can you please merge/rebase and add a bullet to NEWS?

aaronwolen:
Ah, looks like this was already addressed in #321.

PR #307, opened 2017-05-26
duplicated word removed

rmsharp:
Just in case: “I assign the copyright of this contribution to Hadley Wickham”


hadley:
Thanks!

PR #312, opened 2017-06-19
Fix link format in vignette

thomasp85:
The internal link to #multiple-types had square and round braces reversed leading to broken format...


hadley:
Thanks!

PR #319, opened 2017-06-25
Fix #254.

### LINKED ISSUE #254: spread() with drop = FALSE applied to var column including NA values results an error
Below is an example data.frame.

``` R
> (df_with_na <- data.frame(var = factor(c(LETTERS[1:3], NA), LETTERS[1:4]), val = seq(4)))
   var val
1    A   1
2    B   2
3    C   3
4 <NA>   4
> df_with_na$var
[1] A    B    C    <NA>
Levels: A B C D
```

When I spread this df by the "var" column, call like 

``` R
> tidyr::spread(df_with_na, var, val)
  A B C <NA>
1 1 2 3    4
```

In this case, I got an column "<NA>" originate from the row that contains NA in "var" column.
Then I added "drop = FALSE" option to keep all levels (including "D"), and got this:

``` R
> tidyr::spread(df_with_na, var, val, drop = FALSE)
Error in `colnames<-`(`*tmp*`, value = c("A", "B", "C", "D", "<NA>")) : 
  length of 'dimnames' [2] not equal to array extent
```

I think this is not an intended behavior and it'd better to return a data.frame with columns for all levels of "var" column and  NA.

ngr-t:
This is a draft PR to fix #254 .
It seems that the problem is caused by that the `n` attribute in the return value of `id_var()` is different from the expected value when it's called with `x` including `NA` and `drop = FALSE`. It doesn't count `NA`.
This leads to wrong code path in if clause at https://github.com/tidyverse/tidyr/blob/master/R/spread.R#L100. I think we have to enter L101 if we have missing values in `key`, but we enter `else` clause now (if `drop = FALSE` is given).

I modified `id()` to count `NA` when `drop = FALSE` is given.

Thanks.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,6 +15,7 @@
   unquote symbols or strings). This type of NSE is now strongly
   discouraged in the tidyverse; symbols should represent real objects.
 
+* Fixed that `spread()` fails when the `key` column includes `NA` and `drop` is `FALSE` (#254).
 
 # tidyr 0.6.3
 
--- a/R/id.R
+++ b/R/id.R
@@ -41,8 +41,9 @@ id_var <- function(x, drop = FALSE) {
   if (!is_null(attr(x, "n")) && !drop) return(x)
 
   if (is.factor(x) && !drop) {
-    id <- as.integer(addNA(x, ifany = TRUE))
-    n <- length(levels(x))
+    x_na <- addNA(x, ifany = TRUE)
+    id <- as.integer(x_na)
+    n <- length(levels(x_na))
   } else if (length(x) == 0) {
     id <- integer()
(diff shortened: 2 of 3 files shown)


hadley:
Looks good! Can you please add a bullet point to NEWS?

ngr-t:
I've added an item in the `tidyr 0.6.3.9000` section of `NEWS.md`. I'm not an English native so I'm concerned whether the meaning is clear...

hadley:
Thanks. I'll merge after we release this version of tidyr (we want to do a release just for tidyeval support)

hadley:
Thanks!

ngr-t:
Thanks! Glad to contribute this great package!

Issue #308, opened 2017-05-31
separate_rows does not work with empty data frame


bheavner:
Like issue #253 - separate_rows() fails on empty data frames:
```
data.frame(a=character(), b=character()) %>% separate_rows(b)
Error in rep(1:nrow(data), n[[1]]) : invalid 'times' argument
```

duttashi:
The first part of the code, `data.frame(a=character(), b=character())` will create an empty data frame whose structure is ;
'data.frame':	0 obs. of  2 variables:
 $ a: Factor w/ 0 levels: 
 $ b: Factor w/ 0 levels: 

As per the help file for `> ?separate_rows()`, "If a variable contains observations with multiple delimited values, this separates the values and places each one in its own row.". 

So, this indicates the error is correct.  The `separate_rows()` function will not work because it requires "variable with multiple values", and, you are passing it a variable with no values. Thus, I think the error is correct.

hadley:
This now gives a somewhat better error:

```R
df <- data.frame(a = character(), b = character())
df %>% separate_rows(b)
#> Error: Strings must match column names. Unknown columns: b
```

which occurs because of what looks like a `unnest()` bug:

```R
df %>% unnest(b)
#> # A tibble: 0 x 1
#> # ... with 1 variables: a <chr>
```

hadley:
I've made that a separate issue #378

PR #320, opened 2017-06-26
Fix corner case for unnest

### LINKED ISSUE #253: unnest does not work with empty data frame
```R
library(tidyr)
tibble::tibble(a = character(), b = list()) %>% unnest()
#> Error in rep(1:nrow(data), n[[1]]): invalid 'times' argument
```

krlmlr:
Closes #253.

For #308, more work is needed, because this PR loses columns of empty data frames when calling `separate_rows()`.


hadley:
Thanks!

PR #334, opened 2017-07-28
Release 0.6.4 (WIP)

lionel-:
Should fix many revdeps issues. Having a bit of trouble with revdeps package.


hadley:
I'm running the revdeps now. I'll check in once they're done

PR #336, opened 2017-08-01
Release fix revdeps

lionel-:
- the `eurostats` package is now fixed

- the objects and not found functions errors are now included in the breaking changes sections. Also the `...` arguments of selecting functions have now better documentation.


NEWS.md
```diff
@@ -9,6 +9,35 @@ backend.
 
 ## Breaking changes
 
+- The selecting functions (e.g. `gather()` and its `...` argument)
```
hadley: I think you need to lead with the error message, i.e. " If you see error messages about objects or functions not found, it is because the select functions ..."

And I think you could make it a bit more accessible by starting with something like:

What does the following code do?

```R
x <- 3
df <- tibble(w = 1, x = 2, y = 3)
select(df, 1:x)
```

Does it select the first three columns (using the `x` defined in the global environment), or does it select the first two columns (using the column named `x`)? In this version of tidy we have resolved the ambiguity, now ...

R/gather.R
```diff
@@ -14,9 +14,26 @@
...
 #'   symbols do not represent actual objects is now discouraged in the
 #'   tidyverse; we support it here for backward compatibility).
-#' @param ... Specification of columns to gather. Use bare variable names.
-#'   Select all variables between x and z with `x:z`, exclude y with
-#'   `-y`. For more options, see the [dplyr::select()] documentation.
+#' @param ... A selection of columns. These arguments are passed to
+#'   [tidyselect::vars_select()] and are treated specially for
+#'   selection:
+#'
+#'     * Calls and complex expressions are evaluated in the context.
```
hadley: This gets too technical too fast. I think it would be better to keep the basics as before, and move the details to a new section.

lionel-:
@hadley could you check the new wordings please?

PR #258, opened 2016-11-16
Add replace_na.default for vectors

paulponcet:
Add replace_na.default for replacement of missing values in (numeric, character, logical) vectors.

PR #342, opened 2017-08-10
Add replace_na.default for vectors (alternative approach)

flying-sheep:
I saw #258, but didn’t like the idea of using a single-element list.

This version is natural, but not type-stable. (`replace` is expected to be a list when `data` is a data.frame, else a vector)

There’s another approach, which is removing “replace” and using `...` (backwards-compatible):

```r
replace_na.default <- function(data, ...) {
  args <- list(...)
  stopifnot(length(args) == 1L)
  stopifnot(is_null(names(args)))
  replace <- args[[1L]]
  …
}
replace_na.data.frame <- function(data, ...) {
  replace <- list(...)
  if (length(replace) == 1L && is_list(replace[[1L]]))
    replace <- replace[[1L]]  # old signature
  …
}
```

Please tell me what you’d rather like to see.


hadley:
It's been a while since you submitted so I wanted to check before starting the review process: are you interested in finishing this off?

flying-sheep:
Sure, let’s go! I might not respond to comments before sunday though.

R/replace_na.R
```diff
@@ -1,19 +1,31 @@
...
 #' @export
-replace_na <- function(data, replace = list(), ...) {
+replace_na <- function(data, replace = NULL, ...) {
   UseMethod("replace_na")
 }
 #' @export
-replace_na.data.frame <- function(data, replace = list(), ...) {
+replace_na.default <- function(data, replace = NULL, ...) {
+  if (is_null(replace))
+    stop("If 'data' is a vector, 'replace' has to be provided.")
```
hadley: Why?

I think instead it would be better to assert that `replace` has length 1

flying-sheep: Well, the two things that that can happen are the following. Pretty clear message:

```r
> data[1:5] <- NULL
Error in data[1:5] <- NULL: replacement has length zero
> data[1:5] <- 1:3
Warning message in data[1:5] <- 1:3:
“number of items to replace is not a multiple of replacement length”
```

If this now happens through our function here, we get:

```r
> replace_na(data)
Error in data[are_na(data)] <- replace: replacement has length zero
> replace_na(data, 1:3)
Warning message in data[are_na(data)] <- replace:
“number of items to replace is not a multiple of replacement length”
```

Of which I only think the last one is clear.

hadley: But this might also happen: `data[1:6] <- 1:3`. We want to avoid accidentally recycling the `replace` value

flying-sheep: OK, done!

R/replace_na.R
```diff
@@ -1,19 +1,31 @@
...
+replace_na <- function(data, replace = NULL, ...) {
   UseMethod("replace_na")
 }
 #' @export
-replace_na.data.frame <- function(data, replace = list(), ...) {
+replace_na.default <- function(data, replace = NULL, ...) {
+  if (length(replace) != 1L)
+    stop("If 'data' is a vector, 'replace' has to be of length 1, not ", length(replace))
+
+  data[are_na(data)] <- replace
```
flying-sheep: maybe I should use `replace[[1L]]` here to allow people to call `replace_na(vec, list(5))`?

hadley: That's a good idea! It's more consistent with the rest of the API. (But you'll need to check that `replace()` is a length 1 list)

flying-sheep: We require it to be length 1 anyway, so no matter if list or not, `replace[[1L]]` will always be correct, no?

hadley: I think it's better to be strict about the types here

flying-sheep: OK, done!

hadley:
I ended up going in a slightly different direction; as well as adding a better error message for the data frame case, and adding support for list-cols. Sorry I didn't merge your PR. I really appreciate the effort 😁

flying-sheep:
no worries!

PR #348, opened 2017-08-22
Update website to incorporate  latest release.

maurolepore:
Thanks for this version! Just to let you know that it seems that __tidyr__'s website is outdated.


hadley:
Thanks - I've moved this to an issue #376

PR #362, opened 2017-09-12
Fix use of STRING_PTR() on R devel

lionel-:
`STRING_PTR()` is used in the melting code to access elements of a list. This is no longer allowed on R devel: https://github.com/wch/r-source/commit/6ff8e95674c0437d3de4edc04276327f87c64ca8

This should fix it. cc @kevinushey


kevinushey:
Thanks!

I'll miss you, lazy inappropriate `STRING_PTR` hack.

PR #368, opened 2017-10-21
Use purrr::map_dfr() instead of plyr::ldply() in tidy-data vignette

yutannihilation:
I think purrr package is preferable to describe about the concept of "tidy" as it is the member of tidyverse.


hadley:
Thanks!

PR #380, opened 2017-11-17
Fix handling of negative integer values in strsep

### LINKED ISSUE #315: separate() right edge fixed position splitting is off by 1
According to the documentation for the `sep` argument:

> If numeric, interpreted as positions to split at. Positive values start at 1 at the far-left of the string; negative value start at -1 at the far-right of the string.

However, to split the right-most character into a new column, `sep`must be set to -2. Reproducible example:

```
library(tidyverse)

df <- expand.grid(group = c("alpha", "beta"),
                  dimension = c(1, 2),
                  stringsAsFactors = F) %>%
        tbl_df() %>%
        mutate(gr_dim = paste0(group, dimension))


# splits the one left most character
df %>%
  separate(gr_dim, sep = 1, into = c("gr", "dim"))

# creates two columns, one of which is the original column, the other blank
df %>%
  separate(gr_dim, sep = -1, into = c("gr", "dim"))

# splits the two left-most characters
df %>%
  separate(gr_dim, sep = 2, into = c("gr", "dim"))

# splits the one right-most character
df %>%
  separate(gr_dim, sep = -2, into = c("gr", "dim"))
```

markdly:
This is a draft PR to Fix #315. All values passed to `stringi::stri_sub` in the `strsep` function are now non-negative as I also noticed a related issue where the use of relatively large negative sep values (relative to the string length) could cause unexpected results.  I'll add a note about this over in the issues section...

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -31,7 +31,8 @@
 
 * Fixed that `spread()` fails when the `key` column includes `NA` and `drop` is `FALSE` (#254).
 
-* `separate()` now works as described in documentation when using negative integer values with `sep` argument (#315).
+* `separate()` now works as described in documentation when using negative
+  integer values with `sep` argument (@markdly, #315).
 
 # tidyr 0.7.2
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -31,6 +31,8 @@
 
 * Fixed that `spread()` fails when the `key` column includes `NA` and `drop` is `FALSE` (#254).
 
+* `separate()` now works as described in documentation when using negative integer values with `sep` argument (#315).
+
 # tidyr 0.7.2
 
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -31,6 +31,8 @@
 
 * Fixed that `spread()` fails when the `key` column includes `NA` and `drop` is `FALSE` (#254).
 
+* `separate()` now works as described in documentation when using negative integer values with `sep` argument (#315).
```
hadley: Can you please credit yourself with `(@yourname, #issuenumber)`?  And break the line to be <80 characters?

hadley:
Thanks!

markdly:
Thanks for making this package (and all the others) available in the first place!

Issue #328, opened 2017-07-17
Allow unnest with list columns of differing lengths?


karldw:
`unnest` currently can't handle multiple list columns with different lengths.  If the user requests an unnesting of one list column from a dataframe with multiple, `unnest` will fail if the number of elements differs. Would it be possible for `unnest` to copy the other list columns, just as it copies values from standard, atomic columns?

In particular, the current behavior means `unnest` doesn't work with `sf` data, since the `geometry` column is already a list column. 

```r

library(sf)
library(dplyr)
library(tidyr)
nc <- st_read(system.file("shape/nc.shp", package = "sf")) %>% 
  slice(1:3) %>%
  select(NAME) %>%
  mutate(y = strsplit(c("a", "d,e,f", "g,h"), ","))

nc
#> Simple feature collection with 3 features and 2 fields
#> geometry type:  MULTIPOLYGON
#> dimension:      XY
#> bbox:           xmin: -84.32385 ymin: 33.88199 xmax: -75.45698 ymax: 36.58965
#> epsg (SRID):    4267
#> proj4string:    +proj=longlat +datum=NAD27 +no_defs
#> # A tibble: 3 x 3
#>        NAME         y          geometry
#>      <fctr>    <list>  <simple_feature>
#> 1      Ashe <chr [1]> <MULTIPOLYGON...>
#> 2 Alleghany <chr [3]> <MULTIPOLYGON...>
#> 3     Surry <chr [2]> <MULTIPOLYGON...>


# Current behavior:
unnest(nc, y, .drop = FALSE)
#>  Error: All nested columns must have the same number of elements.


# Expected behavior: values in geometry column copied for the newly-created rows
unnest(nc, y, .drop = FALSE)
#> # A tibble: 6 x 2
#>        NAME     y           geometry
#>      <fctr> <chr>   <simple_feature>
#> 1      Ashe     a  <MULTIPOLYGON...>
#> 2 Alleghany     d  <MULTIPOLYGON...>
#> 3 Alleghany     e  <MULTIPOLYGON...>
#> 4 Alleghany     f  <MULTIPOLYGON...>
#> 5     Surry     g  <MULTIPOLYGON...>
#> 6     Surry     h  <MULTIPOLYGON...>
```

Ref: https://github.com/r-spatial/sf/issues/426

hadley:
Oooh good idea!

hadley:
I think this will need some extra syntax, maybe something like `unnest(df, y, preserve = x)`

[...]

PR #383, opened 2017-11-20
Add failing test for unnest with .preserve

karldw:
Add test for comment in #328


karldw:
I made this a PR to (hopefully) make life easier, but feel free to reject if you can add this test more easily yourself.

hadley:
I probably won't take as is, but copying and pasting it will be a great start and it's super helpful. Thanks!

Issue #317, opened 2017-06-23
A feature request for adding margin when spreading


therimalaya:
I would help in lot of situation if we can have an option to add margin when spreading data with `spread` function.

hadley:
That's out of scope for tidyr, sorry.

PR #392, opened 2018-01-06
styling with styler

lorenzwalthert:
conservative styling with scope = "spaces" and strict = FALSE.

### DIFF
--- a/R/separate-rows.R
+++ b/R/separate-rows.R
@@ -35,7 +35,7 @@ separate_rows.data.frame <- function(data, ..., sep = "[^[:alnum:].]+",
 
   data[vars] <- map(data[vars], stringi::stri_split_regex, sep)
   data <- unnest(data, !!! syms(vars), .drop = FALSE)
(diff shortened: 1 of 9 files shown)


lorenzwalthert:
Reference: https://github.com/krlmlr/tidyr/pull/1
cc: @krlmlr

hadley:
Thanks!

lorenzwalthert:
Should we do a follow-up with `scope > "spaces"`?

hadley:
Yeah, let's take a look. Does styler have the ability to ignore `tribble()` spacing yet?

lorenzwalthert:
Ok, cool. No, it has not. Maybe we should add that. We can use `strict = FALSE`, which just ensures *at least* one space around certain operators. That should leave many tribble calls  untouched.

lorenzwalthert:
I opened #317 to discuss that further (and labeled it with `Status: Postponed` and closed it for now because me and @krlmlr don't really have time to work on it now).

PR #393, opened 2018-01-06
styling comprehensively with styler

lorenzwalthert:
First styled with `strict = FALSE` and `scope = "tokens"`. 
Being more invasive than this configuration (e.g. with`strict = TRUE`) did expose one issue with `styler@HEAD`: lines breaks in function calls are not taking into account whether arguments are named or not (https://github.com/tidyverse/style/issues/39). 
I attempted resolve that issue in r-lib/styler#318 and used this version to style with `scope = "tokens"`, `strict = TRUE`. The `enframe_data()` call in `data-raw/smiths.R` was excluded from styling since it needed tribble-like alignment (https://github.com/r-lib/styler/issues/317).

### DIFF
--- a/R/expand.R
+++ b/R/expand.R
@@ -138,7 +138,6 @@ crossing <- function(...) {
       "Each element must be either an atomic vector or a data frame.
        Problems: {problems}."
     ))
-
   }
 
   # turn each atomic vector into single column data frame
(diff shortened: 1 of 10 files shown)


hadley:
Thanks!

PR #400, opened 2018-01-29
now to no

wibeasley:
I'm pretty sure it's "no~~w~~ longer checks" instead of "now ~~longer~~ checks"
https://github.com/tidyverse/tidyr/commit/d2669af7c71ccbe6661de1f2b31e120835ebd3d4#diff-5fe5f808344c27d942e0c448a16fe07c


hadley:
Thanks!

PR #402, opened 2018-02-01
Add mention of `extract` in `separate` docs

markdly:
This is a suggestion to help improve the discoverability of `extract` by adding an @seealso entry to `separate`.  

I thought this might be worth suggesting after responding to a question on the RStudio community site where a user was aware of `separate` but not `extract` (see https://community.rstudio.com/t/getting-parts-of-string-using-regex-and-capturing-parens-in-mutate). This was very similar to my own experience before discovering `extract` existed.

I also thought this might be helpful as `extract` is not currently included on the RStudio `tidyr` cheatsheet while `separate` is.

### DIFF
--- a/R/separate.R
+++ b/R/separate.R
@@ -32,7 +32,8 @@
 #'   `as.is = TRUE` on new columns. This is useful if the component
 #'   columns are integer, numeric or logical.
 #' @param ... Additional arguments passed on to methods.
-#' @seealso [unite()], the complement.
+#' @seealso [unite()], the complement, [extract()] which uses regular
+#'   expression capturing groups.
 #' @export
 #' @examples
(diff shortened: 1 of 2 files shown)


jameshowison:
Good plan.  FWIW `extract` is also not mentioned in the tidy data vignette at http://tidyr.tidyverse.org/articles/tidy-data.html  (`extract_numeric` is, although looks like that is deprecated in favor of `readr::parse_number` ...). Also doesn't show up in http://r4ds.had.co.nz/tidy-data.html chapter (well, it does in one spot, an exercise asking to compare `separate` and `extract` but not elsewhere.

hadley:
Thanks!

PR #403, opened 2018-02-01
Typo in unnest docs

markdly:
In the docs for `unnest` I thought 

`@param ... Specification of columns to nest.`

should instead be:

`@param ... Specification of columns to unnest.`


hadley:
Thanks!

For future reference, there's no need to describe your precise change in text - I can see it in the diff.

PR #405, opened 2018-02-02
Update drop-na.R

chris-park:
Intuitively, I would have thought `is_complete(NULL)` should return `FALSE`, but the call results in the following error: `"In is.na(x) : is.na() applied to non-(list or vector) of type 'NULL'".` I added an extra branch (instead of say `x_type %in% c("list", "NULL")`) since a call to `vapply()` with `NULL` as its main argument returns a zero-length logical vector.

### DIFF
--- a/R/drop-na.R
+++ b/R/drop-na.R
@@ -44,9 +44,13 @@ complete_cases <- function(x, fun) {
 }
 
 is_complete <- function(x) {
-  if (typeof(x) == "list") {
+  x_type <- typeof(x)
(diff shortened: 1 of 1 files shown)


hadley:
What are you trying to achieve here?

chris-park:
I thought it might affect some obscure edge cases, but I guess you can't have data frames with `NULL` elements (I was tinkering around with functions for dropping `NA`s from vectors in general when I came across this code). Sorry about that!

Issue #145, opened 2015-12-29
unnest() should preserve column order


hadley:
i.e.

```R
dplyr::data_frame(x = 1:2, y = list(3, 4), z = list(5, 6:7)) %>% 
  unnest(y)
```

hadley:
Given that the semantics of unnest are mutate-like, not select-like, I don't think this is possible without a lot of work.

PR #406, opened 2018-02-05
Preserve column order after unnesting

aaronwolen:
I run into this issue frequently enough that I thought I'd take a crack at it. The proposed change is pretty straightforward, so much so that I wonder if I'm overlooking something—especially in light of #145's conclusion. But I haven't found a case where it breaks down and all tests pass after adjusting a few that assumed the columns would be reordered.

### DIFF
--- a/R/unnest.R
+++ b/R/unnest.R
@@ -157,11 +157,20 @@ unnest.data.frame <- function(data, ..., .drop = NA, .id = NULL,
   rest <- data[rep(seq_nrow(data), n[[1]]), group_vars, drop = FALSE]
   out <- dplyr::bind_cols(rest, unnested_atomic, unnested_dataframe)
 
-  # Preserve column order
-  all_vars <- intersect(names(data), names(out))
-  all_vars <- c(all_vars, setdiff(names(out), all_vars))
+  # Restore column order
+  nested_vars <- imap(data, function(.x, .y) {
+    if (is_empty(.x)) {
+      return(.y)
+    } else if (is.data.frame(.x[[1]])) {
+      return(names(.x[[1]]))
+    } else {
+      return(.y)
+    }
+  })
+  ordered_vars <- intersect(unlist(nested_vars, use.names = FALSE), names(out))
+  ordered_vars <- c(ordered_vars, setdiff(names(out), ordered_vars))
 
-  reconstruct_tibble(data, out[all_vars])
+  reconstruct_tibble(data, out[ordered_vars])
 }
 
 list_col_type <- function(x) {
--- a/R/unnest.R
+++ b/R/unnest.R
@@ -157,7 +157,11 @@ unnest.data.frame <- function(data, ..., .drop = NA, .id = NULL,
   rest <- data[rep(seq_nrow(data), n[[1]]), group_vars, drop = FALSE]
   out <- dplyr::bind_cols(rest, unnested_atomic, unnested_dataframe)
 
-  reconstruct_tibble(data, out)
+  # Preserve column order
+  all_vars <- intersect(names(data), names(out))
+  all_vars <- c(all_vars, setdiff(names(out), all_vars))
+
+  reconstruct_tibble(data, out[all_vars])
 }
 
 list_col_type <- function(x) {
--- a/tests/testthat/test-unnest.R
+++ b/tests/testthat/test-unnest.R
@@ -196,6 +196,6 @@ test_that("column order is preserved", {
   expect_named(unnest(df), c("x", "y", "z"))
 
   df$y <- list(tibble(y1 = "b", y2 = "e"), tibble(y1 = "c", y2 = "f"))
-  expect_named(unnest(df), c("x", "y", "z"))
+  expect_named(unnest(df), c("x", "y1", "y2", "z"))
 })
 
--- a/tests/testthat/test-unnest.R
+++ b/tests/testthat/test-unnest.R
@@ -195,7 +195,7 @@ test_that("column order is preserved", {
   df <- tibble(x = "a", y = list("b", "c"), z = "d")
   expect_named(unnest(df), c("x", "y", "z"))
 
-  df$y <- list(tibble(y = "b"), tibble(y = "c"))
+  df$y <- list(tibble(y1 = "b", y2 = "e"), tibble(y1 = "c", y2 = "f"))
   expect_named(unnest(df), c("x", "y", "z"))
 })
 
--- a/tests/testthat/test-unnest.R
+++ b/tests/testthat/test-unnest.R
@@ -190,3 +190,12 @@ test_that("unnest() recognize ptype", {
   res <- unnest(tbl)
   expect_equal(res, tibble(x = integer(), y = double()))
 })
+
+test_that("column order is preserved", {
+  df <- tibble(x = "a", y = list("b", "c"), z = "d")
+  expect_named(unnest(df), c("x", "y", "z"))
+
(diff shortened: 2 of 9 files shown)


tests/testthat/test-unnest.R
```diff
@@ -162,3 +162,8 @@ test_that("grouping is preserved", {
   expect_equal(class(df), class(rs))
   expect_equal(dplyr::groups(df), dplyr::groups(rs))
 })
+
+test_that("column order is preserved", {
+  df <- tibble(x = "a", y = list("b", "c"), z = "d")
+  expect_equal(df %>% unnest() %>% names(), c("x", "y", "z"))
```
hadley: You can use `expect_named()` here

tests/testthat/test-unnest.R
```diff
@@ -162,3 +162,8 @@ test_that("grouping is preserved", {
   expect_equal(class(df), class(rs))
   expect_equal(dplyr::groups(df), dplyr::groups(rs))
 })
+
+test_that("column order is preserved", {
```
hadley: Could you add another test for the case where you are unnesting a list of data frames?

aaronwolen:
Cool, I just implemented your suggestions and updated the NEWS.

aaronwolen:
Also updated the existing tests to use `expect_named()` where appropriate and fixed a typo. I thought a single PR would be easier for you but I'm happy to split these up if that's preferable.

tests/testthat/test-unnest.R
```diff
@@ -167,3 +167,11 @@ test_that("grouping is preserved", {
...
+  df <- tibble(x = "a", y = list("b", "c"), z = "d")
+  expect_named(unnest(df), c("x", "y", "z"))
+
+  df$y <- list(tibble(y = "b"), tibble(y = "c"))
```
krlmlr: The test would be clearer if the internal tibbles had two columns with different names than `y` .

aaronwolen: I see what you mean. What do you think about:

```r
df$y <- list(tibble(y = "b", y2 = "e"), tibble(y = "c", y2 = "f"))
```

This verifies the order of the original columns (`x`, `y`, and `z`) is preserved and new columns from the internal tibbles (`y2`) are appended.

```r
unnest(df)
## # A tibble: 2 x 4
##   x     y     z     y2   
##   <chr> <chr> <chr> <chr>
## 1 a     b     d     e    
## 2 a     c     d     f 
```

krlmlr: I'd have expected the new column order to be x, y, y2, z, same if y isn't part of the nested tibble. What's the rationale behind leaving only y in the original place?

hadley:
@aaronwolen would you be interested in finishing this off for the next version of tidyr? If I don't hear back from you in a week, I'll assume that you're busy with other stuff, and I'll take it over the finish line.

aaronwolen:
Absolutely. I'll incorporate @krlmlr's suggestion re column ordering and push an update this week.

aaronwolen:
Accounting for the order of variables within nested data frames made this fix _slightly_ more complicated 😬

tests/testthat/test-unnest.R
```diff
@@ -154,21 +154,21 @@ test_that("unnest drops list cols if expanding", {
...
-  expect_equal(names(out), c("x", "z"))
+  expect_named(out, c("x", "z"))
 })
 
 test_that("unnest keeps list cols if not expanding", {
   df <- tibble(x = 1:2, y = list(3, 4), z = list(5, 6:7))
   out <- df %>% unnest(y)
 
-  expect_equal(names(out), c("x", "z", "y"))
+  expect_equal(class(out$z), "list")
```
hadley: I think you should preserve the original test here.

tests/testthat/test-unnest.R
```diff
@@ -154,21 +154,21 @@ test_that("unnest drops list cols if expanding", {
...
 })
 
 test_that("unnest respects .drop_lists", {
   df <- tibble(x = 1:2, y = list(3, 4), z = list(5, 6:7))
 
-  expect_equal(df %>% unnest(y, .drop = TRUE) %>% names(), c("x", "y"))
-  expect_equal(df %>% unnest(z, .drop = FALSE) %>% names(), c("x", "y", "z"))
+  expect_named(df %>% unnest(y, .drop = TRUE), c("x", "y"))
```
hadley: Since there's now only a single `%>%`, I think you should just get rid of it too.

aaronwolen:
Thanks! I just included your suggestions and updated the news. 

(Loving the new pivot fxns, btw!)

R/unnest.R
```diff
@@ -157,7 +157,20 @@ unnest.data.frame <- function(data, ..., .drop = NA, .id = NULL,
   rest <- data[rep(seq_nrow(data), n[[1]]), group_vars, drop = FALSE]
   out <- dplyr::bind_cols(rest, unnested_atomic, unnested_dataframe)
 
-  reconstruct_tibble(data, out)
+  # Restore column order
```
hadley: It just occurred to me to wonder if it might be simpler to this a bit further up, i.e.:

```R
out <- c(rest, unnested_atomic, unnested_dataframe)
# fix order
out <- dplyr::bind_cols(out)
reconstruct_tibble(data, out)
```

aaronwolen: We can capture the original order/structure of column names immediately, so the `nested_vars` object could be created much earlier in the function. But the actual reordering requires the output of `bind_cols(rest, unnested_atomic, unnested_dataframe)`, so I'm not sure that bit can be relocated.

aaronwolen: Oh... 💡!

You meant reconstruct the tibble first and *then* take care of reordering the columns!

Something like this?

```r
  rest <- data[rep(seq_nrow(data), n[[1]]), group_vars, drop = FALSE]
  out <- reconstruct_tibble(
    data,
    dplyr::bind_cols(rest, unnested_atomic, unnested_dataframe)
  )

  # Restore column order
  nested_vars <- imap(data, function(.x, .y) {
    if (is_empty(.x)) {
      return(.y)
    } else if (is.data.frame(.x[[1]])) {
      return(names(.x[[1]]))
    } else {
      return(.y)
    }
  })
  ordered_vars <- intersect(unlist(nested_vars, use.names = FALSE), names(out))
  out[c(ordered_vars, setdiff(names(out), ordered_vars))]
```

R/unnest.R
```diff
@@ -157,7 +157,20 @@ unnest.data.frame <- function(data, ..., .drop = NA, .id = NULL,
...
+  nested_vars <- imap(data, function(.x, .y) {
+    if (is_empty(.x)) {
+      return(.y)
+    } else if (is.data.frame(.x[[1]])) {
+      return(names(.x[[1]]))
+    } else {
+      return(.y)
+    }
+  })
```
aaronwolen: I feel like this approach is easier to follow. Thoughts?

```r
  nested_vars <- imap(data, function(.x, .y) {
    if (is_empty(.x)) {
      return(.y)
    }
    .x <- .x[[1]]
    if (is.data.frame(.x)) {
      names(.x)
    } else {
      .y
    }
  })
```

R/unnest.R
hadley: Oh yeah, maybe you're right and I'm confused about this structure.

R/unnest.R
aaronwolen: Not at all! You gotta do what's best for the package and I'll just be happy to see the issue fixed. Either way, I'm always happy to help out where I can.

PR #411, opened 2018-02-13
fixing typos

markdly:


README.Rmd
```diff
@@ -53,12 +53,12 @@ There are two fundamental verbs of data tidying:
 * `gather()` takes multiple columns, and gathers them into key-value pairs: it 
   makes "wide" data longer.
 
-* `spread()`. takes two columns (key & value) and spreads in to multiple 
-  columns, it makes "long" data wider. 
+* `spread()` takes two columns (key & value), and spreads across multiple
```
batpigandme: Thanks for these. My one suggestion would be to use "into" (as one word, of course) instead of across; for some reason the latter sounds like it's overwriting something to me. That's just my 2¢, though!

markdly: Yes I think 'into' is better too. I've updated the PR. I was 'on the fence' about using 'access' but saw it was used in the help description for spread `?spread`:

> Spread a key-value pair across multiple columns.

batpigandme: Ah, that makes sense. As a fan of internal consistency, I'll leave this one up to @hadley… I honestly have no rationale for _why_ it sounds weird to me! ¯\\(°_o)/¯

hadley:
Thanks!

PR #416, opened 2018-02-15
Add missing backtick + fix spelling error

batpigandme:
Presumes `opperation` is _not_ an alternate spelling that I can't find. 😉

PR #417, opened 2018-02-15
fix typo in separate

markdly:


markdly:
Sorry! I made the PR before seeing #416 includes an update to `unncount.rd` making my change obsolete. Have attempted to remove this. I think the fix for the typo is still needed though...

batpigandme:
Thanks, @markdly!

@jennybc no rush, but just wanted to double check with you that merging #416 & then this one would !⇨ git borking… 😰 (I don't think it should)

hadley:
Thanks!

PR #420, opened 2018-02-18
Use the word "kind" instead of "type"

jjchern:
The word "type" has a special meaning in the vignette: it means a level of observation. I thus propose using a different word when referred to a category of things.


batpigandme:
I think that in at least one of these instances there is sufficient semantic distance as to avoid referential ambiguity without changing the word. 
e.g. `An example of this type of cleaning...` has the necessary syntactic clues to, in effect, bind "type" to "cleaning."
https://github.com/tidyverse/tidyr/pull/420/files#diff-d233317393e3292d36fa8ce49f742e5fL291

PR #445, opened 2018-04-18
Omit columns named NA from separate result

### LINKED ISSUE #397: In separate an into= column name can be NA to not generate that column
In the example below, modified from https://stackoverflow.com/questions/48255809/to-change-tabular-data-to-a-different-format-in-r/48256012#48256012 , the `X` column is a junk column that just contains the second underscore delimiter for those rows that have it.  If the `separate` statement would allow `NA` or `""`, say, as a column name with the meaning _do not generate a column for that component_ then the `select(-X)` could have been omitted and the `separate` statement would become `separate(variable, c("CurveGroup", NA, "suffix"), sep = 5:6, fill = "right")` .  Not only would this reduce the code by one line but it would also eliminate the need to generate a junk name for a column that is subsequently dropped.

```
DF <- structure(list(pos = 1:2, BZ_SP = c(-300000L, 0L), BZ_SP_m1 = c(2L, 
  0L), BZ_SP_m2 = c(3L, 0L), CL_SP = c(2540544L, -118621L), CL_SP_m1 = c(1L, 
  3L), CL_SP_m2 = c(2L, 4L)), .Names = c("pos", "BZ_SP", "BZ_SP_m1", 
  "BZ_SP_m2", "CL_SP", "CL_SP_m1", "CL_SP_m2"), 
  class = "data.frame", row.names = c(NA, -2L))

DF
##   pos   BZ_SP BZ_SP_m1 BZ_SP_m2   CL_SP CL_SP_m1 CL_SP_m2
## 1   1 -300000        2        3 2540544        1        2
## 2   2       0        0        0 -118621        3        4

library(dplyr)
library(tidyr)

DF %>%
   gather(variable, value, -pos) %>% 
   separate(variable, c("CurveGroup", "X", "suffix"), sep = 5:6, fill = "right") %>% 
   select(-X) %>%
   spread(suffix, value)
```

giving:

```
  pos CurveGroup      V1 m1 m2
1   1      BZ_SP -300000  2  3
2   1      CL_SP 2540544  1  2
3   2      BZ_SP       0  0  0
4   2      CL_SP -118621  3  4

```

markdly:
This is a draft PR to fix #397 which addresses the main issue of being able to use `NA` with the `into` argument in `separate` to specify columns which should be omitted / not included in the result. 

For the moment I haven't added anything re an option for trimming white space but would be happy to try if that was considered worthwhile.


hadley:
Thanks! The next release is going to be very small,  but the implementation of this is so straightforward and its a nicer additional feature.

PR #459, opened 2018-06-04
Clean and improve tests of codes in /R -> 100% coverage

rbloehm:
This is a PR that cleans a duplicate test (for gather) and adds further tests to spread (with fill) and id, bringing code coverage of files in /R to 100%.
As it is my first pull request into a public repo, I am more than open for comments on how to improve my process of generating a good pull request. Thanks!

### DIFF
--- a/tests/testthat/test-gather.R
+++ b/tests/testthat/test-gather.R
@@ -95,24 +95,6 @@ test_that("can gather all atomic vectors", {
   df4 <- data.frame(x = 1, y = "a", stringsAsFactors = FALSE)
   df5 <- data.frame(x = 1, y = 1 + 1i, stringsAsFactors = FALSE)
 
-  gathered <- function(val) {
-    data.frame(x = 1, key = "y", val = val, stringsAsFactors = FALSE)
-  }
-
-  expect_equal(gather(df1, key, val, -x), gathered(FALSE))
-  expect_equal(gather(df2, key, val, -x), gathered(1L))
-  expect_equal(gather(df3, key, val, -x), gathered(1))
-  expect_equal(gather(df4, key, val, -x), gathered("a"))
-  expect_equal(gather(df5, key, val, -x), gathered(1 + 1i))
-})
-
-test_that("can gather all atomic vectors", {
-  df1 <- data.frame(x = 1, y = FALSE)
-  df2 <- data.frame(x = 1, y = 1L)
-  df3 <- data.frame(x = 1, y = 1)
-  df4 <- data.frame(x = 1, y = "a", stringsAsFactors = FALSE)
-  df5 <- data.frame(x = 1, y = 1 + 1i, stringsAsFactors = FALSE)
-
   gathered_val <- function(val) {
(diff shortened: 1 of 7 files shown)


tests/testthat/test-gather.R
```diff
@@ -88,24 +88,6 @@ test_that("overwrites existing vars", {
 
 # Column types ------------------------------------------------------------
 
-test_that("can gather all atomic vectors", {
```
hadley: Why did you remove these tests?

tests/testthat/test-spread.R
```diff
@@ -2,14 +2,14 @@ context("Spread")
 library(dplyr, warn.conflicts = FALSE)
 
 test_that("order doesn't matter", {
-  df1 <- data.frame(x = c("a", "b"), y = 1:2)
-  df2 <- data.frame(x = c("b", "a"), y = 2:1)
+  df1 <- tibble(x = c("a", "b"), y = 1:2)
```
hadley: Changing from `data.frame()` to `tibble()` subtly changes the test because `tibble()` doesn't turn strings into factors.

tests/testthat/test-spread.R
```diff
@@ -243,3 +239,21 @@ test_that("spread doesn't convert data frames into tibbles", {
   df <- data.frame(x = c("a", "b"), y = 1:2)
   expect_equal(class(spread(df, x, y)), "data.frame")
 })
+
+test_that("spread with fill replaces explicit missing values", {
+  df <- tibble(key = c("a"), value = c(NA))
```
hadley: Can you please eliminate the needless `c()` here?

tests/testthat/test-underscored.R
```diff
@@ -2,7 +2,7 @@ context("Deprecated SE variants")
 
 test_that("complete_()", {
   df <- tibble(x = 1:2, y = 1:2, z = 3:4)
-  out <- complete_(df, list("x", ~y))
+  out <- complete_(df, list("x", ~ y))
```
hadley: Can you please remove the space? See http://style.tidyverse.org/syntax.html#spacing

tests/testthat/test-gather.R
rbloehm: Because the one below is exactly the same, but does even more. It even has the same name "can gather all atomic vectors".

tests/testthat/test-spread.R
rbloehm: good point, added a factor() to all relevant places in test-spread

tests/testthat/test-spread.R
rbloehm: Sure, also did so in 1 or 2 other places

tests/testthat/test-underscored.R
rbloehm: Done, in 2 places

tests/testthat/test-gather.R
hadley: Oh, got it!

hadley:
Thank you!!

PR #492, opened 2018-09-09
Preserve order of rows in spread() (#453)

### LINKED ISSUE #453: Spread does not preserve order
When going from a wide dataset to long and then back, there seems to be an implicit sort that occurs on non key/value variables. This prevents the ability to recover an initial data set by undoing the `gather()` operation with `spread()`.

Consider the following example:

``` r
library("tidyr")
(init_data = tibble::tibble(
  Group = c("B", "C", "A"),
  `Cost Week 1` = c(1, 2, 3),
  `Cost Week 2` = c(4, 5, 6),
  `Cost Week 3` = c(7, 8, 9)
))
#> # A tibble: 3 x 4
#>   Group `Cost Week 1` `Cost Week 2` `Cost Week 3`
#>   <chr>         <dbl>         <dbl>         <dbl>
#> 1 B                1.            4.            7.
#> 2 C                2.            5.            8.
#> 3 A                3.            6.            9.

long = init_data %>% 
  gather(key = "Week", value = "Cost", -Group)

long
#> # A tibble: 9 x 3
#>   Group Week         Cost
#>   <chr> <chr>       <dbl>
#> 1 B     Cost Week 1    1.
#> 2 C     Cost Week 1    2.
#> 3 A     Cost Week 1    3.
#> 4 B     Cost Week 2    4.
#> 5 C     Cost Week 2    5.
#> 6 A     Cost Week 2    6.
#> 7 B     Cost Week 3    7.
#> 8 C     Cost Week 3    8.
#> 9 A     Cost Week 3    9.

wide = long %>%
  spread(key = "Week", value = "Cost")

wide
#> # A tibble: 3 x 4
#>   Group `Cost Week 1` `Cost Week 2` `Cost Week 3`
#>   <chr>         <dbl>         <dbl>         <dbl>
#> 1 A                3.            6.            9.
#> 2 B                1.            4.            7.
#> 3 C                2.            5.            8.

init_data
#> # A tibble: 3 x 4
#>   Group `Cost Week 1` `Cost Week 2` `Cost Week 3`
#>   <chr>         <dbl>         <dbl>         <dbl>
#> 1 B                1.            4.            7.
#> 2 C                2.            5.            8.
#> 3 A                3.            6.            9.
```

mikmart:
Added an option to `id()` and `id_var()` that allows creating identifiers that do not get ordered in the same way as the original values, but are instead ordered by the order of appearance of unique levels of the original variable. Using this for row identifiers in `spread()` preserves row order, and thus closes #453.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # tidyr 0.8.1.9000
 
+* `spread()` now always preserves row order (@mikmart, #453).
+
 # tidyr 0.8.1
 
 * `unnest()` weakens test of "atomicity" to restore previous behaviour when
--- a/R/id.R
+++ b/R/id.R
@@ -1,4 +1,4 @@
-id <- function(.variables, drop = FALSE) {
+id <- function(.variables, drop = FALSE, match_order = TRUE) {
   if (length(.variables) == 0) {
     n <- nrow(.variables) %||% 0L
     return(structure(seq_len(n), n = n))
@@ -6,11 +6,11 @@ id <- function(.variables, drop = FALSE) {
 
   # Special case for single variable
   if (length(.variables) == 1) {
-    return(id_var(.variables[[1]], drop = drop))
+    return(id_var(.variables[[1]], drop = drop, match_order = match_order))
   }
 
   # Calculate individual ids
-  ids <- rev(map(.variables, id_var, drop = drop))
+  ids <- rev(map(.variables, id_var, drop = drop, match_order = match_order))
   p <- length(ids)
 
   # Calculate dimensions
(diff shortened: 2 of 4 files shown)


hadley:
This changes the API so I'll consider next time I do a bigger release

mikmart:
Okay. Would you consider a fix for making the ordering reordering consistent for now then? As [mentioned in #453](https://github.com/tidyverse/tidyr/issues/453#issuecomment-419712710) currently if the number of unique row id combinations is > 2^31 then the original order is preserved. This is especially a problem when `spreading()` columns with many variables contributing to the row id as the number of combinations grows quickly, and adding a new column could cause the ordering of the result to change.

hadley:
Unfortunately not, because I need to get a patch release out today — we can think through the full issue when I have more time.

hadley:
As I've thought more about `spread()` and `gather()`, I think a more radical rewrite is needed. So I'm going to leave `spread()` and `gather()` as is, and introduce new verbs (probably `pivot_long()` and `pivot_wide()`. These should preserve row order automatically (because they're built on the more robust tools provided by vctrs).

PR #495, opened 2018-09-18
Update lazyeval compat file

lionel-:


lionel-:
* Update lazyeval compat file which contained lots of old functions

* Unquote quosure with `!!` rather than `!!!`

* Use `ensym()` rather than `enexpr()` to capture symbols. This is slightly incompatible because the former approach allowed creating non-syntactic column names like `foo(bar)`, though I doubt many users do that.

This is compatible with both CRAN and dev rlang.

PR #513, opened 2018-10-24
Use variant of ensym() that supports quosures

lionel-:
Part of #512
See r-lib/rlang#675

I checked that it fixes the 3 package failures you mentioned on Slack.


hadley:
Thanks!

PR #528, opened 2019-01-19
Specify what replace_na() returns. Closes #500.

### LINKED ISSUE #500: replace_na() documentation lacks @return
The documentation for `replace_na()` does not mention the returned data type. Newcomers to R or the tidyverse are forced to guess the type, and may (incorrectly) guess the wrong one

The code in R/replace_na.R is:

```
#' Replace missing values
#'
#' @param data A data frame or vector.
#' @param replace If `data` is a data frame, a named list giving the value to
#'   replace `NA` with for each column. If `data` is a vector, a single value
#'   used for replacement.
#' @param ... Additional arguments for methods. Currently unused.
#' @export
#' @examples
<snip>
```

I suggest adding

```
#' @return If `data` is a data frame, returns a data frame. If `data` is a vector, returns a vector of class determined by the union of `data` and `replace`.

juliasilge:
It was an enlightening experience submitting this one line documentation change. I had to install the development versions of dplyr and tibble (lots of compiled code) to document and check (didn't work the first time; I had to restart R a couple of times), got [this error](https://github.com/tidyverse/readxl/issues/520) when running `devtools::document()`, and more. Not necessarily bad/wrong, but enlightening to realize how high the barrier is for even very simple changes.

### DIFF
--- a/R/replace_na.R
+++ b/R/replace_na.R
@@ -5,6 +5,9 @@
 #'   replace `NA` with for each column. If `data` is a vector, a single value
 #'   used for replacement.
 #' @param ... Additional arguments for methods. Currently unused.
+#' @return If `data` is a data frame, returns a data frame. If `data` is a
+#'   vector, returns a vector of class determined by the union of `data` and
+#'   `replace`.
 #' @seealso \code{\link[dplyr]{na_if}} to replace specified values with a `NA`.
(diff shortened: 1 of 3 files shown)


hadley:
Yeah, I often "cheat" by just documenting and not checking, but that requires an equally sophisticated model (and I often still get it wrong)

Issue #485, opened 2018-08-07
fill() `...` description is misleading


DavisVaughan:
According to the docs, `fill()` supposedly fills on all columns if none are selected. But this is not the case:

``` r
library(tidyr)

df <- data.frame(Month = 1:5, Year = c(2000, rep(NA, 4)))
fill(df)
#>   Month Year
#> 1     1 2000
#> 2     2   NA
#> 3     3   NA
#> 4     4   NA
#> 5     5   NA

fill(df, Year)
#>   Month Year
#> 1     1 2000
#> 2     2 2000
#> 3     3 2000
#> 4     4 2000
#> 5     5 2000
```

Created on 2018-08-07 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

It is written this way because it inherits that parameter from `gather()`, which does select all if none are specified.

I like the idea of filling all columns if none are specified, so maybe you could use the following.

``` r
fill2 <- function (data, ..., .direction = c("down", "up")) 
{
  quos <- rlang::quos(...)
  if(rlang::is_empty(quos)) {
    fill_cols <- names(data)
  } else {
    fill_cols <- unname(tidyselect::vars_select(names(data), !!! quos))
  }
  
  .direction <- match.arg(.direction)
  fillVector <- switch(.direction, down = tidyr:::fillDown, up = tidyr:::fillUp)
  for (col in fill_cols) {
    data[[col]] <- fillVector(data[[col]])
  }
  data
}

df <- data.frame(Month = 1:5, Year = c(2000, rep(NA, 4)))
fill2(df)
#>   Month Year
#> 1     1 2000
#> 2     2 2000
#> 3     3 2000
#> 4     4 2000
#> 5     5 2000
```

Here's the clean version without `::`'s.

<details>

```r
fill.data.frame <- function (data, ..., .direction = c("down", "up")) 
{
  quos <- quos(...)
  if(is_empty(quos)) {
    fill_cols <- names(data)
  } else {
    fill_cols <- unname(tidyselect::vars_select(names(data), !!! quos))
  }
  
  .direction <- match.arg(.direction)
  fillVector <- switch(.direction, down = fillDown, up = fillUp)
  for (col in fill_cols) {
    data[[col]] <- fillVector(data[[col]])
  }
  data
}
```

</details>

leungi:
I was just about to report this; I believe this behaviour is only seen after recent upgrade.

hadley:
[...]

PR #529, opened 2019-01-19
fix issue #485

eokodie:


hadley:
I think you'll also need to do something like `@params ...`.  Currently the documentation is being inherited from `gather()` which is yielding the incorrect advice in the generated docs.

R/fill.R
```diff
@@ -3,14 +3,15 @@ NULL
...
-#' Fills missing values in using the previous entry. This is useful in the
+#' Fills missing values in selected columns using the previous entry. This is useful in the
 #' common output format where values are not repeated, they're recorded
 #' each time they change.
 #'
 #' Missing values are replaced in atomic vectors; `NULL`s are replaced
 #' in list.
 #'
 #' @inheritParams gather
+#' @param ... A selection of columns. If empty, nothing happens. You can supply bare variable names, select all variables between x and z with x:z, exclude y with -y. For more selection options, see the [dplyr::select()] documentation.
```
hadley: This is the right fix, but can you please wrap these lines at ~80 cols? Thanks!

eokodie: Fixes #485

eokodie: Thanks

hadley:
Thank you!

PR #530, opened 2019-01-19
The nest function now returns a tibble list rather than data frames

### LINKED ISSUE #475: Return tibble when nesting
Currently, `mtcars %>% group_by(am, cyl) %>% nest()` returns a tibble whereas `mtcars %>% nest(-am, -cyl)` returns a data.frame with list-cols, which doesn't print well. 

Since `tibble` is already a dependency of tidyr, I imagine that it could be as simple as adding `as_tibble` to the beginning of [`nest.data.frame`](https://github.com/tidyverse/tidyr/blob/581c2c0fa0557e6c4ec2add986ce6f1d3746c580/R/nest.R#L37) but perhaps there's a more elegant method?

Coercing the _output_ to a tibble results in a list-col of data.frames, whereas coercing the _input_ returns a list-col of tibbles:

``` r
library(dplyr)
library(tidyr)

mtcars %>% nest(-am, -cyl) %>% as_tibble()
#> # A tibble: 6 x 3
#>     cyl    am data                 
#>   <dbl> <dbl> <list>               
#> 1    6.    1. <data.frame [3 x 9]> 
#> 2    4.    1. <data.frame [8 x 9]> 
#> 3    6.    0. <data.frame [4 x 9]> 
#> 4    8.    0. <data.frame [12 x 9]>
#> 5    4.    0. <data.frame [3 x 9]> 
#> 6    8.    1. <data.frame [2 x 9]>

mtcars %>% as_tibble() %>% nest(-am, -cyl)
#> # A tibble: 6 x 3
#>     cyl    am data             
#>   <dbl> <dbl> <list>           
#> 1    6.    1. <tibble [3 x 9]> 
#> 2    4.    1. <tibble [8 x 9]> 
#> 3    6.    0. <tibble [4 x 9]> 
#> 4    8.    0. <tibble [12 x 9]>
#> 5    4.    0. <tibble [3 x 9]> 
#> 6    8.    1. <tibble [2 x 9]>
```

stufield:
Fixes #475


R/nest.R
```diff
@@ -33,10 +33,15 @@
 nest <- function(data, ..., .key = "data") {
   UseMethod("nest")
 }
+
 #' @export
 nest.data.frame <- function(data, ..., .key = "data") {
   key_var <- as_string(ensym2(.key))
 
+  if (!inherits(data, "tbl_df")) {
```
hadley: I was thinking that we'd do this unconditionally, simply because nested data frames are so painful to work with if you don't have the nice tlbble print methods.

stufield: Ok, done! Update committed and pushed. Thanks!

tests/testthat/test-nest.R
```diff
@@ -49,3 +49,13 @@ test_that("nesting works for empty data frames", {
   out <- nest(df, x, y)
   expect_equal(out$data, list(df))
 })
+
+test_that("tibble conversion occurs in the `nest.data.frame()` method", {
+  tbl <- mtcars %>% nest(-am, -cyl)
+  expect_s3_class(tbl, "tbl_df")
+  expect_true(
```
hadley: I think this is probably slightly too much — I'd recommend just checking the value of the first element of the list (i.e. `tbl$data[[1]]`)

stufield: Sure no problem ... though it seems this change has induced some downstream unit test failures elsewhere. I'll look into it some more and see if I can come up with something better.

stufield: It looks like unconditionally converting to tibble class causes the nesting variable, e.g. `x`, to remain in the returned object once `nest()` is called. It should be dropped. This is causing all the downstream issues. I'll do some digging, but I was hoping this one was going to be relatively simple :(

stufield: Here is the issue ... the `as_tibble()` ungroups the tibble if it's already a tibble.
```r
df <- tibble(x = c(1, 1, 1), y = 1:3) %>% dplyr::group_by(x)
df %>% dplyr::is_grouped_df()     # TRUE   
df %>% as_tibble() %>% dplyr::is_grouped_df().  # FALSE
```
This might be difficult to work around without digging into `as_tibble()` (aside, the new tibble pkg `v2.0.1` has been causing me headaches), but coercing the output seems like a bit of a hack? Maybe?
Stripping the `grouped_df` class causes a bunch of downstream things *not* to happen, like getting the `group_vars` etc. This seems a bit like a `tibble::as_tibble()` issue :(

Perhaps the condition is the simplest way forward? At least until `as_tibble()` doesn't strip the `grouped_df` class. What do you suggest?

hadley: Maybe we need another `tbl.df` method that doesn't do the coercion? So `nest.data.frame()` would coerce to a tibble, and then call `nest.tbl_df()`?  (Let me know if that's not enough info to know what to do next)

hadley: I'd put it above it, but yeah.

stufield: Hold on ... I just re-read what you said and realize I misunderstood. There's a lot of duplicated code in my solution, I'll just call `nest.tbl_df()` rather than repeat the action part of the function. Can you squash the commits on your end?

R/nest.R
```diff
@@ -35,12 +35,45 @@ nest <- function(data, ..., .key = "data") {
...
+  out
+}
+
+#' @export
+nest.data.frame <- function(data, ..., .key = "data") {
+
+  key_var   <- as_string(ensym2(.key))
+  data      <- tibble::as_tibble(data)
   nest_vars <- unname(tidyselect::vars_select(names(data), ...))
+
```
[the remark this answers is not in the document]
stufield: Yes, that is what I did when I realized that I had misunderstood your comment. That *should* have happened in commit 3531371, so I'm not sure why you're not seeing it. Perhaps I didn't push? I can double-check. The `data.frame` method should now look like this:
```r
nest.data.frame <- function(data, ..., .key = "data") {
  .key <- enquo(.key)
  data <- tibble::as_tibble(data)
  nest.tbl_df(data, ..., .key = !! .key)
}
```
By the way, I also just realized there's a remnant `new <-` line 18 in the unit testing code from debugging I would like to remove. I really hope you can squash all this into one or two commits, cause this looks pretty ugly. I will re-push to my fork in a few minutes.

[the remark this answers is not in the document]
hadley: Hmmm, maybe I was looking at an old version? It all looks good now!

(And yes, I'll do the squashing when I merge)

hadley:
Thank you!

Issue #526, opened 2019-01-08
Add quasiquotation example to `separate()` examples


batpigandme:
The [`separate()`](https://connect.rstudioservices.com/content/204/top_ref_pages.html) function reference is the most viewed page on tidyr.tidyverse.org. 

* Since `col` supports quasiquotation, it might be worth adding one example of this.
* Review existing examples to make sure they're clear and cover most-common scenarios https://github.com/tidyverse/tidyr/blob/master/R/separate.R#L39-L58

PR #532, opened 2019-01-19
Added and reviewed separate() examples

jminnier:
Added a few examples using:

- convert=TRUE
- regex for column separations
- quasiquotation

#tidy-dev-day addresses issue #526

### DIFF
--- a/R/separate.R
+++ b/R/separate.R
@@ -63,6 +63,11 @@
 #' df <- data.frame(x = c(NA, "a?b", "a.d", "b:c"))
 #' df %>% separate(x, c("A","B"), sep = "([\\.\\?\\:])")
 #'
+#' # convert = TRUE detects column classes
+#' df <- data.frame(x = c("a:1", "a:2", "c:4", "d", NA))
+#' df %>% separate(x, c("key","value"), ":") %>% str
+#' df %>% separate(x, c("key","value"), ":", convert = TRUE) %>% str
+#'
 #' # Argument col can take quasiquotation
 #' quasi_commasep <- function(data, col_name) {
 #'    # quote the column input variable
--- a/R/separate.R
+++ b/R/separate.R
@@ -69,9 +69,8 @@
 #' df %>% separate(x, c("key","value"), ":", convert = TRUE) %>% str
 #'
 #' # Argument col can take quasiquotation to work with strings
-#' df <- data.frame(pets = c("dog, cat, bunny","rock, cat, ferret","chinchilla, cat, pig"))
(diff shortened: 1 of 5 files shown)


hadley:
It looks like you have a few other Rd differences. Can you please check you have the latest roxygen2?

R/separate.R
```diff
@@ -48,14 +48,42 @@
...
 #' # Another option:
 #' df %>% separate(x, c("a", "b"), extra = "merge", fill = "left")
+#' # Or you can keep all three
+#' df %>% separate(x, c("a", "b", "c"))
 #'
 #' # If only want to split specified number of times use extra = "merge"
 #' df <- data.frame(x = c("x: 123", "y: error: 7"))
 #' df %>% separate(x, c("key", "value"), ": ", extra = "merge")
+#'
+#' # Use regular expressions to separate on multiple characters:
```
hadley: This is great!

R/separate.R
```diff
@@ -48,14 +48,42 @@
...
+#' df <- data.frame(x = c(NA, "a?b", "a.d", "b:c"))
+#' df %>% separate(x, c("A","B"), sep = "([\\.\\?\\:])")
+#'
+#' # convert = TRUE detects column classes
+#' df <- data.frame(x = c("a:1", "a:2", "c:4", "d", NA))
+#' df %>% separate(x, c("key","value"), ":") %>% str
+#' df %>% separate(x, c("key","value"), ":", convert = TRUE) %>% str
+#'
+#' # Argument col can take quasiquotation
+#' quasi_commasep <- function(data, col_name) {
```
hadley: I wonder if there's a way to make this simpler somehow. I'm worried it's a bit much to take in in an example.

jminnier: I was thinking the same thing... maybe a function is too much and I should just use the !! directly. I'll try again!

R/separate.R
```diff
@@ -48,14 +48,29 @@
...
+#' df %>% separate(x, c("A","B"), sep = "([\\.\\?\\:])")
+#'
+#' # convert = TRUE detects column classes
+#' df <- data.frame(x = c("a:1", "a:2", "c:4", "d", NA))
+#' df %>% separate(x, c("key","value"), ":") %>% str
+#' df %>% separate(x, c("key","value"), ":", convert = TRUE) %>% str
+#'
+#' # Argument col can take quasiquotation to work with strings
+#' var <- "x"
+#' df %>% separate(!!var, c("key","value"),":")
```
hadley:
```suggestion
#' df %>% separate(!!var, c("key","value"), ":")
```

hadley:
Thank you!!

PR #533, opened 2019-01-19
change complete() to use full_join(). fixes #493

### LINKED ISSUE #493: Using complete with a subset of levels silently drops observations
Based upon the description of complete  its purpose is to add missing observations for specific combinations of variable values. I believe therefore it is not very intuitive for it to silently drop observations if certain levels are not present,  for example:

```
library(tidyr)
library(dplyr)

iris %>% 
    as_data_frame() %>% 
    mutate( Species = as.character(Species)) %>% 
    complete( Species = c("versicolor", "virginica"))
```

This code silently drops all observations with Species = setosa

I would propose either changing the function to use a full_join()  to ensure the original data is preserved or at the very least adding a warning/message so the user is aware that data has been lost. 

I am happy to make a PR if people agree this is a beneficial change.

Ryo-N7:
use full_join() instead of left_join() so that original data is preserved even if certain levels are not specified in complete()


hadley:
Does this fix an existing issue?

hadley:
Can you please add a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

hadley:
Thank you!

PR #535, opened 2019-01-19
Closes #483 (Unnesting tibble with zero rows removes unnested column)

### LINKED ISSUE #483: Unnesting tibble with zero rows removes unnested column
Here is an example that produces the problem:

```
df <- tibble(
 a = list(c(1, 2), 3),
 b = c(4,5)
)

df %>%
  unnest(a) %>%
  colnames
# [1] "b" "a"

df %>%
  filter(b==1) %>%
  unnest(a) %>%
  colnames
# [1] "b"
```

If the special case occurs that the tibble being unnested has no rows, all code acting on the unnested column breaks.

khailper:


hadley:
I think this should work, rather than throwing an error.

khailper:
Are you using the original reprex? I think your running into the same dev dplyr issue as before. I'm getting the same ` Error in data.matrix(data): (list) object cannot be coerced to type 'double'` you got, and it's happening at `filter`. on the other hand, the reprex you generated of:

```
data.frame(a = character(), b = character()) %>% 
  unnest(b)
```
is working fine.  I don't have a reprex::reprex()  example since devtools::install() is giving me a weird make-related error.

hadley:
Sorry, I see now that it's a warning — but I don't think even that should be necessary.  Would you mind also adding a simple unit test to make sure that the column names are preserved as expected?

khailper:
Will do.

hadley:
@khailper I did a bit of exploration and I think the underlying problem was a bit different — thanks for the unit test and the motivation to work on this problem!

khailper:
Glad I could contribute and thank you for the opportunity to learn development skills.

Issue #491, opened 2018-08-29
Not specifying columns in separate-rows does not select all columns


gfiumara:
The documentation for `...` in `separate_rows` says:
>A selection of columns. If empty, all variables are selected.

This does not appear to be the case:
```
> squares <- data.frame(original = "1,2,3", squared = "1,4,9")
> tidyr::separate_rows(squares)
> squares
  original squared
1    1,2,3   1,4,9

> tidyr::separate_rows(squares)
  original squared
1    1,2,3   1,4,9
```

If all the columns are specified, `separate_rows` works correctly:

```
> tidyr::separate_rows(squares, tidyselect::everything())
  original squared
1        1       1
2        2       4
3        3       9
```

If the documentation is correct, I believe the issue has to do with: https://github.com/tidyverse/tidyr/blob/581c2c0fa0557e6c4ec2add986ce6f1d3746c580/R/separate-rows.R#L28
```
> unname(tidyselect::vars_select(names(squares))) 
character(0)
```

tidyr_0.8.1

hadley:
I think the documentation is wrong; we should be consistent with what tidyselect does here.

stephenrknight:
I've got it.... :D

PR #536, opened 2019-01-19
#491 Documentation corrected for ... selection of columns

stephenrknight:
#491 Documentation corrected for ... selection of columns


hadley:
Can you please coordinate with #529? I think it might be better to merge that first, and then you can just add a new `@inheritParams`

hadley:
Superseded by #538

PR #538, opened 2019-01-19
Fix #491: after merge and added in @inheritParams fill

### LINKED ISSUE #491: Not specifying columns in separate-rows does not select all columns

stephenrknight:
Fix #491: after merge and added in @inheritParams fill

### DIFF
--- a/R/separate-rows.R
+++ b/R/separate-rows.R
@@ -4,8 +4,8 @@
 #' separates the values and places each one in its own row.
 #'
 #' @inheritSection gather Rules for selection
-#' @inheritParams gather
 #' @inheritParams fill
+#' @inheritParams gather
 #' @param sep Separator delimiting collapsed values.
 #' @export
 #' @examples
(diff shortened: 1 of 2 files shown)


hadley:
I think you need to flip the order? It doesn't look like this changed the docs.

Issue #460, opened 2018-06-05
Unnesting list-columns gave error even they all have equal lengths


yuanwxu:
Here is an example that reproduces the error.

```R
> df <- tibble(x = c("a", "b", "c"), y = list(a = 1:3, b = 4:6, c = 7:9), z = list(1:3))
> df
# A tibble: 3 x 3
  x     y         z        
  <chr> <list>    <list>   
1 a     <int [3]> <int [3]>
2 b     <int [3]> <int [3]>
3 c     <int [3]> <int [3]>
> unnest(df)
Error: All nested columns must have the same number of elements.
```
However, I note that slice the whole data frame and then unnest works

```R
> df %>% slice(1:n()) %>% unnest()
# A tibble: 9 x 3
  x         y     z
  <chr> <int> <int>
1 a         1     1
2 a         2     2
3 a         3     3
4 b         4     1
5 b         5     2
6 b         6     3
7 c         7     1
8 c         8     2
9 c         9     3
```

markdly:
Hi @yuanwxu, just another community member here.
Here's a slightly more minimal [reprex](http://reprex.tidyverse.org) of what I think the issue is:

**Issue**
``` r
library(tidyverse)
df1 <- tibble(x = c("a"), y = list(1:3), z = list(4:6))
df1 %>% unnest()
#> # A tibble: 3 x 3
#>   x         y     z
#>   <chr> <int> <int>
#> 1 a         1     4
#> 2 a         2     5
#> 3 a         3     6

# add name to list in y causes issue
df2 <- tibble(x = c("a"), y = list(foo = 1:3), z = list(4:6))
df2 %>% unnest()
#> Error: All nested columns must have the same number of elements.
```

**Workaround**

Another possible workaround instead of slicing could be to unname the list columns first
``` r
df2 %>% mutate_if(is.list, unname) %>% unnest()
#> # A tibble: 3 x 3
#>   x         y     z
#>   <chr> <int> <int>
#> 1 a         1     4
#> 2 a         2     5
#> 3 a         3     6
```

**Why is this happening?**
Looking at the code, I think this behaviour comes from 

https://github.com/tidyverse/tidyr/blob/cbdd14e90b2a771e242a44d1ed5eea84d53da642/R/unnest.R\#L96-L100

If we adapt that for our example we can see why the error message occurs:

``` r
# lengths considered the same
n1 <- df1 %>% select(-x) %>% map(~ map_int(., NROW))
n1
#> $y
[...]

PR #539, opened 2019-01-20
Fixed issue 460 with named lists

hlendway:
Removed list names for NROW check so that it wouldn't include list name in row count. Added a test case so that lists of the same length with and without names would pass tests. Fixed test case wording that I don't think made sense, "can unnested lists" I think was meant to be "can unnest nested lists" . 
Feel free to request changes to my PR, thanks. #460


hlendway:
I will update NEWS.md today and add to PR.

R/unnest.R
```diff
@@ -94,8 +94,7 @@ unnest.data.frame <- function(data, ..., .drop = NA, .id = NULL,
   }
 
   nested <- dplyr::transmute(dplyr::ungroup(data), !!! quos)
-  nested_unnamed <- nested %>%
-    dplyr::mutate_if(is.list, unname)
+  nested_unnamed <- map(nested, unname_list)
```
hadley: Oh, actually I think you can use `purrr::map_if()` — sorry for not thinking of this earlier!

hadley:
Thanks @hlendway — I had to run this locally so I understood exactly what the problem was, and then I thought of an even simpler solution. Thanks for your help with this, it's much appreciated!

PR #540, opened 2019-01-27
Correctly handle missing values when spreading nested columns

### LINKED ISSUE #522: spread define fill for list-columns
`fill` argument in `spread` seems not to be working with list columns. Consider the following example:

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(tidyr)
df <- tribble(
    ~ind,   ~arm, ~var_1, ~var_2,
    1,    "A",      5,      3,
    1,    "B",      1,      1,
    2,    "A",      2,      4
)

dfn <- nest(df, var_1, var_2)
dfn
#> # A tibble: 3 x 3
#>     ind arm   data            
#>   <dbl> <chr> <list>          
#> 1     1 A     <tibble [1 x 2]>
#> 2     1 B     <tibble [1 x 2]>
#> 3     2 A     <tibble [1 x 2]>

dfn %>%
    group_by(ind) %>% 
    spread(arm, data, fill = data_frame(var_1 = 0, var_2 = 0))
#> Warning in if (!is.na(fill)) {: the condition has length > 1 and only the
#> first element will be used
#> # A tibble: 2 x 3
#> # Groups:   ind [2]
#>     ind A                B               
#>   <dbl> <list>           <list>          
#> 1     1 <tibble [1 x 2]> <tibble [1 x 2]>
#> 2     2 <tibble [1 x 2]> <NULL>
```
I would expect this \<NULL\> to be filled with the tibble I passed to the `fill` argument.

mtoto:
This PR attempts to fix #522 by adding an extra condition for list columns when using the `fill=` argument inside `spread()`. Because the fill value is likely to be more than just a scalar, I wrapped `!is.na(fill)` inside `any()`.

### DIFF
--- a/R/spread.R
+++ b/R/spread.R
@@ -98,7 +98,7 @@ spread.data.frame <- function(data, key, value, fill = NA, convert = FALSE,
 
   value <- data[[value_var]]
   ordered <- value[overall]
-  if (all(!is.na(fill))) {
+  if (any(!is.na(fill))) {
     if (class(ordered) == "list") {
       ordered[ordered == "NULL"] <- list(fill)
     }
--- a/R/spread.R
+++ b/R/spread.R
@@ -98,7 +98,10 @@ spread.data.frame <- function(data, key, value, fill = NA, convert = FALSE,
 
   value <- data[[value_var]]
   ordered <- value[overall]
-  if (!is.na(fill)) {
+  if (all(!is.na(fill))) {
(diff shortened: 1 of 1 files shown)


hadley:
This isn't quite the right approach, unfortunately, and I think considering non-atomic values of `fill` will require a little more thought. But I appreciate your willingness to take a swing at it 😄

Issue #150, opened 2016-01-02
Implement multi-gather


hadley:
``` r
library(dplyr, warn.conflicts = FALSE)
library(tidyr)

# From Jenny Bryan --------------------------------------------------------

input <- tribble(
  ~hw,   ~name,  ~mark,   ~pr,
  "hw1", "anna",    95,  "ok",
  "hw1", "alan",    90, "meh",
  "hw1", "carl",    85,  "ok",
  "hw2", "alan",    70, "meh",
  "hw2", "carl",    80,  "ok"
)

# Want:
input %>%
  gather(key = element, value = score, mark, pr) %>%
  unite(thing, hw, element, remove = TRUE) %>%
  spread(thing, score, convert = TRUE)
#> # A tibble: 3 x 5
#>   name  hw1_mark hw1_pr hw2_mark hw2_pr
#>   <chr>    <int> <chr>     <int> <chr> 
#> 1 alan        90 meh          70 meh   
#> 2 anna        95 ok           NA <NA>  
#> 3 carl        85 ok           80 ok

# http://stackoverflow.com/questions/33599665 -----------------------------

anscombe %>%
  gather() %>%
  separate(key, c("var", "ex"), 1) %>%
  group_by(var) %>%
  mutate(id = row_number()) %>%
  spread(var, value)
#> # A tibble: 44 x 4
#>    ex       id     x     y
#>    <chr> <int> <dbl> <dbl>
#>  1 1         1    10  8.04
#>  2 1         2     8  6.95
#>  3 1         3    13  7.58
#>  4 1         4     9  8.81
#>  5 1         5    11  8.33
#>  6 1         6    14  9.96
#>  7 1         7     6  7.24
#>  8 1         8     4  4.26
#>  9 1         9    12 10.8 
#> 10 1        10     7  4.82
#> # … with 34 more rows

# http://stackoverflow.com/questions/27247078 -----------------------------

df <- tribble(
  ~id, ~type,     ~transactions, ~amount,
  20,  "income",  20,            100,
  20,  "expense", 25,            95,
  30,  "income",  50,            300,
  30,  "expense", 45,            250
)

df %>%
  gather(var, val, transactions:amount) %>%
  unite(var2, type, var) %>%
  spread(var2, val)
#> # A tibble: 2 x 5
#>      id expense_amount expense_transactio… income_amount income_transactio…
#>   <dbl>          <dbl>               <dbl>         <dbl>              <dbl>
[...]

PR #544, opened 2019-02-05
WIP: Enhance gather() to support data.frame columns

yutannihilation:
Related: #150 

(This is mainly for PoC. I'm not sure I can finish this PR with my poor C++ skill...)

`gather()` seems to support data.frame column if we modify C++ functions `rep_()` and `concatinate()` to support data.frame. It might be just doing recursion, but I might oversee something.

current progress:

``` r
library(tibble)
library(tidyr)

d <- tibble(id = 1:2,
            x = tibble(a = c("a", "b"), b = 1:2, c = Sys.Date() + b),
            y = tibble(a = c("c", "d"), b = 3:4, c = Sys.Date() + b))

gather(d, k, v, x:y)
#> # A tibble: 4 x 3
#>      id k     v$a      $b $c        
#>   <int> <chr> <chr> <int> <date>    
#> 1     1 x     a         1 2019-02-06
#> 2     2 x     b         2 2019-02-07
#> 3     1 y     c         3 2019-02-08
#> 4     2 y     d         4 2019-02-09
```

<sup>Created on 2019-02-05 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -13,6 +13,10 @@ rep_ <- function(x, n, var_name) {
     .Call(`_tidyr_rep_`, x, n, var_name)
 }
 
+concatenate <- function(x, ind, factorsAsStrings) {
+    .Call(`_tidyr_concatenate`, x, ind, factorsAsStrings)
+}
+
 melt_dataframe <- function(data, id_ind, measure_ind, variable_name, value_name, attrTemplate, factorsAsStrings, valueAsFactor, variableAsFactor) {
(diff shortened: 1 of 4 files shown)


yutannihilation:
@hadley @krlmlr 
Do you have any thoughts or plans about using vctrs in tidyr? I don't know how to use vctrs in C++ codes, but I feel I really need it here...

hadley:
Yeah, the plan is to overhaul all this code to use vctrs instead. That’s likely to require changes to vctrs so that the code is both performant and accessible from C.

yutannihilation:
Wow, cool! It sounds like vctrs is the next Rcpp...

Then I close this PR. Making this PR was a nice lesson that teached me we really need vctrs. Looking forward to the day!

PR #549, opened 2019-02-07
library reference corrected in section Examples

larspijnappel:
changed library reference from `dplyr` to `tidyr` in Examples section


hadley:
You don't need to load the package to which the examples apply.

PR #552, opened 2019-02-11
Closes #515 and adds link

### LINKED ISSUE #515: Animated gif in documentation
## The gif

@apreshill made the best explanation for `tidyr::gather()` that I have ever seen:

gather

Thank you, Alison! :100: 

The fundamental problem this gif solved for me is the confusion caused by the `key` and `value` arguments.

Before seeing this gif, I falsely (and instinctively) equated these things:
- the `key` and `value` arguments in `tidyr::gather()`
- the `id.vars` and `measure.vars` arguments in `reshape2::melt()`

This gif finally opened my eyes to the fact that `key` and `value` in `tidyr::gather()` are analogous to `variable.name` and `value.name` in `reshape2::melt()`. So many struggles :sob: 

Could we include this gif directly in the documentation on [this page][2]?

[2]: https://tidyr.tidyverse.org/reference/gather.html

untitled drawing

Right now the gif is [posted on Twitter][1], but I'd like to see it in more places where people can find it while they're searching for information about `tidyr::gather()`.

[1]: https://twitter.com/WeAreRLadies/status/1059520693857996800

## A complaint about magical expressions

Also, I want to complain that the example in the documentation uses the naked words `stock` and `price` instead of `"stock"` and `"price"` as I would have expected. This is one of the magical things about R that was always difficult to accept as a beginner (and continues to be a challenge as a more experienced R user). It is difficult to accept that R will just *know* that `stock` should be turned into `"stock"` as a column name, instead of throwing an error because there is no `stock` variable defined anywhere, and there is no `stock` column in the input dataframe, either.

My expectation is this:

```
data %>% gather(stock, ...)
Error: object 'stock' not found
```

## A direct comparison to melt

Lastly, you might consider whether the documentation for `gather()` would be enhanced by showing a direct side-by-side comparison of the same operation executed  2 times, once by `gather()` and again by `melt()`. Personally, I would find it helpful. Others might find this confusing if they have never seen either function before, so it might not be a good idea.

Something like this:

```r
library(dplyr)
library(tidyr)

mini_iris <- iris %>%
  group_by(Species) %>%
  slice(1)

# Let's gather all the measurements (except Species) into 1 column:
mini_iris %>% gather(key = flower_att, value = measurement, -Species)

# Now let's do the same thing again with reshape2::melt()
library(reshape2)

melt(
  data = mini_iris,
  variable.name = "flower_att",
  value.name = "measurement",
  id.vars = c("Species")
)
```

khailper:
Gif is now part of tidyr::gather documentation, with link to gif if help page doesn't render properly.


hadley:
Thanks for all your work on this @khailper — it's made me realise just how useful these animations will be for people. But since I'm working on updates to gather/spread, I think I'm going to hold off on merging these gifs in favour of working out animations for the new tools.

khailper:
@hadley  Thanks for the opportunity and for all the work you and others at RStudio did to make Tidy Dev Day possible. It was an incredibly helpful way to learn about collaborative development and building packages for CRAN.

PR #575, opened 2019-03-19
Typos in pivot vignette

lazappi:
I was just reading through the vignette for the new pivot functions and noticed there were a few typos it might be good to fix. The list of things I have changed is below (I find the diff sometimes isn't useful for text where there can be multiple changes on the same line). Some of this might just be my opinion so feel free to ignore anything you don't like.

I also thought the sentence on line 24 wasn't quite clear but I wasn't sure how to fix it, so I left it.

P.S. The new functions look excellent! 👍

Changes:

```
L21: "It's" -> "It"
L55: Add "about"
L55: "there" -> "their
L55: "annaul" -> "annual"
L82: "manully" -> "manually"
L152: Add "and"
L201: "inside" -> "instead"
L234: "even" -> "ever"
L234: "access a" -> "access to a"
L274: "with columns with" -> "with columns"
L294: Removed extra space
L294: "describe" -> "describes"
L306: "record is means" -> "record means"
L318: "collected" -> "containing"
L337: "as the" -> "as for"
L339: "data set" -> "dataset"
L352: "income" -> "dataset"
L358: Add "and"
L377: "get into" -> "get them into"
L409: "complet" -> "complete"
L442: Remove "comes"
L456: "time see" -> "time we see"
```

### DIFF
--- a/vignettes/pivot.Rmd
+++ b/vignettes/pivot.Rmd
@@ -19,7 +19,7 @@ options(tibble.print_max = 10)
 
 This vignette describes the use of the new `pivot_long()` and `pivot_wide()` functions. Their goal is to improve the usability of `gather()` and `spread()`, and incorporate state-of-art features found in other packages.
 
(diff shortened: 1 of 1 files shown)


hadley:
Thanks! No need to list the changes in the PR description as I'll see them in the diff.

PR #588, opened 2019-03-20
R 3.1 support

### LINKED ISSUE #565: Resolve R 3.1 failure
See https://travis-ci.org/tidyverse/tidyr/jobs/500822234

Root cause is likely to be https://github.com/r-lib/vctrs/issues/114

DavisVaughan:
Re-enables R 3.1 travis build.

In the process, I figured out that the changes to `fill()` broke one of the tests on 3.1. Specifically, the `...` don't get passed along correctly for the grouped data frame method on 3.1. I think `NextMethod()` evaluates the `...` too early on 3.1 so [this test](https://github.com/tidyverse/tidyr/blob/master/tests/testthat/test-fill.R#L81) was failing with an `object 'y' not found` error. Since the grouped method doesn't seem to be required, I removed it to fix the issue.

Closes #565


hadley:
Thanks!!

Issue #596, opened 2019-03-25
Linking between pivot_* docs and gather/spread docs


martinjhnhadley:
The new `pivot_*` functions solve a clear need. Both in providing advanced features to `tidyr` and hopefully in simplifying folks' workflows. However, IMO it's important that users do not feel overwhelmed or that the existing functions will be deprecated. 

In particular, I think the following should be communicated to users:

- We acknowledge folks found `spread`/`gather` difficult to memorise and use, as mentioned in the vignette.
- We assure users that `spread`/`gather` are not going to be deprecated. They do not need to use the new `pivot_*`
- We tell users that new things are possible with the `pivot_*` functions
- We tell new users who meet `pivot_*` first that `spread`/`gather` exist, to reduce confusion when reading old SO answers etc.
- We provide more examples real dataset examples of how to switch between `spread`/`gather` and the `pivot_*` functions.

Here are places where I think that we need to add notes or examples, there might be more I've not considered.

- `spread`/`gather` docs pages
- `pivot_*` docs pages

hadley:
Would you be interested in contributing something to the pivot vignette too? I was thinking maybe a spread/gather vs pivot_long/pivot_wide section at the end. (I don't think it belongs at the start, because enough people found spread/gather confusing that it's not the right place to start the description)

martinjhnhadley:
I’ll happily make some suggestions for the vignette. I agree it’s a necessary thing, and definitely doesn’t belong at the beginning of the document.

PR #597, opened 2019-03-25
Linking between pivot_* docs and spread/gather

martinjhnhadley:
This is a PR to address Issue https://github.com/tidyverse/tidyr/issues/596.
Information about `pivot_*` siblings has been added to the `spread`/`gather` docs, and vice versa.


R/gather.R
```diff
@@ -53,6 +53,22 @@
 #' @param factor_key If `FALSE`, the default, the key values will be
 #'   stored as a character vector. If `TRUE`, will be stored as a factor,
 #'   which preserves the original ordering of the columns.
+#'
+#' @details
```
hadley: I like the intent here, but I think:

* It should be it's own section, e.g. `@section Compared to `pivot_long()`
* It can be much shorter, and you don't need to reference the tidy data paper (which uses reshape2 anyway)
* Needs to link to `pivot_long()` with `[pivot_long()]`
* It should be a bit more assertive about switching to `pivot_long()` - `gather()` isn't going away but `pivot_long()` is more featureful and should be easier for most people to learn.

R/pivot-long.R
```diff
@@ -26,6 +26,25 @@
 #'   automatically guess from the data. Supply this argument when you want to
 #'   override that default. Should be a named list, where the names are
 #'   given by the value columns.
+#'
+#' @details
+#'
+#' `pivot_long` is a generalisation of the function `gather`, it differs from
```
hadley: This feels a bit too long for `pivot_long()` - I think you can indicate that it supersedes `gather()`, and maybe include a brief note on translation `gather()` code to `pivot_long()`? (i.e. `key` is now `names_to` and `value` is `values_to`)

hadley:
Are you interested in finishing this off?

R/gather.R
martinjhnhadley: I agree with all these comments, and have pushed these changes

R/pivot-long.R
martinjhnhadley: Should we include translation notes in the @section and also in the `Examples`, as I believe many users simply scroll to the bottom of a documentation page?

hadley:
Thanks!

PR #602, opened 2019-03-29
Clarify pivot docs for pivot_longer when multiple value columns are named

apreshill:
In testing out the new `pivot_longer` function in the multiple value columns case, I didn't see immediately (until I referenced the [vignette](https://tidyr.tidyverse.org/dev/articles/pivot.html#multiple-value-columns)) that `values_to` doesn't apply in this case. I tried to add a brief section in the reference doc here that would have helped me, drawing on the helpful portion of text in the existing vignette.

### DIFF
--- a/R/pivot-long.R
+++ b/R/pivot-long.R
@@ -12,6 +12,10 @@
 #'   the cell values (`values_to`) turn into in the result.
 #'
 #'   Note that these variables do not exist, so must be specified as strings.
+#'
+#'   If `names_to` contains multiple columns (i.e., `c(".value", "time")`),
+#'   you won't need to specify `values_to` because the special variable name
+#'   `.value` defines the name of the output value columns.
 #' @param names_prefix A regular expression used to remove matching text
 #'   from the start of each variable name.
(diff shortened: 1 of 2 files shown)


R/pivot-long.R
```diff
@@ -12,6 +12,10 @@
 #'   the cell values (`values_to`) turn into in the result.
 #'
 #'   Note that these variables do not exist, so must be specified as strings.
+#'
+#'   If `names_to` contains multiple columns (i.e., `c(".value", "time")`),
```
hadley: It's only when you use the special `".value"` name that that you can omit the `values_to`. But I'm not sure how to phrase that.

apreshill: Ah- so for the case when you have multiple names to but you don't use the `.value`? The current wording sounds like you always need to specify both args as a pair, so it sounds like you won't need to use `values_to` if `names_to` contains multiple columns and one of those is the special variable name `.value`?

hadley: Correct

PR #608, opened 2019-04-05
Correct typo in expand() example

Rekyt:
I noticed a small typo in the examples of `tidyr` it is the only occurrence of 'measurment' across the whole package. I couldn't update the .Rd file due to `vctrs` install issues on my machine.


hadley:
Thanks!

PR #610, opened 2019-04-14
.name and .value should always be first and second columns

### LINKED ISSUE #607: ".value" always need to come first in names_to?
The new pivot functions are amazing. One issue - In the multiple observations per row case, does the special variable `".value"` always have to come first as in the code and datasets in the vignette?

Here is an example with two datasets; only the variable naming conventions are different. My own dataset is in the second case, but I want output that looks like the output of `pivot_longer` in the first case with my existing dataset.  I thought I could simply swap the content of the `name_to` vector and get the same output, but it seems not (and I couldn't find this documented).

``` r

library(tibble)
library(tidyr)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 

# data
samp <- tibble(
  i = 1:4,
  y_t1 = rnorm(4),
  y_t2 = rnorm(4),
  z_t1 = rep(3, 4),
  z_t2 = rep(-2, 4),
)
samp
#> # A tibble: 4 x 5
#>       i   y_t1   y_t2  z_t1  z_t2
#>   <int>  <dbl>  <dbl> <dbl> <dbl>
#> 1     1 -0.702 -0.315     3    -2
#> 2     2  0.196  0.156     3    -2
#> 3     3  0.777 -1.64      3    -2
#> 4     4  1.06  -0.185     3    -2

# pivot - expected behavior
pivot_longer(samp, -i, names_to = c(".value", "time"), names_sep = "_")
#> # A tibble: 8 x 4
#>       i time       y     z
#>   <int> <chr>  <dbl> <dbl>
#> 1     1 t1    -0.702     3
#> 2     1 t2    -0.315    -2
#> 3     2 t1     0.196     3
#> 4     2 t2     0.156    -2
#> 5     3 t1     0.777     3
#> 6     3 t2    -1.64     -2
#> 7     4 t1     1.06      3
#> 8     4 t2    -0.185    -2


# sometimes variables are named like this
samp2 <- samp %>% 
  rename(t1_y = y_t1, t2_y = y_t2, t1_z = z_t1, t2_z = z_t2)

# why can't this recover the  same output as the first pivot?
pivot_longer(samp2, -i, names_to = c("time", ".value"), names_sep = "_")
#> # A tibble: 8 x 4
#>       i .value       y     z
#>   <int> <chr>    <dbl> <dbl>
#> 1     1 y       -0.315    NA
#> 2     1 z       NA        -2
#> 3     2 y        0.156    NA
#> 4     2 z       NA        -2
#> 5     3 y       -1.64     NA
#> 6     3 z       NA        -2
#> 7     4 y       -0.185    NA
#> 8     4 z       NA        -2
```

<sup>Created on 2019-04-04 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

romatik:
Fixes #607

### DIFF
--- a/R/pivot-long.R
+++ b/R/pivot-long.R
@@ -74,8 +74,8 @@ pivot_longer <- function(df,
   # Quick hack to ensure that split() preserves order
   v_fct <- factor(spec$.value, levels = unique(spec$.value))
   values <- split(spec$.name, v_fct)
-  value_keys <- split(spec[setdiff(names(spec), c(".name", ".value"))], v_fct)
-  keys <- vec_unique(spec[setdiff(names(spec), c(".name", ".value"))])
+  value_keys <- split(spec[-(1:2)], v_fct)
+  keys <- vec_unique(spec[-(1:2)])
 
   vals <- set_names(vec_na(list(), length(values)), names(values))
   for (value in names(values)) {
--- a/R/pivot-long.R
+++ b/R/pivot-long.R
@@ -65,10 +65,8 @@ pivot_longer <- function(df,
       names_pattern = names_pattern,
       col_ptypes = col_ptypes
     )
-    spec <- check_spec(spec)
-  } else {
-    spec <- check_spec(spec)
   }
+  spec <- check_spec(spec)
   spec <- deduplicate_names(spec, df)
 
   # Quick hack to ensure that split() preserves order
--- a/R/pivot-long.R
+++ b/R/pivot-long.R
@@ -73,8 +73,8 @@ pivot_longer <- function(df,
   # Quick hack to ensure that split() preserves order
   v_fct <- factor(spec$.value, levels = unique(spec$.value))
   values <- split(spec$.name, v_fct)
-  value_keys <- split(spec[-(1:2)], v_fct)
-  keys <- vec_unique(spec[-(1:2)])
+  value_keys <- split(spec[setdiff(names(spec), c(".name", ".value"))], v_fct)
+  keys <- vec_unique(spec[setdiff(names(spec), c(".name", ".value"))])
 
   vals <- set_names(vec_na(list(), length(values)), names(values))
   for (value in names(values)) {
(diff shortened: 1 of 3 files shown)


hadley:
That’s not quite the right fix - there should be a function that’s called earlier to make sure they are in standard positions. I think it probably was written before I added support for multiple names_to and so needs to be called in a different place.

romatik:
`check_spec` checks for `.name` and `.value` to be first two columns, but it is called only when `spec` is provided explicitly. So fix might be to use `check_spec` regardless of whether `spec` was provided or not.

This is where `check_spec` checks for that:
https://github.com/tidyverse/tidyr/blob/e92e74e3c213f5073c8cf78a27ed8eaa93007c0d/R/pivot.R#L17

And here it is called only in the `else` branch, so can be moved out of `if`:
https://github.com/tidyverse/tidyr/blob/e92e74e3c213f5073c8cf78a27ed8eaa93007c0d/R/pivot-long.R#L69

romatik:
@hadley, what do you think of the fix now? 
Travis builds failed, but I'm pretty sure it has nothing to do with the change I've made.

hadley:
Perfect, thanks!

R/pivot-long.R
```diff
@@ -65,6 +65,7 @@ pivot_longer <- function(df,
       names_pattern = names_pattern,
       col_ptypes = col_ptypes
     )
+    spec <- check_spec(spec)
   } else {
     spec <- check_spec(spec)
   }
```
hadley: Oh actually, now that `check_spec()` is called in both branches, you could simplify the code but doing the checking after the if statement. 

And could you please make the same change in `pivot_wider()`?

romatik: Yes, I was thinking to move it out as well, but then `else` branch doesn't do anything. But if you are OK, with that, I'll move it.
Will update `pivot_wider`.

hadley:
Thanks!!

PR #615, opened 2019-04-24
Fixed unnest .sep used on empty tibbles

mitchelloharawild:
## About

Unnesting an empty tibble errors when a separator is specified, as `paste("a", chr(), sep = "_")` gives `"a_"` not `chr()`.

## Before
``` r
library(tidyr)
tibble(data = list(tibble())) %>% 
  unnest(data, .sep = "_")
#> `nm` must be `NULL` or a character vector the same length as `x`
```

<sup>Created on 2019-04-24 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

---

## After

``` r
library(tidyr)
tibble(data = list(tibble())) %>% 
  unnest(data, .sep = "_")
#> # A tibble: 0 x 0
```

<sup>Created on 2019-04-24 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

### DIFF
--- a/R/unnest.R
+++ b/R/unnest.R
@@ -126,7 +126,7 @@ unnest.data.frame <- function(data, ..., .drop = NA, .id = NULL,
     unnested_dataframe <- imap(
       unnested_dataframe,
       function(df, name) {
-        set_names(df, paste(name, names(df), sep = .sep))
+        set_names(df, sprintf("%s%s%s", name, .sep, names(df)))
       }
     )
   }


hadley:
Unfortunately I'm just about to completely rip out that implementation. Can you see if the same problem occurs in `unnest2()`? If it does, it's probably better to file an issue, since I'm elbow deep in this part of the code right now.

mitchelloharawild:
Yep - the `paste()` call in `check_unpack()` causes the same issue. The new implementation looks much neater/modular than before. :+1:

``` r
library(tidyr)
tibble(data = list(tibble())) %>% 
  unnest2(data, names_sep = "_")
#> Error in attr(x, "names") <- as.character(value): 'names' attribute [1] must be the same length as the vector [0]
```

<sup>Created on 2019-04-24 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

Issue #618, opened 2019-04-26
demo("dadmom", package = "tidyr") fails [easy fix]


jmcastagnetto:
The `dadmomw.dta` file is now available via HTTPS not HTTP.

``` r
demo("dadmom", package = "tidyr")
#> 
#> 
#>  demo(dadmom)
#>  ---- ~~~~~~
#> 
#> > library(tidyr)
#> 
#> > library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
#> 
#> > dadmom <- foreign::read.dta("http://www.ats.ucla.edu/stat/stata/modules/dadmomw.dta")
#> Error in foreign::read.dta("http://www.ats.ucla.edu/stat/stata/modules/dadmomw.dta"): not a Stata version 5-12 .dta file
```

<sup>Created on 2019-04-26 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

Fix is quick: just modify the url in https://github.com/tidyverse/tidyr/blob/17bb14c2586b924786512ea1281862186f0c5951/demo/dadmom.R#L4

(I'm too lazy to make a fork for a 1 char fix :-)

cderv:
When I try I get a Invalide certificate Error in my browser. And R sees this too (error message in french sorry...)
``` r
foreign::read.dta("https://www.ats.ucla.edu/stat/stata/modules/dadmomw.dta")
#> Warning in download.file(file, tmp, quiet = TRUE, mode = "wb"):
#> InternetOpenUrl failed: 'Le nom de l'h<U+00F4>te du certificat n'est pas valide ou
#> ne correspond pas'
#> Error in download.file(file, tmp, quiet = TRUE, mode = "wb"): cannot open URL 'https://www.ats.ucla.edu/stat/stata/modules/dadmomw.dta'
```
Does it work for you without error ? 

<sup>Created on 2019-04-27 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>


> (I'm too lazy to make a fork for a 1 char fix :-)

Know that you can do this without explicitly making a fork by using the github interface for such a PR.
https://help.github.com/en/articles/editing-files-in-another-users-repository

This is very useful for such edition (and lazy people) ! 😉

jmcastagnetto:
[...]

PR #620, opened 2019-04-28
fix for #618

jmcastagnetto:
The URL for the data file changed to a new subdomain and protocol (HTTPS)


hadley:
All the demos are very dated, so I think I'm going to just remove them all.

PR #625, opened 2019-05-07
Fix typo in extract() documentation

tdjames1:


batpigandme:
Thanks for the fix. I just ran `devtools::document()`, so that it will be updated in the documentation as well.

hadley:
Thanks!

PR #633, opened 2019-05-17
spread (not gather)  is equivalent to pivot_wider

SimonCoulombe:


hadley:
I love that I keep getting this wrong 🤣

PR #667, opened 2019-07-01
fix typo in warning message

hplieninger:
Copy/pasting the current warning throws an unrelated error message, because `unnest()` looks for a column named `name_sep` (instead of recognizing the argument `names_sep`).


hadley:
Thanks!

Issue #666, opened 2019-07-01
More intuitive example for "Wider" section of "Manual Spec" in pivoting vignette


atusy:
In the "Wider" section of "Manual Spec" in "Pivoting" vignette, customized manual spec is created by a combination of `expand` and `mutate`.
The order of rows and columns of the customized spec are different from that of the basic custom spec.
It is hard to compare them.

I suggest mutating the result of `pivot_wider_spec` is much easier and safer.

By mutating, the order of columns and rows in customized manual spec and basic spec are identical.
This is significantly easy to compare.

``` r
library(tidyr)
library(dplyr, warn.conflicts = FALSE)
  
# Base spec
us_rent_income %>% 
  pivot_wider_spec(names_from = variable, values_from = c(estimate, moe))
#> # A tibble: 4 x 3
#>   .name           .value   variable
#>   <chr>           <chr>    <chr>   
#> 1 estimate_income estimate income  
#> 2 estimate_rent   estimate rent    
#> 3 moe_income      moe      income  
#> 4 moe_rent        moe      rent

# Customized manual spec by Hadley
spec <- us_rent_income %>% 
  expand(variable, .value = c("estimate", "moe")) %>% 
  mutate(
    .name = paste0(variable, ifelse(.value == "moe", "_moe", ""))
  )
spec
#> # A tibble: 4 x 3
#>   variable .value   .name     
#>   <chr>    <chr>    <chr>     
#> 1 income   estimate income    
#> 2 income   moe      income_moe
#> 3 rent     estimate rent      
#> 4 rent     moe      rent_moe

# Customized manual spec by Atusy
spec2 <- us_rent_income %>%
  pivot_wider_spec(names_from = variable, values_from = c(estimate, moe)) %>%
  mutate(.name = paste0(variable, ifelse(.value == "moe", "_moe", "")))
spec2
#> # A tibble: 4 x 3
#>   .name      .value   variable
#>   <chr>      <chr>    <chr>   
#> 1 income     estimate income  
#> 2 rent       estimate rent    
#> 3 income_moe moe      income  
#> 4 rent_moe   moe      rent

# Check results
hadley <- us_rent_income %>% pivot_wider(spec = spec)
atusy <- us_rent_income %>% pivot_wider(spec = spec2)
identical(hadley, atusy[names(hadley)])
#> [1] TRUE
```

<sup>Created on 2019-07-01 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

[...]

PR #669, opened 2019-07-01
Better example for Wider section of Manual Spec in pivoting vignette

atusy:
Modified example code blocks based on #666, and descriptions as well.

Beside what is discussed in #666, I reordered column names described in L505, so that the order is consistent to the output data frame.


hadley:
Thanks!

For future reference, if you include `Fixes #xyz` in a commit message (or the text of the PR), the corresponding issue will be automatically closed when the PR is merged.

atusy:
Thank you too!

How nice the `Fixes #xyz`! I will from the next time : )

Issue #665, opened 2019-07-01
Better error for pivot_longer() when no column names are given


yutannihilation:
As `pivot_longer()` accepts select helpers, it might be a common mistake to pass 0 columns to `cols` argument. I think this error message can be improved. (Might this be a good issue for tidy dev day?)

``` r
library(tidyr)

iris %>% 
  pivot_longer(matches("foo"))
#> Error in `[[<-.data.frame`(`*tmp*`, ".value", value = "value"): replacement has 1 row, data has 0
```

<sup>Created on 2019-07-01 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

hadley:
Good catch, thanks!

PR #670, opened 2019-07-01
pivot_longer(cols = NULL) as default

### LINKED ISSUE #661: Define a default value of the cols argument of pivot_longer
`pivot_longer`'s `cols` argument currently misses a default argument.
What about `everything()` because

- it is consistent to what `gather` does.
- it avoids RStudio saying "argument 'cols' is missing, with no default" when using manual spec.

image

atusy:
- It returns error if cols nor spec is specified by user.
- It avoids RStudio IDE saying "argument 'cols' is missing. with no default" when using manual spec

Fixes #661

### DIFF
--- a/R/pivot-long.R
+++ b/R/pivot-long.R
@@ -95,7 +95,7 @@
 #'    names_pattern = "(.)(.)"
 #'  )
 pivot_longer <- function(data,
-                         cols,
+                         cols = NULL,
                          names_to = "name",
                          names_prefix = NULL,
                          names_sep = NULL,


hadley:
Where does the check for `is.null(spec) && is.null(cols)` occur?

atusy:
At first, I thought just letting `cols = NULL` is enough because after #665 is fixed, both the `is.null(spec) && is.null(cols)` will tell no column names are given.

Now I agree that it's better to check arguments explicitly, and return error message like "Users must specify `cols` or `spec`" when needed.

If we make `cols = NULL` as a default value, I will make updates.
But I'd like to wait and see your reaction to @DavisVaughan 's comment (https://github.com/tidyverse/tidyr/issues/661#issuecomment-507653195).

atusy:
I close this PR because the spec functions will be pulled out from the pivot functions https://github.com/tidyverse/tidyr/issues/661#issuecomment-513978225

PR #671, opened 2019-07-03
fix typo in example section of tidyr's `extract()`

gederajeg:
In the example for tidyr's `extract()`, it is written `library(dplyr)` rather than `library(tidyr)`. I'm using tidyr version 0.8.3


hadley:
R automatically loads the package, so I think the correct fix here is to remove that line altogether. Thanks for spotting the problem!

PR #678, opened 2019-07-08
Improve error message when `matches()` is used for `col` argument in `pivot_longer()`

brentthorne:
Improve error message when `matches()` is used for `col` argument in `pivot_longer()`, see #665.

### DIFF
--- a/R/pivot-long.R
+++ b/R/pivot-long.R
@@ -208,6 +208,11 @@ pivot_longer_spec <- function(data, cols,
     names[[col]] <- vec_cast(names[[col]], names_ptypes[[col]])
   }
 
+  # test the value of `col when it` uses `matches()` arg
+  if (length(cols) == 0) {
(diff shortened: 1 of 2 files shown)


R/pivot-long.R
```diff
@@ -208,6 +208,11 @@ pivot_longer_spec <- function(data, cols,
     names[[col]] <- vec_cast(names[[col]], names_ptypes[[col]])
   }
 
+  # test the value of `col when it` uses `matches()` arg
+  if (length(cols) == 0) {
+    abort(glue::glue("Val ue in `matches()` has no corresponding value in `data`."))
```
hadley: I think this needs to be a bit more general because it could be the result of any tidyselect function failing to match. Maybe something like `` `cols` must select at least one column. ``?

Do you want to also have a go at adding a test?

brentthorne: I gave it a try! Thanks for the feedback.

hadley:
Thanks @brentthorne!

PR #690, opened 2019-07-25
patch full_seq to work for differences that are negative (but within tolerance)

### LINKED ISSUE #657: full_seq only checks for tolerance in one direction
`full_seq()` has a `tol` parameter that is the tolerance for checking periodicity. However, the implementation of the check, which uses the `%%` operator only checks for tolerances in one direction:

``` r
tidyr::full_seq(c(0, 10, 20), 11, tol = 2)
#> Error: `x` is not a regular sequence.
```

<sup>Created on 2019-06-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Here's an updated `full_seq.numeric()` that I think solves the issue, and also pads the range if the last element falls short of being k*period more than the first element:

``` r
full_seq.numeric.patched <- function(x, period, tol = 1e-6) {
    rng <- range(x, na.rm = TRUE)
    if (any((x - rng[1]) %% period > tol & 
            period - (x - rng[1]) %% period > tol)) {
        stop("`x` is not a regular sequence.", call. = FALSE)
    }
    
    if (period - ((rng[2] - rng[1]) %% period) <= tol)
        rng[2] <- rng[2] + tol
            
    seq(rng[1], rng[2], by = period)
}

full_seq.numeric.patched(c(0, 10, 20), 11, tol = 2)
#> [1]  0 11 22
full_seq.numeric.patched(c(0, 10, 20), 11, tol = 1.8)
#> Error: `x` is not a regular sequence.
full_seq.numeric.patched(c(0, 10, 17), 11, tol = 5)
#> [1]  0 11 22
full_seq.numeric.patched(c(0, 10, 16), 11, tol = 5)
#> [1]  0 11
```

<sup>Created on 2019-06-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

ha0ye:
(the %% operator only returns differences in one direction)
closes #657

### DIFF
--- a/R/seq.R
+++ b/R/seq.R
@@ -18,9 +18,15 @@ full_seq <- function(x, period, tol = 1e-6) {
 #' @export
 full_seq.numeric <- function(x, period, tol = 1e-6) {
   rng <- range(x, na.rm = TRUE)
-  if (any((x - rng[1]) %% period > tol)) {
+  if (any(((x - rng[1]) %% period > tol) &
+          (period - (x - rng[1]) %% period > tol))) {
     stop("`x` is not a regular sequence.", call. = FALSE)
   }
+
+  # in cases where the last element is within tolerance, pad it so that
+  #   the output length is correct
+  if (period - ((rng[2] - rng[1]) %% period) <= tol)
+    rng[2] <- rng[2] + tol
 
   seq(rng[1], rng[2], by = period)
 }
--- a/R/seq.R
+++ b/R/seq.R
@@ -25,8 +25,9 @@ full_seq.numeric <- function(x, period, tol = 1e-6) {
 
   # in cases where the last element is within tolerance, pad it so that
   #   the output length is correct
-  if (period - ((rng[2] - rng[1]) %% period) <= tol)
+  if (period - ((rng[2] - rng[1]) %% period) <= tol) {
     rng[2] <- rng[2] + tol
+  }
 
   seq(rng[1], rng[2], by = period)
 }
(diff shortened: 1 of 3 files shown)


R/seq.R
```diff
@@ -18,10 +18,16 @@ full_seq <- function(x, period, tol = 1e-6) {
 #' @export
 full_seq.numeric <- function(x, period, tol = 1e-6) {
   rng <- range(x, na.rm = TRUE)
-  if (any((x - rng[1]) %% period > tol)) {
+  if (any(((x - rng[1]) %% period > tol) &
```
hadley: Is it possible to express this more simply by using `abs()`?

R/seq.R
```diff
@@ -18,10 +18,16 @@ full_seq <- function(x, period, tol = 1e-6) {
...
+          (period - (x - rng[1]) %% period > tol))) {
     stop("`x` is not a regular sequence.", call. = FALSE)
   }
 
+  # in cases where the last element is within tolerance, pad it so that
+  #   the output length is correct
+  if (period - ((rng[2] - rng[1]) %% period) <= tol)
```
hadley: Can you please wrap the block in `{}`, as described in https://style.tidyverse.org/syntax.html#inline-statements

R/seq.R
ha0ye: If we assume period > 0, then %% returns a value between [0, period), and we want to filter on the values at the edges... I suppose you could do something like abs(period/2 - (x - rng[1]) %% period)) <= (period/2 - tol). I think that's marginally faster in terms of vectorized operations, but hard to say if that's "more simply expressed"...

hadley: Yeah, sounds like the current code is best

hadley:
Thanks for contributing!

Issue #692, opened 2019-07-25
Process revdep failures


hadley:
https://docs.google.com/spreadsheets/d/13XXN8sepsc0Ftl5nVz4YdJCONbD_k00HOZUxbPosThM/edit#gid=0

Each failure needs to be triaged and a resolution found; for failures related to the interface changes in `nest()` and `unnest()` the plan is to help individual package authors with the needed fixes. I'm reasonably confident the deprecation warnings should catch problems in existing scripts.

jennybc:
I'm using some download stats to prioritize my efforts:

revdep-sitrep.R
================
jenny
2019-07-30

``` r
library(commonmark)
library(xml2)
library(tidyverse)
library(cranlogs)

new_problems_path <- here::here("revdep/problems.md")
md <- readLines(new_problems_path)
pkg <- md %>%
  str_subset("^#[^#]") %>%
  str_extract("[[:alnum:]]+")

dl <- cran_downloads(when = "last-month", packages = pkg)

dl %>%
  count(package, wt = count) %>%
  arrange(desc(n)) %>%
  mutate(
    prop = n / sum(n),
    cum_prop = cumsum(prop)
  ) %>%
  print(n = 20)
#> # A tibble: 68 x 4
#>    package           n    prop cum_prop
#>    <chr>         <dbl>   <dbl>    <dbl>
#>  1 modelr       205720 0.390      0.390
#>  2 ggpubr        94105 0.179      0.569
#>  3 recipes       93673 0.178      0.747
#>  4 survminer     18635 0.0354     0.782
#>  5 d3r           17290 0.0328     0.815
#>  6 sunburstR     15903 0.0302     0.845
#>  7 sjstats       14758 0.0280     0.873
#>  8 sjPlot        11225 0.0213     0.894
#>  9 tidyquant      5698 0.0108     0.905
#> 10 gutenbergr     4081 0.00774    0.913
#> 11 fuzzyjoin      3146 0.00597    0.919
#> 12 tibbletime     3092 0.00587    0.925
#> 13 widyr          2995 0.00568    0.930
#> 14 ggstatsplot    2955 0.00561    0.936
#> 15 tsibble        2928 0.00556    0.942
#> 16 bench          2671 0.00507    0.947
#> 17 broomExtra     2484 0.00471    0.951
#> 18 groupedstats   1846 0.00350    0.955
#> 19 ggalluvial     1789 0.00339    0.958
#> 20 anomalize      1640 0.00311    0.961
[...]

PR #706, opened 2019-08-08
Vignette on usage and migration

jennybc:
Very much a WIP but happy for others to take it from here.

There's content here that arguably belongs (?also or instead?) elsewhere, such as in https://tidyeval.tidyverse.org or https://principles.tidyverse.org or https://r-pkgs.org. But I don't think that should prevent us from getting it written down. We can move it later and/or repeat the content in more than one place.


jennybc:
Note I use the tidyverse metapackage. I use both dplyr and purrr., which really makes sense. Not sure what to do about that.

DavisVaughan:
In "why it changed" for `nest()` I might put that you can now do a multi-nest? I found that rather neat: `nest(mtcars, x1 = c(mpg, cyl), x2 = c(vs, disp))`. This also seems related to why `.key` is no longer enough on its own (can't specify both new column names)

jennybc:
Good idea, @DavisVaughan. There is an example along those lines but in the `unnest()` section (it shows a multi-`unnest()`). That could make a first appearance earlier.

DavisVaughan:
Is it worth putting in a usethis issue for `use_tidyselect()`?
https://github.com/tidyverse/tidyr/blob/00a66b44d2f304a73e953dd73139db61ed9cc289/R/tidyr.R#L29-L53

jennybc:
> Is it worth putting in a usethis issue for `use_tidyselect()`?

Seems sensible to me. Needs to be thought about in the context of the existing `use_tidy_eval()`.

hadley:
Fixes #692 

Yes, an issue in usethis would be handy; I'm just linking to what tidyr does in the meantime.

vignettes/usage-and-migration.Rmd
```diff
@@ -0,0 +1,340 @@
...
+
+If you use travis already, we **strongly** recommend adding a build that tests with the development version of tidyr; see above for details.
+
+### Conditional code
+
+Sometimes you'll be able to write code that works with v0.8.3 and v1.0.0. But this often forces you to write code that's not particularly natural for either version and you'd be better off to (temporarily) have separate code paths, each containing non-contrived code. You get to re-use your existing code in the "old" branch, which will eventually be phased out, and write clean, forward-looking code in the "new" branch.
+
+The basic template looks like this:
+
+```{r}
```
jennybc: Needs `eval = FALSE`

vignettes/usage-and-migration.Rmd
```diff
@@ -0,0 +1,340 @@
...
+tidyr_new_interface <- function() {
+  packageVersion("tidyr") > "0.8.99"
```
jennybc: Should we mention the importance of doing this in a function? I.e. so that it's determined at run time not install time? Otherwise, I think people could easily get creative and "simplify" this.

vignettes/in-packages.Rmd
```diff
@@ -182,17 +188,17 @@ my_function_inside_a_package <- function(...)
 
   if (tidyr_new_interface()) {
     # Freshly written code for v1.0.0
-    out <- tidyr::nest_legacy(df, x, y, z)
+    out <- tidyr::nest(df, data = one_of("x", "y", "z"))
```
jennybc: I have two reasons for changing this:
1. We haven't mentioned `nest_legacy()` yet, so it's a bit out of the blue.
1. Seems like we should show real code, not code that we describe as a "quick and dirty fix".

vignettes/in-packages.Rmd
```diff
@@ -182,17 +188,17 @@ my_function_inside_a_package <- function(...)
...
   } else {
     # Existing code for v0.8.3
-    out <- tidyr::nest(df, c(x, y, z))
+    out <- tidyr::nest(df, x, y, z)
```
jennybc: I think this was just wrong?

hadley: I think it worked but wasn't idiomatic previously

vignettes/in-packages.Rmd
```diff
@@ -0,0 +1,399 @@
...
+library(purrr)
+
+(mini_iris <- iris %>% 
+    as_tibble() %>% 
+    .[c(1, 2, 51, 52, 101, 102), ])
+```
+
+## Using tidyr in packages
+
+Compared to dplyr and ggplot2, most tidyr functions have __select semantics__, not the more common mutate semantics. This means that you typically provide tidyr functions with an expression that works with column names (like `starts_with("x")` or `a:c`) not with column values (like `y = x * 2`). As a consequence, it's worthwhile to clarify the best patterns for calling tidyr functions inside another package.
```
lionel-: > not the more common mutate semantics

Mutate semantics are also about vectorisation, and sequential updating. Should this be "action" semantics instead?

vignettes/in-packages.Rmd
```diff
@@ -0,0 +1,399 @@
...
+
+What changed:
+
+*   The to-be-nested columns are no longer accepted as "loose parts".
+*   The new list-column's name is no longer provided via the `.key` argument.
+*   Now we use a construct like this: `new_col = <something about existing cols>`.
+  
+Why it changed:
+
+*   The use of `...` to make data is a problematic pattern we're moving away from.
```
lionel-: I think `...` to make data is exactly when it's not problematic. It's the use of `...` for metadata (e.g. a selection) that's problematic, IIUC.

vignettes/in-packages.Rmd
jennybc: Yeah ok. It is v0.8.3 code, written with a v1.0.0 sensibility.

PR #707, opened 2019-08-08
Fix tests for dev tibble

krlmlr:
Closes tidyverse/tibble#630.


hadley:
I think tibble is at fault here; lets discuss in the other issue.

PR #708, opened 2019-08-08
Fix test warning

krlmlr:
This test tests an error that comes from tibble. Is it worth to catch and rethrow an own error here?


hadley:
Thanks — definitely not worthwhile to rethrow here; the intent of the test is to ensure that `name_repair` is passed along appropriate

PR #728, opened 2019-09-09
Make copy edits to `NEWS`.

batpigandme:


batpigandme:
**Don't merge yet!**
I have a couple of questions that I'll add as review comments to my own PR.

NEWS.md
```diff
@@ -25,7 +25,7 @@ See `vignette("in-packages")` for a detailed transition guide.
   request the previous approach.
 
 * `unnest_()`/`nest_()` and the lazyeval methods for `unnest()`/`nest()` are 
-  now defunct. They have been deprecated for some time, and since the interface
+  now defunct. They have been deprecated for some time, and, since the interface
```
batpigandme: Would it be worth adding a link to the lifecycle package's [definitions](https://lifecycle.r-lib.org/articles/lifecycle.html) of deprecated and defunct?

NEWS.md
```diff
@@ -54,7 +54,7 @@ tidyr contains four new functions to support **rectangling**, turning a deeply n
 
 `unnest_longer()` and `unnest_wider()` make it easier to unnest list-columns of vectors into either rows or columns (#418). `unnest_auto()` automatically picks between `_longer()` and `_wider()` using a little magic.
```
batpigandme: Do we have a PR number of "using a little magic"? Or is that why we're using the phrase?

NEWS.md
hadley: Yeah, that'd be great

NEWS.md
hadley: There's no PR because it's all new code. Maybe "heuristics based on the presence of common names" would be better?

hadley:
Thanks!

PR #729, opened 2019-09-10
Fix compatibility layer of `.key`

lionel-:
Part of r-spatial/sf#1068

`.key` needs to be quasiquoted for compatibility. Thanks to @yutannihilation for spotting the problem!

### DIFF
--- a/R/nest.R
+++ b/R/nest.R
@@ -136,12 +136,12 @@ nest.data.frame <- function(.data, ..., .key = deprecated()) {
     .data <- as_tibble(.data)
   }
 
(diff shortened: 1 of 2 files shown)


yutannihilation:
Thanks, but, is this supposed to work? I was trying to fix on sf's side by removing quasiquotation.

lionel-:
Yes the argument used to be quasiquoted. It should work with a deprecation warning.

It is still a good idea to update sf not to use `.key` as it is now deprecated.

yutannihilation:
I see.

yutannihilation:
Is this likely to be merged and released in the near future?

PR #766, opened 2019-09-30
Update tidy-data.Rmd

william3031:
replaced `stringr` function (str_remove) with `base` function (`gsub`).


william3031:
testing

PR #767, opened 2019-09-30
removed stringr use

william3031:
removed `str_remove` and added `gsub`.


hadley:
You missed a couple of other places, and it was easy enough to do, so I made the fix directly. Thanks!

PR #768, opened 2019-10-02
Explicitly state everything() will pivot all cols

martinjhnhadley:
In the now outdated `gather()` function it was very simple to pivot on all cols:

```r
anscombe %>%
  gather()
```
In my experience, it's quite rare for a general tidyverse user to know what `everything()` does - or that it even exists.


hadley:
Thanks!

PR #777, opened 2019-10-14
Delete one unnecessary period

xkdog:
Delete unnecessary period at the of the sentence "Every row is an observation..".


jennybc:
Thanks!

`README.md` is actually generated from `README.Rmd`, so I've made the same edit there. That is really the source file, in this case.

xkdog:
OK！

PR #781, opened 2019-10-17
Added additional details in unnest warning message

william3031:
I had the old message come up when writing some code and I had to think about where the message was coming from. This clarifies where it is coming from. It might be useful.


hadley:
Thanks!

PR #785, opened 2019-10-19
Improve documentation for extract()

### LINKED ISSUE #693: extract() does not honor regex options
`extract()` does not honor regex options. Here's a reprex:

``` r
# Example data.
df <- data.frame(x="A")
# Extract the first case-insensitive (English) letter.
out <- tidyr::extract(df,
  x,
  c("A"),
  stringr::regex("([a-z])", ignore_case = TRUE)
)

# Expect this to be "A"
out$A
#> [1] NA
```

I saw #223, but I think this is a different issue.

I'm really not familiar with R as a language or the tidyverse as a system so I'm not sure if this is intentional behavior, but I wasn't able to find documentation that `extract()` treats the regex as a simple character vector. The documentation for `extract()` refers to it only as "a regular expression."

The apparent issue is this line in extract.R:

``` r
  matches <- stringi::stri_match_first_regex(x, regex)[, -1, drop = FALSE]
```

I would expect it to pass the options of the regex to `stri_match_first_regex` using its `opts_regex` parameter.

As far as I can tell, tidyr doesn't import stringr at all. So `extract()`s lack of awareness of stringr's regex types could be an intentional design choice of the tidyverse. But that should still be better documented.

I attempted to write a fix with the assumption that use of stringr was allowed, but ran into the problem that stringr's method of getting options from a regex seems to be `opts()`, which is not exported. In any case, my changes were

``` r
# extract.R:57
# Doesn't work. `stringr::opts()` not available.
  matches <- stringi::stri_match_first_regex(x,
    regex,
    opts_regex = stringr::opts(regex)
  )[, -1, drop = FALSE]
```

``` r
# test-extract.R:48
test_that("honors regex options", {
  df <- data.frame(x="A")
  out <- df %>% extract(x, c("A"), stringr::regex("([a-z])", ignore_case = TRUE))
  expect_equal(out$A, "A")
})

```

namelessjon:
Fixes #693

Note that we're using a stringi regular expression.  Also note it has to be a string, not something like a stringr::regex


hadley:
Hmm this isn't quite right because I gave a bad explanation in the linked issue. Sorry 😞

PR #796, opened 2019-10-30
Dory is not a clownfish

BrianDiggs:
For the examples for hoist, the species for Dory is listed as clownfish. Marlin and Nemo are clownfish while Dory is a blue tang, or a Pacific Regal Blue Tang. I went with the shorter version.


hadley:
Thanks!

PR #807, opened 2019-11-20
Preserve emptiness in hoisted and unnested list-cols

### LINKED ISSUE #806: Rectanglers over-eager to git rid of emptiness
Adapting the Toothless & Dory example, but where I've removed the `films` metadata for Toothless. When I `hoist()` or `unnest_wider()`, `films` is destined to become a list-column. I think the entry for Dory should be `NULL`, not an `unspecified` vector of length 1. Or maybe even an `unspecified` vector of length 0? Or a character vector of length 0?

``` r
library(tidyverse)

df <- tibble(
  character = c("Toothless", "Dory"),
  metadata = list(
    list(
      species = "dragon",
      color = "black"
    ),
    list(
      species = "clownfish",
      color = "blue",
      films = c("Finding Nemo", "Finding Dory")
    )
  )
)
df
#> # A tibble: 2 x 2
#>   character metadata        
#>   <chr>     <list>          
#> 1 Toothless <named list [2]>
#> 2 Dory      <named list [3]>

df %>% 
  hoist(metadata, films = "films")
#> # A tibble: 2 x 3
#>   character films     metadata        
#>   <chr>     <list>    <list>          
#> 1 Toothless <???>     <named list [2]>
#> 2 Dory      <chr [2]> <named list [2]>

(df2 <- df %>% 
  unnest_wider(metadata))
#> # A tibble: 2 x 4
#>   character species   color films    
#>   <chr>     <chr>     <chr> <list>   
#> 1 Toothless dragon    black <???>    
#> 2 Dory      clownfish blue  <chr [2]>
df2$films
#> [[1]]
#> <unspecified> [1]
#> 
#> [[2]]
#> [1] "Finding Nemo" "Finding Dory"
lengths(df2$films)
#> [1] 1 2
```

<sup>Created on 2019-11-20 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

jennybc:
Closes #806. Rather, ideally, it would.

I'm not going to finish this. Everything I do to fix it, causes trouble elsewhere. I think you @hadley have to decide how you want to fix it. >= 3 helpers are interacting, which are also used in other functions.

Diagnosis:

`simplify_col()` calls `init_col()`:

https://github.com/tidyverse/tidyr/blob/cb512471f2a38871fa168c21590925364a3e5c4a/R/rectangle.R#L327-L329

`init_col()` turns `NULL`s and anything of size 0 into something of size 1.

https://github.com/tidyverse/tidyr/blob/c73979dbe0bc88a8855cf4dc6a6cc9bdb7c2ecfe/R/utils.R#L102-L110

But that's delicate to change that because `vec_simplify()` won't simplify unless every element has size 1.

https://github.com/tidyverse/tidyr/blob/cb512471f2a38871fa168c21590925364a3e5c4a/R/rectangle.R#L340-L344

### DIFF
--- a/tests/testthat/test-rectangle.R
+++ b/tests/testthat/test-rectangle.R
@@ -47,6 +47,16 @@ test_that("input validation catches problems", {
   expect_error(df %>% hoist(x, "a"), "named")
 })
 
+# https://github.com/tidyverse/tidyr/issues/806
+test_that("hoisted list-col represents emptiness with a length-0 element", {
+  df <- tibble(x = list(
+    list(universal = "a", patchy = 1:2),
+    list(universal = "b")
+  ))
+  out <- df %>% hoist(x, patchy = "patchy")
+  expect_length(out$patchy[[2]], 0)
+})
+
 # strike ------------------------------------------------------------------
 
 test_that("strike can remove using a character vector", {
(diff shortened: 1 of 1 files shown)


hadley:
Fixed with a total revamp of `vec_simplify()`

Issue #734, opened 2019-09-12
Corner case: nest(data = everything()) with zero-row input


krlmlr:
Do we expect `nest()` to return the same number of rows as `summarize()` in all cases? This currently doesn't work with empty groups or empty data frames:

``` r
library(tidyverse)

tibble(a = factor(levels = letters[1:3])) %>%
  group_by(a, .drop = FALSE) %>%
  summarize(n = n())
#> # A tibble: 3 x 2
#>   a         n
#>   <fct> <int>
#> 1 a         0
#> 2 b         0
#> 3 c         0

tibble(a = factor(levels = letters[1:3])) %>%
  group_by(a, .drop = FALSE) %>%
  nest()
#> # A tibble: 0 x 2
#> # Groups:   a [3]
#> # … with 2 variables: a <fct>, data <list<df[,0]>>

tibble(a = character()) %>%
  summarize(n = n())
#> # A tibble: 1 x 1
#>       n
#>   <int>
#> 1     0

tibble(a = character()) %>%
  nest(data = everything())
#> # A tibble: 0 x 1
#> # … with 1 variable: data <list<df[,1]>>
```

<sup>Created on 2019-09-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

hadley:
I don't think this can be resolved until we have a version of `vec_unique()` that uses that same principles as `group_by()` `.drop = FALSE`

krlmlr:
I'm not sure we can implement this correctly with `vec_unique()`. It looks like we need:

- special handling for the zero-row case for ungrouped data frames
- a dedicated implementation for grouped data frames that uses `group_vars()` and a variant of `map(group_vars$.rows, vec_slice(cols, .x))`

krlmlr:
Can we implement the second part with `group_nest()`?

krlmlr:
Looks like `nest_by()` is the better alternative for programming. All corner cases work better there.

Should we base `nest()` on `nest_by()` to achieve the same guarantees here, for free?

``` r
library(tidyverse)

tibble(a = factor(levels = letters[1:3])) %>%
  nest_by(a, .drop = FALSE)
#> # A tibble: 3 x 2
#> # Rowwise:  a
#>   a                   data
#>   <fct> <list<tbl_df[,0]>>
#> 1 a                [0 × 0]
#> 2 b                [0 × 0]
#> 3 c                [0 × 0]

tibble(a = character()) %>%
  nest_by()
[...]

PR #809, opened 2019-11-25
Fix corner case: nesting all columns of an empty data frame

krlmlr:
Addresses part of #734.


tests/testthat/test-nest.R
```diff
@@ -68,7 +68,7 @@ test_that("nesting works for empty data frames", {
 
   out <- nest(df, data = c(x, y))
   expect_named(out, "data")
-  expect_equal(nrow(out), 0L)
+  expect_equal(nrow(out), 1L)
```
hadley, line 71: I don't understand why this yields 1 row but the example yields 0 rows? 

I'm no longer confident that the `summarise()` justification is quite right (because `nest()` isn't necessarily perfectly aligned with dplyr reasoning), so unless you feel strongly about this, I'd rather close, preferring to fix related issues in `nest_by()`.

krlmlr:
I'd rather reimplement using `nest_by()` .

Issue #688, opened 2019-07-24
`unnest` removes `rsample` attributes


IndrajeetPatil:
I am not sure if this is intended behavior, but `rsample` + `tidyr::unnest()` workflow doesn't seem to work.

Let's say I want to calculate bootstrap confidence intervals for `lm` model using `rsample`.

Without unnesting the list column, everything works as expected-

``` r
library(tidyverse)
library(rsample)

set.seed(123)
(df_nested <-
  rsample::bootstraps(
    data = iris,
    times = 100,
    apparent = TRUE
  ) %>% 
  dplyr::mutate(
    .data = .,
    results = purrr::map(
      .x = splits,
      .f = ~ broom::tidy(stats::lm(formula = Sepal.Length ~ Species, data = .))
    )
  ) )
#> # Bootstrap sampling with apparent sample 
#> # A tibble: 101 x 3
#>    splits           id           results         
#>  * <list>           <chr>        <list>          
#>  1 <split [150/56]> Bootstrap001 <tibble [3 x 5]>
#>  2 <split [150/58]> Bootstrap002 <tibble [3 x 5]>
#>  3 <split [150/50]> Bootstrap003 <tibble [3 x 5]>
#>  4 <split [150/58]> Bootstrap004 <tibble [3 x 5]>
#>  5 <split [150/56]> Bootstrap005 <tibble [3 x 5]>
#>  6 <split [150/55]> Bootstrap006 <tibble [3 x 5]>
#>  7 <split [150/56]> Bootstrap007 <tibble [3 x 5]>
#>  8 <split [150/52]> Bootstrap008 <tibble [3 x 5]>
#>  9 <split [150/60]> Bootstrap009 <tibble [3 x 5]>
#> 10 <split [150/55]> Bootstrap010 <tibble [3 x 5]>
#> # ... with 91 more rows

class(df_nested)
#> [1] "bootstraps" "rset"       "tbl_df"     "tbl"        "data.frame"

rsample::int_pctl(df_nested, results)
#> Warning: Recommend at least 1000 non-missing bootstrap resamples for terms:
#> `(Intercept)`, `Speciesversicolor`, `Speciesvirginica`.
#> # A tibble: 3 x 6
#>   term              .lower .estimate .upper .alpha .method   
#>   <chr>              <dbl>     <dbl>  <dbl>  <dbl> <chr>     
#> 1 (Intercept)        4.91      5.00    5.10   0.05 percentile
[...]

Issue #800, opened 2019-11-16
Make pivot_longer and pivot_wider generic


hadley:

chester-gan:
Specific feature request: please enable pivot_longer's values_ptypes argument to handle exceptional cases whereby the input columns used to generate the variables generated from values_to are not of the same type.

hadley:
@chester-gan please file an issue with reprex; commenting on unrelated issues is not a good way to get help.

chester-gan:
> @chester-gan please file an issue with reprex; commenting on unrelated issues is not a good way to get help.

Noted, filed under reprex.

Issue #626, opened 2019-05-10
Feature request: new `pivot_*` to be S3 generics


earowang:
Can new `pivot_longer()` and `pivot_wider()` be S3 generics so that we can extend them to other tibble-like objects? Thanks.

hadley:
I'd prefer to not make them generic for at least one version. And you should be able to get a lot of genericity via vctrs.

edzer:
I'm confused: via vctrs one can't set the (sub)class of the object returned by `pivot_*`, right?

hadley:
@edzer right, but (in principle) you shouldn't need to for subclasses of data.frame/tibble as vctrs should take care of the details (it's totally possible it doesn't right now, but I'd prefer to attack it upstream before adding extension points in vctrs)

edzer:
See https://github.com/r-spatial/sf/issues/1149 : `vctrs` seems to work only on the columns, not on the object as a whole. It sounds like you take the standpoint of "the world doesn't need anything else than tibbles, everything special should be done inside columns", but I think both time series (tsibble) and spatial data (sf) have shown that that is not enough.

yutannihilation:
I don't see the whole picture yet, but at least tidyr seems not ready to accept other than `tibble` and `data.frame`. For example, `pivot_longer_spec()` probably needs either `.ptype` specified on `vec_cbind()`:
https://github.com/tidyverse/tidyr/blob/4618116a9296e0f11337c416966ea0f5a62f3732/R/pivot-long.R#L187-L192

or `vec_restore()` instead of this tibble-specific function:
https://github.com/tidyverse/tidyr/blob/4618116a9296e0f11337c416966ea0f5a62f3732/R/pivot-long.R#L194

Yet, it's still unclear to me if it's really possible to restore the result only by defining vctrs functions because, in tidyr's case, the shape of the input and that of output are different.

earowang:
It's also not clear to me that how we can just define `vctrs::vec_restore()` for our tibble-like objects to achieve different results as expected. For example, this is how tsibble currently handles `gather()` and `spread()`: `gather()` adds the new melted column to the key specification, and `spread()` removes the column from the key. Other `dplyr` methods also conditionally validate if the result is a valid tsibble.

```r
library(tidyr)
# reshaping examples from tidyr
stocks <- tsibble::tsibble(
  time = as.Date("2009-01-01") + 0:9,
  X = rnorm(10, 0, 1),
  Y = rnorm(10, 0, 2),
[...]

PR #812, opened 2019-11-26
Restore data frames generically

lionel-:
Depends on r-lib/vctrs#690.
Progress towards  #800 and #626.

Switch to generic approach to reconstruct data frames. Requires:

- base `[` method for reorganising data frames inside `df_cast()`.
- vctrs `vec_proxy()` and `vec_restore()` methods, through the use of `vec_cbind()` in `df_append()`.
- vctrs `vec_cast()` method for the actual restoration of semantics.

Data frames are restored by casting. The `df_cast()` method wraps `vec_cast()` with additional behaviour to deal with variable transmformations.

- The type of `to` is used to restore the original semantics of columns.

- The new variables (passed to `df_cast()` via `new_vars`) are removed from `to` before taking the common type (via `setdiff()`), this way they lose the original semantics they had in the input data frame. This also avoids common type errors since the new columns might have diffent types.

- The columns that were removed from the output are also removed (via `intersect()`) so they are not added back via the common type.

Since the new variables are removed with `[`, this approach relies on well-behaved `[` methods.

- We'll need to update the grouped-df method so it doesn't lose all grouping variables as soon as one is removed. It should retain as many grouping variables as possible.

- A quick test with sf data frames suggests it will sort of work once vctrs methods are implemented for sf data frames (I'll work on these next).

  However we do get warnings from the pivot implementation because `[.sf` always restores the geometry columns, which breaks the fundamental invariant of R programming that the result of `[` should always be as long as the index. I hope this can be fixed in sf eventually, I'll open an issue downstream. I also suspect operating on the geometry columns themselves will produce errors because of this `[.sf` behaviour. cc @edzer

### DIFF
--- a/R/extract.R
+++ b/R/extract.R
@@ -46,7 +46,7 @@ extract.data.frame <- function(data, col, into, regex = "([[:alnum:]]+)",
 
   new_cols <- str_extract(value, into = into, regex = regex, convert = convert)
   out <- append_df(data, new_cols, var, remove = remove)
-  reconstruct_tibble(data, out, if (remove) var else chr())
+  df_cast(out, data, if (remove) var else chr())
 }
 
 str_extract <- function(x, into, regex, convert = FALSE) {
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -37,7 +37,7 @@ Imports:
     tibble (>= 2.1.1),
     tidyselect (>= 0.2.5),
     utils,
-    vctrs (>= 0.2.0.9007),
+    vctrs (>= 0.2.0.9008),
     lifecycle
 Suggests: 
     covr,
@@ -56,4 +56,4 @@ LazyData: true
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.0.2
 Remotes:
-  r-lib/vctrs
+  lionel-/vctrs@add-grouped-df
--- a/R/append.R
+++ b/R/append.R
@@ -1,4 +1,5 @@
 append_df <- function(x, y, after = length(x), remove = FALSE) {
+  skip("FIXME: append")
   if (is.character(after)) {
     after <- match(after, dplyr::tbl_vars(x))
   } else if (!is.integer(after)) {
--- a/R/append.R
+++ b/R/append.R
@@ -1,5 +1,5 @@
+
 append_df <- function(x, y, after = length(x), remove = FALSE) {
-  skip("FIXME: append")
   if (is.character(after)) {
     after <- match(after, dplyr::tbl_vars(x))
   } else if (!is.integer(after)) {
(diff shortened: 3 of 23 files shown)


R/extract.R
```diff
@@ -45,7 +45,7 @@ extract.data.frame <- function(data, col, into, regex = "([[:alnum:]]+)",
 
   new_cols <- str_extract(value, into = into, regex = regex, convert = convert)
   out <- append_df(data, new_cols, var, remove = remove)
-  reconstruct_tibble(data, out, if (remove) var else chr())
+  df_cast(out, data, if (remove) var else chr())
```
hadley: If you split `x` into `a`, `b`, `c` and `x` is grouped, wouldn't you want the result to be grouped by `a`, `b`, `c`?

lionel-: This makes sense but I'm not sure how we could express that generically.

hadley: I was thinking that rather than using a vector of new variable names, you provide an integer vector that gives the source variable in the input for each variable in the output (using NA to indicate that it’s new)

lionel-: Right, but how could we express that through the vector interface of base or vctrs?

The only thing I can think of is to pass an ad hoc argument to `vec_cast()`. This may make sense, but since it's not generic and since that feature didn't previously exist, maybe we should wait for more use cases before introducing this approach.

hadley: Yeah, we'd need to figure that out in the long term; but in tidyr it seems like we could just make it work through `[.grouped.df`, or some new generic specifically for this case? Maybe it would eventually become a standard argument for `vec_restore()` for container classes?

hadley:
Does this fix #688?

R/extract.R
lionel-: I think I can make it work with only `[`, `vec_cbind()`, and `names<-`. But we need to add the restriction that splitted variables are a subtype of the original variable, which seems to make sense. This will also need fixes in other packages, including dplyr, because `names<-` is often buggy in instrumented data frame classes.

I will implement this new feature in a new PR.

lionel-:
It doesn't, rsample likely needs to implement ptype2 and cast, and possibly proxy and restore.

lionel-:
`df_cast()` will also be useful in tibble and perhaps in dplyr. We can copy it in packages where it is needed while we gain experience with it.

krlmlr:
Is tidyverse/tibble@dce86f2b still relevant?

lionel-:
yes

hadley:
What if we make `df_cast()` a generic? 

* If you can trivially transform a data frame back to your class, all you need to is provide `df_cast.myclass()`.

* If `df_cast()` works in most cases, then you can for the methods that need extra work you can coerce-down and then back up again:

    ```R
    separate.myclass <- function(.data, ...) {
      out <- separate(as.data.frame(.data))
      # restore out back to same thing as .data
    }
    ```

* If `df_cast()` doesn't work at all, then you'll have to provide a method for every generic. 

* If `df_cast()` doesn't work and your class is subclass of data frame, you might still provide `df_cast.myclass <- function(...) { stop("Not implemented") }`; the error would come after a bunch of work had been done, but you'd at least still get.

And let's call it `tidyr_df_cast()` so we can handle differently in dplyr until we figure out if there's some unifying principle.

We still also need to make `pivot_longer()`, `pivot_wider()`, etc generic.

lionel-:
Thinking more about this, `df_cast()` might not be the right name, because I would expect a cast function to return the same type as the target, or fail with an error. I'm not sure that `df_restore()` would be better because of potential differences with vctrs. So `dplyr_reconstruct()` and `tidyr_reconstruct()` might be safer names.

hadley:
I'll finish this off, using what we've learned from vctrs + dplyr since you worked on this.

hadley:
Closing for now; we'll come back to after dplyr 1.0.0.

PR #815, opened 2019-11-28
Rename data to .data in ... using functions

### LINKED ISSUE #759: Use .data in pack()
In `dplyr` functions, the first argument are mostly `.data` to prevent possible conflicts. But in ` tidyr`, most functions use `data` as the first argument. And it creates an issue that when setting `data` as column name.

For example:

Since `data` is already used as the first argument of `pack()` function, when using "data" as a column, it will throw an error message. 

```r
df <- tibble(x1 = 1:3, x2 = 4:6, x3 = 7:9, y = 1:3)
df %>% pack(data = c(x1, x2, x3), y = y) 
```
throws an error message:
```r
Error: All elements of `...` must be named
```

The same applies to at least `pack`, `unpack`, `chop` and `unchop`

hadley:
Fixes #759

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # tidyr (development version)
 
+* `pack()` and `unnest()` now use `.data` instead of data for the first 
+  argument (#759).
+
 * In `pivot_wider()`, `values_fn` and `values_fill` can now be single values; 
   you now only need to use a named list if you want to use different values
   for different value columns (#739, #746).
--- a/R/nest.R
+++ b/R/nest.R
@@ -208,7 +208,7 @@ check_key <- function(.key) {
 #'   use `names_sep` instead.
 #' @export
 #' @rdname nest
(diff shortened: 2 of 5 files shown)


hadley:
@lionel- can you see any way to do this that won't break existing method definitions?

lionel-:
I think adding `data` back for a period of deprecation should work. If they were generic it would probably be harder. Those signatures will look messy if only one of the arguments is prefixed with a dot. Perhaps it's not too late to change the other arguments as well, since they are relatively new?

One alternative is to keep the signatures as is and require using `:=` to pass arbitrary names through `...` without matching on formal parameters. That's not consistent with our style guide though.

hadley:
I think I'll split the difference and just fix `pack()` since `unnest()` is a generic.

hadley:
Oh and `...` is deprecated in `unnest()` so it doesn't need it.

PR #832, opened 2019-12-13
Add .drop parameter to hoist

### LINKED ISSUE #836: Add .drop parameter to hoist
Often, I don't want to extract every component of a list column but still drop it afterwards. In my opinion, this is actually the more common case, so I would even make .drop = TRUE the default.

mgirlich:
Fixes #836.
Should `.drop = TRUE` be the default? This would be a breaking change but it feels like the more common case to me.


hadley:
Let's dwell on this a little more (see discussion in issue).

PR #833, opened 2019-12-13
automatically name any unnamed pluckers

### LINKED ISSUE #837: automatically name unnamed pluckers in hoist
Often one wants to name the columns produced by `hoist()` the same as the element in the list. So to make `hoist()` less verbose and more comfortable for interactive use I propose automatically naming the column as follows for unnamed arguments:

* if the last part of the pluck specification is a character use that
* if the last part of the pluck specification is a number paste the number to the last character part of the spec, e.g. `list("film", 1)` would produce the column name `film_1`.

mgirlich:
Fixes #837

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -50,6 +50,9 @@
 
 * `hoist()` now uses `rlang::list2()` which means that you can now use `!!!` and 
   `:=` (#801).
+  
+* `hoist()` automatically generates a name for unnamed arguments in `...`
+  (@mgirlich).
 
 * In `extract()`, you can now use `NA` in `into` as documented (#793).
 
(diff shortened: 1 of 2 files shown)


batpigandme:
Is this tied to an issue? If yes, could you please link to it using [GitHub keywords](https://help.github.com/en/enterprise/2.18/user/managing-your-work-on-github/closing-issues-using-keywords#about-issue-references)?

mgirlich:
No, I haven't created an issue for this yet but only the PR. Should I also create an issue for this?

batpigandme:
Yeah, that'd be great! I just saw you opened #835, which is a perfect format (describing the goal, etc). Thanks so much.

hadley:
Closing since we're going to start with the simpler approach — thanks for the code and the discussions that continue to make hoist better!

PR #847, opened 2020-01-02
Fix tests with dev vctrs

krlmlr:
Broken since https://github.com/r-lib/vctrs/issues/660. Also added fixes so that tidyr now tests correctly with dev tibble.

CC @DavisVaughan (I can't request a review).

### DIFF
--- a/R/nest.R
+++ b/R/nest.R
@@ -172,7 +172,7 @@ nest.tbl_df <- function(.data, ..., .names_sep = NULL, .key = deprecated()) {
 
   keys <- .data[asis]
   u_keys <- vec_unique(keys)
-  out <- map(cols, ~ vec_split(set_names(.data[.x], names(.x)), keys)$val)
+  out <- map(cols, ~ vec_split_list_of(set_names(.data[.x], names(.x)), keys)$val)
 
   vec_cbind(u_keys, new_data_frame(out, n = nrow(u_keys)))
 }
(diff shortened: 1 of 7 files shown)


DavisVaughan:
I suppose it is worth discussing if continuing to return a `list_of` is useful?

I think `nest()` will have the same issue?

krlmlr:
`nest()` is fixed with a `vec_split_list_of()` implementation, which perhaps might live in vctrs?

krlmlr:
Are you suggesting to not return `list_of`? I rather like explicitly typed columns.

DavisVaughan:
It's probably worth using `new_list_of()` for performance since you know the input structure

``` r
df <- data.frame(x = 1, y = 2)
x <- rep_len(list(df), 1000)

library(vctrs)

bench::mark(
  as_list_of(x, .ptype = df),
  new_list_of(x, ptype = vec_ptype(df))
)
#> # A tibble: 2 x 6
#>   expression                                min  median `itr/sec` mem_alloc
#>   <bch:expr>                            <bch:t> <bch:t>     <dbl> <bch:byt>
#> 1 as_list_of(x, .ptype = df)             1.87ms  2.41ms      403.    80.8KB
#> 2 new_list_of(x, ptype = vec_ptype(df)) 18.95µs 23.14µs    42927.    17.9KB
#> # … with 1 more variable: `gc/sec` <dbl>
```

<sup>Created on 2020-01-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

hadley:
Sorry I missed this; now in v1.0.1 branch, which I'm about to release to CRAN.

PR #869, opened 2020-01-31
disambiguate context variable x from x column

### LINKED ISSUE #857: Refresh `drop_na()` docs
* [ ] Proofread title, description, and parameters 
* [ ] Proofread comments in examples 
* [ ] Ensure comment headings used to divide into scannable sections 
* [ ] Flag examples that seem overly complex for further review 
* [ ] Switch out mtcars/iris for something more interesting, or a smaller dataset constructed specifically to illustrate one point. 
* [ ] Review size of example outputs to ensure that you can easily see important rows and cols. 
* [ ] Datasets should always be tibbles to avoid use of `as_tibble()` distracting from the main point

shanepiesik:
Closes #857

### DIFF
--- a/R/drop-na.R
+++ b/R/drop-na.R
@@ -8,6 +8,8 @@
 #' df <- tibble(x = c(1, 2, NA), y = c("a", NA, "b"))
 #' df %>% drop_na()
 #' df %>% drop_na(x)
+#' z <- 2
+#' df %>% drop_na(!! z)
 #' @export
 drop_na <- function(data, ...) {
   ellipsis::check_dots_unnamed()
--- a/R/drop-na.R
+++ b/R/drop-na.R
@@ -8,8 +8,9 @@
 #' df <- tibble(x = c(1, 2, NA), y = c("a", NA, "b"))
 #' df %>% drop_na()
 #' df %>% drop_na(x)
+#' # using a contextual object with all_of()
 #' z <- 2
-#' df %>% drop_na(!! z)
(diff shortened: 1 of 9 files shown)


R/drop-na.R
```diff
@@ -8,6 +8,8 @@
 #' df <- tibble(x = c(1, 2, NA), y = c("a", NA, "b"))
 #' df %>% drop_na()
 #' df %>% drop_na(x)
+#' z <- 2
+#' df %>% drop_na(!! z)
```
hadley: I think we'd now prefer `df %>% drop_na(all_of(z))`.

R/gather.R
```diff
@@ -33,9 +33,10 @@
 #' expression, you can unquote them with the tidy eval operator
 #' `!!`. This operator evaluates its argument in the context and
 #' inlines the result in the surrounding function call. For instance,
-#' `c(x, !! x)` selects the `x` column within the data frame and the
-#' column referred to by the object `x` defined in the context (which
-#' can contain either a column name as string or a column position).
+#' `c(x, !! a)` selects the `x` column within the data frame and the
```
hadley: I think this section needs a full revamp to use `any_of()`/all_of()` instead of `!!`. Do you want to take a stab at that or leave it up to us?

shanepiesik: That makes sense, I'll take a stab at it.

hadley:
Thanks @shanepiesik!

PR #873, opened 2020-02-01
Create programming vignette with intro and tidy selection sections

### LINKED ISSUE #854: Programming with tidyr vignette
Drawing from <http://ggplot2.tidyverse.org/dev/articles/ggplot2-in-packages.html> and <http://dplyr.tidyverse.org/dev/articles/programming.html>, but adapting examples to use tidyr functions.  tidyr mostly uses tidy-selection rather than data-masking, so that should be given higher priority.

jenren:
Resolves #854 

Not completed yet, but opening a PR for the draft at the end of tidyverse dev days. This currently includes:
* "Introduction" (largely borrowed from http://dplyr.tidyverse.org/dev/articles/programming.html)
* "Tidy selection" (also largely borrowed from http://dplyr.tidyverse.org/dev/articles/programming.html)

Areas which still need work are:
* Fleshing out "Tidy selection" to include more examples
* Simplify "Tidy selection" code snippets once tidyr uses latest version of tidyselect
* "Data masking"
* "How tos"


hadley:
Thanks @jenren! This is a great start and very helpful. It'll be a little while before I get a chance to review because I'm going on vacation.

hadley:
Overall, looks great, but as you note in the contents, we'll need to wait until we've upgraded tidyr to use the latest tidyselect. So I'll take another pass when that happens.  

Thanks again for your contribution — this really is so helpful for me!

jennybc:
I pushed some changes to the intro. I'll make a few comments on the rest too.

vignettes/programming.Rmd
```diff
@@ -0,0 +1,113 @@
...
+    nest_egg <- function(df, cols) {
+      nest(df, egg = {{ cols }})
+    }
+    
+    nest_egg(mini_iris, -Species)
+    ```
+
+*   When you have a character vector of variable names, you must use `all_of()` 
+    or `any_of()` depending on whether you want the function to error if a 
+    variable is not found. These functions allow you to write for loops or a
```
jennybc, line 90: I don't really see the connection to `for` loops (here or in the intro). It feels like talking about the scenarios of "character vector of names" and "writing your own wrapper function" covers it.

vignettes/in-packages.Rmd
```diff
@@ -63,93 +57,58 @@ mini_iris %>% nest(
...
-sel_vars <- function(df, cols) {
-  tidyselect::vars_select(names(df), {{ cols }})
-}
-
-sel_vars(mini_iris, -Species)
-```
-
-(Many tidyr functions use `...` so you can easily select many variables, e.g. `fill(df, x, y, z)`. I now believe that the disadvantages of this approach outweigh the benefits, and that this interface would have been better as `fill(df, c(x, y, z))`. For new functions that select columns, please just use a single argument and not `...`.)
-
-If you decide to support tidyselect syntax, I recommend re-exporting the select helpers so that users can use (e.g.) `starts_with()` without any extra work. There's no helper to [do this yet](https://github.com/r-lib/usethis/issues/868), but you can copy-and-paste the code that [tidyr uses](https://github.com/tidyverse/tidyr/blob/00a66b44d2f304a73e953dd73139db61ed9cc289/R/tidyr.R#L29-L53).
```
jennybc, line 117: Is this re-exporting advice no longer necessary?

vignettes/in-packages.Rmd
```diff
@@ -63,93 +57,58 @@ mini_iris %>% nest(
...
+The [tidyselect](https://tidyselect.r-lib.org) package offers an entire family of select helpers. You are probably already familiar with them from using `dplyr::select()`.
 
-### Travis-CI
+### Continuous integration
 
-Hopefully you've already adopted continuous integration for your package, in which `R CMD check` (which includes your own tests) is run on a regular basis, e.g. every time you push changes to your package's source on GitHub or similar. The tidyverse team currently relies most heavily on Travis-CI for this, so that will be our example. `usethis::use_travis()` can help you get started.
+Hopefully you've already adopted continuous integration for your package, in which `R CMD check` (which includes your own tests) is run on a regular basis, e.g. every time you push changes to your package's source on GitHub or similar. The tidyverse team currently relies most heavily on GitHub actions, so that will be our example. `usethis::use_github_action()` can help you get started.
```
jennybc, line 77:
```suggestion
Hopefully you've already adopted continuous integration for your package, in which `R CMD check` (which includes your own tests) is run on a regular basis, e.g. every time you push changes to your package's source on GitHub or similar. The tidyverse team currently relies most heavily on GitHub Actions, so that will be our example. `usethis::use_github_action()` can help you get started.
```

vignettes/in-packages.Rmd
```diff
@@ -63,93 +57,58 @@ mini_iris %>% nest(
...
 
-We recommend adding a build to your matrix that targets the devel version of tidyr:
+We recommend adding a workflow that targets the devel version of tidyr:
```
jennybc, line 79:
```suggestion
We recommend adding a job that targets the devel version of tidyr:
```
We should either say "job" or stick with "build to your matrix", since workflow has a very specific meaning (= 1 Github Actions yaml file). 

Or, wait: are we actually recommending a totally separate workflow for this? Because now that I look at the yaml below, that's what it looks like.

vignettes/in-packages.Rmd
```diff
@@ -63,93 +57,58 @@ mini_iris %>% nest(
...
 
 * Always? If your package is tightly coupled to tidyr, consider leaving this 
   build in place all the time, so you know if changes in tidyr affect your 
-  package. Modifiers like `allow_failures` and `fast_finish` may be useful. 
-  Read more about the [Travis build matrix](https://docs.travis-ci.com/user/customizing-the-build/#build-matrix).
+  package. 
   
 * Right before a tidyr release? For everyone else, you could just add a 
-  tidyr-devel build during the period preceding a major tidyr release that has 
+  tidyr-devel workflow during the period preceding a major tidyr release that has
```
jennybc, line 86:
```suggestion
  tidyr-devel job to your build matrix during the period preceding a major tidyr release that has 
```

vignettes/in-packages.Rmd
hadley: tidyselect 1.0.0 automatically makes them available in the right scope.

vignettes/in-packages.Rmd
hadley: Yeah, that's what I was trying to imply. I think it's better to keep your `R-cmd-check.yml` as pure as possible to make it easier to update in the future.

vignettes/programming.Rmd
hadley: It's the other form of indirection we traditionally forgot to talk about in (e.g.) dplyr documentation. But I couldn't come up with an example where it would be helpful in tidyr.

hadley:
My sense is that the programming vignette is a bit of a placeholder at the moment — over time it'll grow to include more examples as they occur. So I'm going to merge it now in the understanding that it's not perfect, but I think it's better than what we had before (esp. in terms of updating the language to be consistent with dplyr etc)

PR #887, opened 2020-02-17
Don't create unspecified vectors when unchopping empty non-lists

lionel-:

Issue #751, opened 2019-09-18
nest() / unnest() in 1.0.0 significantly slower


PMSeitzer:
It appears the new implementation of `nest()` and `unnest()` has resulted in a dramatic slowdown, compared to the previous version of `tidyR`

Perhaps the problem is related to size preallocation?

In that case, `num_rows` needs to be large to observe the slowdown. See code snippet below.

```r
  num_rows <- 100000
  
  x <- dplyr::tibble(first = 1:num_rows,
                     second = 5:(num_rows+5-1),
                     third = 7:(num_rows+7-1))
  
  before <- Sys.time()
  
  y <- dplyr::tibble(first = 1:num_rows,
                     second = 5:(num_rows+5-1),
                     third = 7:(num_rows+7-1)) %>%
      tidyr::nest(second_and_third = c(second, third)) %>%
      tidyr::unnest(second_and_third)
  
  after <- Sys.time()
  
  if(length(which(x != y)) != 0){
    stop("nest() and unnest() procedure results in corrupted data!")
  }
  
  cat(paste("Execution Time:",difftime(after,before,units="secs"),"seconds"))
```
On my system:
```
Execution Time: 61.2449209690094 seconds
```

PMSeitzer:
I tested this with tidyR `0.8.3`:
```r
num_rows <- 100000

x <- dplyr::tibble(first = 1:num_rows,
                   second = 5:(num_rows+5-1),
                   third = 7:(num_rows+7-1))

before <- Sys.time()

y <- dplyr::tibble(first = 1:num_rows,
                   second = 5:(num_rows+5-1),
                   third = 7:(num_rows+7-1)) %>%
  tidyr::nest(second_and_third = c(second, third)) %>%
  tidyr::unnest()

after <- Sys.time()

if(length(which(x != y)) != 0){
  stop("nest() and unnest() procedure results in corrupted data!")
}

cat(paste("Execution Time:",difftime(after,before,units="secs"),"seconds"))
```
and there I saw
```
Execution Time: 5.78480935096741 seconds
```
Slow down of 10.587x

PMSeitzer:
It's worth mentioning that `nest_legacy()` and `unnest_legacy()` are still fast:

```r
  num_rows <- 100000
  
  x <- dplyr::tibble(first = 1:num_rows,
                     second = 5:(num_rows+5-1),
                     third = 7:(num_rows+7-1))
  
  before <- Sys.time()
  
[...]

PR #888, opened 2020-02-17
Improve `unchop()` performance

### LINKED ISSUE #886: `unchop()` fails on data frame columns
``` r
library(tidyr)

df <- tibble(x = 1:2, y = tibble(a = 1:2, b = 3:4, c = 5:6))
df
#> # A tibble: 2 x 2
#>       x   y$a    $b    $c
#>   <int> <int> <int> <int>
#> 1     1     1     3     5
#> 2     2     2     4     6

unchop(df, y)
#> Error in rep(vec_seq_along(data), n): invalid 'times' argument

# should be more like this
unpack(df, y)
#> # A tibble: 2 x 4
#>       x     a     b     c
#>   <int> <int> <int> <int>
#> 1     1     1     3     5
#> 2     2     2     4     6
```

This is because `[[` is being used internally to do the slicing, and this isn't correct for data frame columns. It should use `vec_slice2()` https://github.com/r-lib/vctrs/pull/626

DavisVaughan:
Part of #751 
Closes #886 
Closes #887 

This PR moves these two fairly expensive lines of `unchop()` to C. This greatly reduces the number of list allocations required, and nicely speeds things up.

```r
x <- pmap(as.list(data)[cols], vec_recycle_common)
x <- map(x, ~ new_data_frame(drop_null(.x)))
```

``` r
library(tidyr)
n <- 10000

df <- tibble(
  g = 1:n,
  y = rep(list(tibble(x = 1:5)), n)
)

# Master tidyr, CRAN vctrs
bench::mark(
  unnest(df, y),
  unnest_legacy(df, y),
  iterations = 50
)
#> # A tibble: 2 x 6
#>   expression                min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>           <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 unnest(df, y)         383.4ms  506.6ms      1.98    3.55MB     18.8
#> 2 unnest_legacy(df, y)   54.6ms   68.2ms     14.6     1.34MB     15.7

# Master tidyr, Master vctrs
bench::mark(
  unnest(df, y),
  unnest_legacy(df, y),
  iterations = 50
)
#> # A tibble: 2 x 6
#>   expression                min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>           <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 unnest(df, y)           213ms  305.5ms      3.26     3.5MB     13.6
#> 2 unnest_legacy(df, y)     52ms   55.2ms     17.1     1.34MB     13.0

# This branch tidyr, Master vctrs
bench::mark(
  unnest(df, y),
  unnest_legacy(df, y),
  iterations = 50
)
#> # A tibble: 2 x 6
#>   expression                min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>           <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 unnest(df, y)            88ms  137.7ms      6.65     3.4MB     15.2
#> 2 unnest_legacy(df, y)   51.8ms   74.3ms     13.8     1.34MB     17.1
```


lionel-:
How does this compare with R-level loops? Would be nice to go in the direction of less native code in tidyr, if possible.

DavisVaughan:
Benchmarking a rough implementation that doesn't have support for dropping `NULL`s seems to take about ~245ms. Current master is ~300ms. And this PR is ~135ms.

It's a double loop, so it's not great

For reference, I'm switching out the `unchop_rows()` line with:

```r
  n <- length(cols_lst)

  i_seq <- seq_len(size)
  j_seq <- seq_len(n)

  rows <- vector("list", size)
  sizes <- vector("integer", size)

  cols_names <- names(cols_lst)

  for (i in i_seq) {
    rows_elt <- vector("list", n)

    for (j in j_seq) {
      elt <- cols_lst[[j]][[i]]

      if (is.null(elt)) {
        # break and do null support
      }

      rows_elt[[j]] <- elt
    }

    sizes_elt <- vec_size_common(!!!rows_elt)
    rows_elt <- vec_recycle_common(!!!rows_elt, .size = sizes_elt)

    names(rows_elt) <- cols_names
    rows_elt <- new_data_frame(rows_elt, n = sizes_elt)

    rows[[i]] <- rows_elt
    sizes[[i]] <- sizes_elt
  }
```

hadley:
If R-level loops don't work, is there some other underlying pattern that could move to vctrs?

DavisVaughan:
It does look like there might be some underlying pattern of taking a data frame and "flattening" (?) it. That's what we do to `data[cols]`. I'm not sure what the right word for this is, and at first glance it doesn't seem like it would generalize to non-data-frame types.

But we could rewrite `unchop()` with this, and move the idea of `df_flatten()` to vctrs.

Rewritten `unchop()`:

```r
unchop <- function(data, cols, keep_empty = FALSE, ptype = NULL) {
  cols <- tidyselect::vars_select(tbl_vars(data), !!enquo(cols))
  if (length(cols) == 0) {
    return(data)
  }

  if (keep_empty) {
    for (col in cols) {
      data[[col]][] <- map(data[[col]], init_col)
    }
  }

  # In case `x` is a grouped data frame and any `cols` are lists,
  # in which case `[.grouped_df` will error
  cols <- new_data_frame(as.list(data)[cols])

  res <- df_flatten(cols, ptype)
  new_cols <- res$value
  slice_id <- res$id

  out <- vec_slice(data, slice_id)

  out <- update_cols(out, new_cols)
  reconstruct_tibble(data, out)
}

df_flatten <- function(x, ptype = NULL) {
  if (!is.data.frame(x)) {
    stop("`x` must be a data frame")
  }

  size <- vec_size(x)

  if (size == 0) {
    value <- map(x, ~ attr(.x, "ptype") %||% unspecified(0))
    value <- new_data_frame(value, n = 0L)
    out <- new_data_frame(list(id = integer(), value = value), n = 0L)
    return(out)
  }

  results <- unchop_rows(x, size)
  rows <- results[[1]]
  sizes <- results[[2]]

  id <- rep(seq_len(size), sizes)
  value <- vec_rbind(!!!rows, .ptype = ptype)

  out <- list(id = id, value = value)
  out <- new_data_frame(out)

  out
}
```

The `df_flatten()` idea:

``` r
library(tibble)
df_flatten <- tidyr:::df_flatten

df1 <- tibble(y = list(1:2, 3:4), z = list(c("a", "b"), c("c", "d")))

# `id` maps you back to the original row
# `value` is a data frame column
as_tibble(df_flatten(df1))
#> # A tibble: 4 x 2
#>      id value$y $z   
#>   <int>   <int> <chr>
#> 1     1       1 a    
#> 2     1       2 b    
#> 3     2       3 c    
#> 4     2       4 d

# atomic columns are repeated to match the list columns
df2 <- tibble(x = 1:2, y = list(1:2, 3:4))

as_tibble(df_flatten(df2))
#> # A tibble: 4 x 2
#>      id value$x    $y
#>   <int>   <int> <int>
#> 1     1       1     1
#> 2     1       1     2
#> 3     2       2     3
#> 4     2       2     4

# NULL-dropping behavior is the same as unchop()
df3 <- tibble(x = list(1:2, NULL))
df4 <- tibble(x = list(1:2, NULL), y = list(3, 4))

# NULL rows are dropped
as_tibble(df_flatten(df3))
#> # A tibble: 2 x 2
#>      id value$x
#>   <int>   <int>
#> 1     1       1
#> 2     1       2

# but if any other columns had data, we can't drop NULL rows
as_tibble(df_flatten(df4))
#> # A tibble: 3 x 2
#>      id value$x    $y
#>   <int>   <int> <dbl>
#> 1     1       1     3
#> 2     1       2     3
#> 3     2      NA     4
```

<sup>Created on 2020-02-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

DavisVaughan:
Since this is a rbind-ish operation, maybe converting input to a data frame with `as_df_row()` is the correct way to generalize this.

```r
vec_flat_bind <- function(x, ptype = NULL) {
  x <- vctrs:::as_df_row(x)

  size <- vec_size(x)

  if (size == 0) {
    value <- map(x, ~ attr(.x, "ptype") %||% unspecified(0))
    value <- new_data_frame(value, n = 0L)
    out <- new_data_frame(list(loc = integer(), value = value), n = 0L)
    return(out)
  }

  results <- unchop_rows(x, size)
  rows <- results[[1]]
  sizes <- results[[2]]

  loc <- rep(seq_len(size), sizes)
  value <- vec_rbind(!!!rows, .ptype = ptype)

  out <- list(loc = loc, value = value)
  out <- new_data_frame(out)

  out
}
```

``` r
library(tidyr)

vec_flat_bind_tbl <- function(x, ptype = NULL) {
  as_tibble(tidyr:::vec_flat_bind(x, ptype))
}

vec_flat_bind_tbl(1:2)
#> New names:
#> * `` -> ...1
#> * `` -> ...2
#> # A tibble: 1 x 2
#>     loc value$...1 $...2
#>   <int>      <int> <int>
#> 1     1          1     2

vec_flat_bind_tbl(c(x = 1, y = 2, z = 3))
#> # A tibble: 1 x 2
#>     loc value$x    $y    $z
#>   <int>   <dbl> <dbl> <dbl>
#> 1     1       1     2     3

vec_flat_bind_tbl(list(x = 1:5, y = 2, z = 3))
#> # A tibble: 5 x 2
#>     loc value$x    $y    $z
#>   <int>   <int> <dbl> <dbl>
#> 1     1       1     2     3
#> 2     1       2     2     3
#> 3     1       3     2     3
#> 4     1       4     2     3
#> 5     1       5     2     3
```

<sup>Created on 2020-02-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

DavisVaughan:
So, taking a step back, it turns out there is a much more efficient column wise approach to do this. 

- This avoids the `vec_rbind()` and `pmap()` entirely.
- It's entirely R based, but `vec_lengthen()` could probably move to vctrs and be rewritten in C with nice results
- It's nearly as fast as `unnest_legacy()`


```r
library(tidyr)
n <- 10000

df <- tibble(
  g = 1:n,
  y = rep(list(tibble(x = 1:5)), n)
)

bench::mark(
  unnest(df, y),
  unnest_legacy(df, y),
  iterations = 100
)
#> # A tibble: 2 x 6
#>   expression                min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>           <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 unnest(df, y)          76.7ms   97.4ms      10.0    3.38MB     22.4
#> 2 unnest_legacy(df, y)   50.9ms   61.4ms      16.3    1.34MB     23.0
```

DavisVaughan:
This now also incorporates @lionel-'s change and test in #887 since I've basically rewritten the core of `unchop()`

lionel-:
@DavisVaughan pointed out how the transpose implementation doesn't handle `NULL` elements correctly.

Detecting empty rows is tricky to fix without becoming super slow because we need and this requires a map/loop in the direction of the rows. This commit fixes it with a loop and we retain the good performance: https://github.com/tidyverse/tidyr/commit/4591264faeed46b324b85fcdcebc2d0b6345fa82

The code is definitely uglier though :(

R/chop.R
```diff
@@ -88,27 +88,147 @@ unchop <- function(data, cols, keep_empty = FALSE, ptype = NULL) {
...
+  out <- vec_slice(data, slice_loc)
 
   out <- update_cols(out, new_cols)
   reconstruct_tibble(data, out)
 }
 
-
 # Helpers -----------------------------------------------------------------
 
+vec_lengthen <- function(x, ptype = NULL) {
```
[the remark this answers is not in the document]
hadley: @DavisVaughan could you please add a few comments along these lines, and then ping me to another look.

DavisVaughan:
FYI, I'm going to take another look at this tomorrow

R/chop.R
```diff
@@ -88,27 +88,147 @@ unchop <- function(data, cols, keep_empty = FALSE, ptype = NULL) {
...
+
+finalise_size <- function(size) {
+  if (is.na(size)) {
+    0L
+  } else {
+    size
+  }
+}
+
+tidyr_recycle <- function(x, size) {
```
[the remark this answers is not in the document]
DavisVaughan: Maybe? I feel like for consistency if `size == 0` then it should still generate a `unspecified(0)` if we went this route

R/chop.R
```diff
@@ -96,27 +96,167 @@ unchop <- function(data, cols, keep_empty = FALSE, ptype = NULL) {
...
+      elt <- col[[j]]
+
+      old_size <- sizes[[j]]
+      new_size <- tidyr_size(elt)
+
+      sizes[[j]] <- tidyr_size2(old_size, new_size)
+    }
+  }
+
+  sizes <- map_int(sizes, tidyr_size_finalise)
```
DavisVaughan, line 153: It is possible to simplify this and remove `tidyr_size()`, `tidyr_size2()` and `tidyr_size_finalise()` by using

```r
  row <- vector("list", width)

  for (i in seq_len_size) {
    for (j in seq_len_width) {
      row[[j]] <- x[[j]][[i]]
    }

    sizes[[i]] <- vec_size_common(!!!row)
  }
```

but this ends up being a fair bit slower. Maybe if we reimplement in C this will be a faster approach.

DavisVaughan:
Ok @hadley I've finished

R/chop.R
hadley: This is effectively creating a `vec_size(x)` x `length(x)` matrix containing the sizes of each element, and then computing the common size for each row? (Just want to make sure I understand the algorithm here)

DavisVaughan: Yes, that is exactly right and is a nice way to explain it. I've added another code comment to clarify.

PR #889, opened 2020-02-18
Use `names<-` rather than `set_names()` in `unnest()`

DavisVaughan:
When `unnest()`ing a list column that doesn't have data frames in it, they are converted to `data_frames()` with `as_df()`. There is a call to `set_names()` that is run many times, which ends up being rather expensive. Using `names<-` instead seems to cut a decent chunk of time off.

The `vec_is()` call is also really expensive, but hopefully that will get better with a native implementation in vctrs. Mainly the `is_null()` checks of `vec_is()` take a while, but at the C level they should be faster

```r
library(tidyr)
n <- 10000

df <- tibble(
  g = 1:n,
  y = rep(list(1), n),
  z = rep(list(1), n),
  w = rep(list(1), n)
)

# master tidyr 
bench::mark(unnest(df, c(y, z, w)), iterations = 100)
#> # A tibble: 1 x 6
#>   expression                  min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>             <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 unnest(df, c(y, z, w))    830ms    936ms      1.06    3.83MB     14.1

# with `names<-`
bench::mark(unnest(df, c(y, z, w)), iterations = 100)
#> # A tibble: 1 x 6
#>   expression                  min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>             <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 unnest(df, c(y, z, w))    568ms    707ms      1.41    3.83MB     13.1
```


lionel-:
Can you look into making `set_names()` faster in rlang first, and only then change the implementations elsewhere if it isn't sufficient? This way all packages benefit from the speed up. It probably makes sense to rewrite it in C.

R/nest.R
```diff
@@ -350,7 +350,9 @@ as_df <- function(x, col) {
     x
   } else if (vec_is(x)) {
     # Preserves vec_size() invariant
-    new_data_frame(set_names(list(x), col))
```
hadley, line 353: Could also use `setNames()` here?

lionel-: Davis reimplemented `set_names()` in C in the last CRAN version so maybe this PR can be closed.

DavisVaughan: Even with the update to `set_names()`, `names<-` is still much faster. In this case, `set_names()` basically does a few checks and then eventually calls `names<-` anyways.

``` r
library(rlang)
library(stats)

x <- list(1)
col <- "stuff"

bench::mark(
  set_names(x, col),
  `names<-`(x, col),
  setNames(x, col),
  iterations = 100000
)
#> # A tibble: 3 x 6
#>   expression             min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>        <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 set_names(x, col)   1.18µs   1.42µs   650390.        0B     26.0
#> 2 `names<-`(x, col)    161ns    185ns  4828749.        0B      0  
#> 3 setNames(x, col)     524ns    631ns  1463210.        0B     29.3
```

The best option might be to just pass names through in `new_data_frame()` since you can do this now. I'll update to this approach

```r
library(vctrs)
library(stats)
library(rlang)

x <- list(1)
col <- "stuff"

bench::mark(
  new_data_frame(x, names = col),
  new_data_frame(set_names(x, col)),
  new_data_frame(setNames(x, col)),
  new_data_frame(`names<-`(x, col)),
  iterations = 200000
)
#> # A tibble: 4 x 6
#>   expression                             min median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                        <bch:tm> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 new_data_frame(x, names = col)       900ns 1.29µs   757483.    6.39KB    114. 
#> 2 new_data_frame(set_names(x, col))   2.12µs 2.43µs   382186.        0B     22.9
#> 3 new_data_frame(setNames(x, col))    1.42µs 1.67µs   567186.        0B     28.4
#> 4 new_data_frame(`names<-`(x, col))   1.04µs 1.21µs   763106.        0B     26.7
```

hadley:
Thanks!

PR #934, opened 2020-04-21
Add names_glue argument to pivot_wider()

### LINKED ISSUE #899: pivot_longer() pivot_wider() need examples w/ names_sep in the function reference
For `pivot_longer` function, there are no examples of the arguments `names_sep` and `names_pattern`. The descriptions of those arguments in the documentation seem to suggest that regex is necessary, but it's not clear. On stackoverflow, it seems that there are solutions but rarely, and they never explain why those arguments work in the way they do in the answers; the documentation should do that.

It seems too important to not have examples for these arguments when they are necessary for changing multiple columns.

### LINKED ISSUE #931: FR: Implement names_glue on pivot_wider
Currently, names produced by `pivot_wider` can partly be controlled via `names_sep`.
Further controls require additional renaming by `dplyr::rename`.
I suggest to add `names_glue` option which glues names in the manner of glue package.
In the example below, `.value` is a special key word like that of `names_pattern` of `pivot_longer`

``` r
data.frame(
  x = c('a' , 'b'),
  y = 1:2,
  foo = 1:2,
  bar = 1:2
) %>%
  pivot_wider(
    names_from = c(x, y),
    values_from = c(foo, bar),
    names_glue = '{x}{y}_{.value}'
  )
```

| a1_foo| a1_bar| b1_foo| b2_bar|
|------:|------:|------:|------:|
|      1|      2|      1|      2|

hadley:
Also includes some documentation for names_sep and names_prefix.

Fixes #899. Fixes #931.

@atusy what do you think? Have I missed anything obvious?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # tidyr (development version)
 
+* `pivot_wider()` gains a `names_glue` argument that allows you to construct
+  output column names with a glue specification.
+
 * `pivot_wider()` can now pivot data frame columns (#926)
 
 * `chop()` now creates list-columns of class `vctrs::list_of()`. This
--- a/R/pivot-wide.R
+++ b/R/pivot-wide.R
@@ -38,6 +38,9 @@
 #' @param names_prefix String added to the start of every variable name. This is
 #'   particularly useful if `names_from` is a numeric vector and you want to
 #'   create syntactic variable names.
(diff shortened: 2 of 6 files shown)


atusy:
@hadley Perfect!! I took a look at your code and tried the new feature. Everything seems fine. Thank you!!

hadley:
Thanks for trying it out for me — I appreciate it!

PR #944, opened 2020-04-24
Finish travis --> gh actions transition

jennybc:
The "In packages" vignette needed more work around the switch from travis to gh actions.


hadley:
Thanks!

PR #947, opened 2020-04-27
Give more control over column order in pivot_wider()

### LINKED ISSUE #839: FR: order of columns resulting from pivot_wider
In case I'm not just missing an obvious pre-existing solution, it would be nice to be able to specify how resulting columns from `pivot_wider` will be ordered. Using the `us_rent_income` example,
```
us_rent_income %>%
  pivot_wider(names_from = variable, values_from = c(estimate, moe))
```
produces the columns, `estimate_income`, `estimate_rent`, `moe_income`, and `moe_rent`. In my use-case (getting to a double-header table), I want them to be in the order, `estimate_income`, `moe_income`, `estimate_rent`, and `moe_rent`.

hadley:
Fixes #839

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # tidyr (development version)
 
+* `pivot_wider()` gains a `names_sort` argument which allows you to sort
+  column names in order. The default, `FALSE`, orders columms by their 
+  first appearance (#839).
+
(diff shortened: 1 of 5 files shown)


hadley:
Closing, since I've merged in the first commit into master, and I think there are enough variants to the second problem that it might be best handled at the spec level (i.e. by improving the documentation).

PR #956, opened 2020-05-06
Release 1.0.3

lionel-:
- Fix tests to deal with restricted `vec_cast()` semantics.

- Transform input to data frame in `separate_rows()`. Otherwise the list-column created in the implementation makes `grouped_df()` (called during restoration) fail when the list-col is a grouping variable.


hadley:
Thanks — I'll send to CRAN later this afternoon

PR #962, opened 2020-05-15
pivot_longer(names_to = NULL) to suppress names columns

### LINKED ISSUE #961: Allow pivot_longer(names_to = NULL) to produce no name columns
A suggestion. Sometimes the column names provide no new/useful information as the values encode the same (or better) information. In this case, the names column is unwanted and dropped soon after the pivot.

If `names_to` can create one or more columns based on patterns, could it be extended to allow 0 columns to simplify the `pivot_longer(names_to = "junk") %>% select(-junk)` pattern?

mitchelloharawild:
Resolves #961

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # tidyr (development version)
 
+* `pivot_longer()` now allows `names_to` to be length 0, which prevents name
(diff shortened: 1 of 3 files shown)


hadley:
Thanks!

PR #973, opened 2020-05-28
NA -> NULL in pivot_longer() help page

earowang:
I think you mean `NULL` instead of `NA`?

### DIFF
--- a/R/pivot-long.R
+++ b/R/pivot-long.R
@@ -25,7 +25,7 @@
 #'   or `names_pattern` is provided. In this case, there are two special
 #'   values you can take advantage of:
 #'
-#'   * `NA` will discard that component of the name.
+#'   * `NULL` will discard that component of the name.
 #'   * `.value` indicates that component of the name defines the name of the
 #'     column containing the cell values, overriding `values_to`.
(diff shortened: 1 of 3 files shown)


hadley:
Definitely meant `NA` here 😄 Do you have an example of the confusion?

earowang:
Oh I see.

Actually I don't have an example. But I don't find an example in the help page, related to `NA`. Now I see the difference between `NA` and `NULL`

* `NA` will discard that component of the name.
* `NULL` will discard the name.

``` r
library(tidyr)
anscombe %>%
  pivot_longer(everything(),
               names_to = c(NA, "set"),
               names_pattern = "(.)(.)"
  )
#> # A tibble: 88 x 2
#>    set   value
#>    <chr> <dbl>
#>  1 1     10   
#>  2 2     10   
#>  3 3     10   
#>  4 4      8   
#>  5 1      8.04
#>  6 2      9.14
#>  7 3      7.46
#>  8 4      6.58
#>  9 1      8   
#> 10 2      8   
#> # … with 78 more rows
```

<sup>Created on 2020-05-29 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

PR #974, opened 2020-05-28
Fix broken link for roxygen markdown formatting

batpigandme:
Fixing this in CONTRIBUTING across multiple packages (see r-lib/usethis#1126).


hadley:
Thanks!

PR #975, opened 2020-05-28
element in hoist example not named, throws error

gershomtripp:
The current example using hoist throws error `All elements of ... must be named`. Using `species = "species"` instead of just `"species"` fixes issue.


hadley:
If this really was a problem, it would be coming up in the automated tests.

gershomtripp:
Like the other issue I addressed [here](https://github.com/tidyverse/tidyr/issues/976), this issue is due to tidyr in tidyverse not being the most recent version.

PR #983, opened 2020-06-11
Use cpp11

jimhester:


jimhester:
So everything seems ok except for the error on 32 bit windows. I am seeing a similar issue with readr, so will try to track it down, I am not really sure right now what it could be.

jimhester:
This is now OK, the failure with R devel is due to what seems to be a bad link with the magrittr pipe operator (https://github.com/tidyverse/tidyr/runs/818275208#step:10:69)

hadley:
Fixes #917

PR #986, opened 2020-06-21
Remove stringi dependency #936

### LINKED ISSUE #936: Drop stringi dependency
We only use 6 functions, and it's a fairly heavy dependency.

rjpat:
Closes #936 

Have tried to completely sub out the stringi functions with the equivalent base ones and work without modifying any existing tests.


R/extract.R
```diff
@@ -55,7 +55,9 @@ str_extract <- function(x, into, regex, convert = FALSE) {
     is_character(into)
   )
 
-  matches <- stringi::stri_match_first_regex(x, regex)[, -1, drop = FALSE]
+  matches <- lapply(regmatches(x, regexec(regex, x)),
```
hadley, line 58: Could you please pull this out into a separate function? It's going to need unit tests for edge cases that `stri_match_first_regex()` might handle, but the base methods do not. I don't know exactly what those are, but I'd start with tests around no matches, NA matches, ...

R/separate-rows.R
```diff
@@ -31,7 +31,7 @@ separate_rows.data.frame <- function(data,
                                      convert = FALSE) {
   vars <- tidyselect::eval_select(expr(c(...)), data)
 
-  out <- purrr::modify_at(data, vars, stringi::stri_split_regex, pattern = sep)
+  out <- purrr::modify_at(data, vars, strsplit, split = sep)
```
hadley, line 34: I _think_ you should probably set `perl = TRUE` here (and in all other functions where the user supplies the match) since that'll make the regexp engine as similar as possible to stringi.

R/separate.R
```diff
@@ -131,8 +132,7 @@ str_split_fixed <- function(value, sep, n, extra = "warn", fill = "warn") {
   extra <- arg_match(extra, c("warn", "merge", "drop"))
   fill <- arg_match(fill, c("warn", "left", "right"))
 
-  n_max <- if (extra == "merge") n else -1L
-  pieces <- stringi::stri_split_regex(value, sep, n_max)
+  pieces <- lapply(value, function(x) strpieces(x, sep, n, extra))
```
hadley, line 135: I'd write this as 
```
lapply(value, strpieces, n = n, extra = extra)
```

R/separate.R
```diff
@@ -158,3 +158,15 @@ list_indices <- function(x, max = 20) {
...
+
+strpieces <- function(value, sep, n_max, extra) {
+  pieces <- strsplit(value, sep)[[1]]
+
+  if(extra != "merge" || all(pieces == value) || anyNA(pieces)) {
+    pieces
+  } else {
+    remainder <- substr(value, regexpr(pieces[n_max], value)[[1]][1], nchar(value))
+
+    c(pieces[1:n_max-1], remainder)
```
hadley, line 170:
```suggestion
    c(pieces[1:n_max - 1], remainder)
```

R/separate.R
```diff
@@ -158,3 +158,15 @@ list_indices <- function(x, max = 20) {
 
   paste(x, collapse = ", ")
 }
+
+strpieces <- function(value, sep, n_max, extra) {
+  pieces <- strsplit(value, sep)[[1]]
```
hadley, line 163: Is it possible to vectorise this function? Otherwise this might end up being a major performance hit.

R/extract.R
rjpat: Pulled it out and added a test suite. Based it off of stringi and adjusted to match. stringi had a lot of tests that worked around capture groups and blanks however tidyr won't need it as it checks earlier in extract to ensure the number of capture groups is consistent with the into value

R/separate.R
rjpat: Updated

R/separate.R
rjpat: Updated

R/extract.R
```diff
@@ -88,3 +88,13 @@ str_extract <- function(x, into, regex, convert = FALSE) {
 
   out
 }
+
+str_match_first <- function(x, regex) {
+  if (length(x) == 0) {
+    # Can't determine number of matches
```
hadley, line 94: I think this needs a little extra thought (I added the branch just as a placeholder:

``` r

stringi::stri_match_first(character(), regex = "(.)-(.)")
#>      [,1] [,2] [,3]
tidyr:::str_match_first(character(), regex = "(.)-(.)")
#> NULL
```

R/separate.R
```diff
@@ -158,3 +158,18 @@ list_indices <- function(x, max = 20) {
 
   paste(x, collapse = ", ")
 }
+
+str_pieces <- function(value, sep, n_max, extra) {
+  pieces <- sapply(value, strsplit, split = sep, perl = TRUE)[[1]]
```
hadley, line 163: Something seems weird here. It seems like this will only work when `value` is length 1?

rjpat: I thought because we are calling via `lapply()` it would only need to handle one string at a time. The only time I could see otherwise is when there is a nested column being passed but I don't think it was handled before.

Running the below on both the current release gave me the same output (reprex from release version, using a data example from nest docs);
``` r
library(dplyr, quietly = TRUE, warn.conflicts = FALSE)
library(tidyr, quietly = TRUE)

df <- tibble(x = c(1, 1, 2, 2), y = c("a.b", "c.d", "d.e", "f.g"))
df %>% nest(data = c(y)) %>%
  separate(data, c("A", "B"), extra = "drop")
#> # A tibble: 2 x 3
#>       x A     B    
#>   <dbl> <chr> <chr>
#> 1     1 list  y    
#> 2     2 list  y
```

hadley: I think the vectorisation should happen inside of `str_pieces()` not outside of it.

rjpat: Have rewritten to accept vectors and be more straightforward

R/extract.R
hadley: Hmmmm, also problematic when there are no matches:

``` r
str(stringi::stri_match_first("", regex = "(.)-(.)"))
#>  chr [1, 1:3] NA NA NA
str(tidyr:::str_match_first("", regex = "(.)-(.)"))
#>  chr[1, 0 ]
```

hadley: Just remembered I have some code lying around for exactly this problem 😀

hadley:
Thanks @rjpat!

PR #1002, opened 2020-07-16
Some easy performance improvements for `hoist()`

### LINKED ISSUE #1001: Some easy performance improvements for `hoist()`
Lots of time is wasted by calling `::` in `purrr::map()` which can easily be avoided.

hoist_profvis

mgirlich:
closes #1001.

### DIFF
--- a/R/rectangle.R
+++ b/R/rectangle.R
@@ -392,7 +392,7 @@ simplify_col <- function(x, nm, ptype = list(), transform = list(), simplify = F
     }
   }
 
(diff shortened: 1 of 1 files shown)


hadley:
Thanks! Given the dramatic performance improvement, I think this is worth it, even though the general problem will be resolved in R 4.1.

mgirlich:
Great to hear and also very much looking forward to the fix of the issue in R 4.1!

PR #1009, opened 2020-07-23
Update docs about tidy selection

yutannihilation:
- Replace `-` with `!`
- Wrap predicate functions with `where()`
- Include Rmds from tidyselect in the same way as dplyr
- Import `all_of()` and `any_of()`, and remove `one_of()`


R/tidyr.R
```diff
@@ -53,3 +53,9 @@ tidyselect::starts_with
 #' @importFrom tidyselect last_col
 #' @export
 tidyselect::last_col
+#' @importFrom tidyselect any_of
```
hadley, line 56: I think these can all actually be removed now, since tidyselect automatically provides an environment where they're available.

man/rmd/setup.Rmd
```diff
@@ -0,0 +1,17 @@
+
+```{r, include = FALSE}
```
hadley, line 2: Is this file used anywhere?

R/tidyr.R
yutannihilation: Sure.

Btw, dplyr does reexport `any_of()` and `all_of()`. Do you also want to remove them there? Or, is it too late?

man/rmd/setup.Rmd
yutannihilation: Seems nowhere, thanks for catching.

R/tidyr.R
hadley: I think you can remove all the tidyselect helpers now.

hadley: Well, I guess people might have used `tidyr::starts_with()`, so we probably need to save that for a bigger release. That's more likely to be the case with dplyr, but I bet there's at least one CRAN package that has done it with tidyr too.

yutannihilation: Do I need to add a NEWS item for stopping re-exporting them?

hadley: Sorry, I meant `starts_with()` as an example; we should probably keep all the existing exported tidyselect functions so we can get this PR into the existing patch release. And then remove them for the next minor release.

yutannihilation: Ah, got it! Sorry.

yutannihilation:
Are there still anything I need to fix? (No intent to hurry you. I just wonder if I'm missing something)

hadley:
No, sorry, this just completely slipped my mind and hence didn't make it into the tidyr patch release 😓

hadley:
Thanks @yutannihilation!

yutannihilation:
Never mind, I didn't expect this pull request can get into the patch release when writing this, so waiting for another release is fine to me. Thanks for your review!

PR #1022, opened 2020-08-06
Update chop.R

fartist:
fix a typo


R/chop.R
```diff
@@ -14,7 +14,7 @@
 #' @details
 #' Generally, unchopping is more useful than chopping because it simplifies
 #' a complex data structure, and [nest()]ing is usually more appropriate
-#' that `chop()`ing` since it better preserves the connections between
+#' than `chop()`ing` since it better preserves the connections between
```
erictleung, line 17: It'll also be good to just remove this stray backtick

```suggestion
#' than `chop()`ing since it better preserves the connections between
```

PR #1028, opened 2020-08-13
Update README.Rmd

kelseygonzalez:
While the readme is mostly up-to-date, the translation of pivoting terminology in different places has the retired gather/spread at the forefront instead of the recommended pivot_longer and pivot_wider. This is also out of date on https://tidyr.tidyverse.org/index.html, though I believe that site builds off of this readme.


hadley:
Great idea, thank you!

PR #1043, opened 2020-09-09
expand_grid typo

### LINKED ISSUE #1025: Typo in expand_grid docs
In the Value section, it should say "Varies the first element slowest" (currently says "fastest"). I also suggest changing "Compared to expand.grid" to "Compared to expand.grid, expand_grid will:"

nzare:
closes #1025 

Updated Value section in expand_grid docs-

- "Varies the first element slowest" (from "fastest")

- "Compared to expand.grid" to "Compared to expand.grid , expand_grid :"

### DIFF
--- a/R/expand.R
+++ b/R/expand.R
@@ -130,8 +130,8 @@ nesting <- function(..., .name_repair = "check_unique") {
 
 #' Create a tibble from all combinations of inputs
 #'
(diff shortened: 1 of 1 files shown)


hadley:
Thanks! I took a slightly different approach, but I really appreciate you caring about this problem enough to provide a fix 😄

Issue #941, opened 2020-04-23
Drop purrr dependency


hadley:
Since it's only used in 3 places.  Probably need to keep it as suggested since it's so important for `hoist()`

rjpat:
Been working through this but not sure if I have misunderstood. Been working through the imported `{purrr}` functions to transition them over to base so that `{purrr}` could be explicitly called in `hoist()` and it just be suggested.

This is the [commit](https://github.com/rjpat/tidyr/commit/6b32129b395d47e1f656ee2f68f1d1f012f17e99) with the current work, I guess just a gut check if this is the right way or if it should only be where `{purrr}` is explicitly referenced.

hadley:
Oh I guess I missed that some purrr functions were imported. What you've done so far looks reasonable to me.

hadley:
Given that purrr is now quite lightweight, and it's interwined throughout tidyr, I no longer think this is worth it.

PR #1046, opened 2020-09-19
purrr removal #941

rjpat:
To address issue #941 

Think my replacement for [`purrr::transpose()`](https://github.com/rjpat/tidyr/blob/86bec61f3b9a7605d0361acbb6e0486093ea6f88/R/utils.R#L117) and [this line](https://github.com/rjpat/tidyr/blob/86bec61f3b9a7605d0361acbb6e0486093ea6f88/R/extract.R#L71) should be reviewed as not particularly happy with them but don't have the mental model of what would be better.

This has left the following with `{purrr}` functions;

1. `hoist()`
2. `unnest_*`
3. `separate_rows()` as it has `purrr::modify_at()` in it


hadley:
I'm now wondering whether or not this is worth it — it's a lot more work than I had anticipated, and purrr is not a very heavy package.

rjpat:
I agree it did touch a lot of functions and may not be worth going from tidyverse to base R for a small performance gain

hadley:
Btw why do you think there will be a performance gain?

rjpat:
Did a very quick microbenchmark of pivot_wider and there was a fractional speed increase. Probably not enough of a sample set for me to have mentioned it

hadley:
Ah ok, cool, just wanted to make sure there wasn't an obvious performance issue in purrr.

Issue #1047, opened 2020-09-23
Unnest() pgkdown page version 0.8.3 available via google


Megaira:
Hello and thanks for the great package!

When `nest` and  `unnest` where moved into one file in https://github.com/tidyverse/tidyr/commit/03bb1985f364938953eced375504096d3e222d98 the old unnest documentation remained available and can be found via google: https://tidyr.tidyverse.org/reference/unnest.html (vs. the up-to-date one: https://tidyr.tidyverse.org/reference/nest.html)

jennybc:
Looks like a duplicate of #1042

Issue #1042, opened 2020-09-09
tidyr.tidyverse.org documentation is behind in version


mmuurr:
[tidyr.tidyverse.org](tidyr.tidyverse.org) docs still reflect v0.8.3.
For some objects this is fine, but for others (e.g. [unnest](https://tidyr.tidyverse.org/reference/unnest.html)) there are fundamentally-different patterns used in v1+ that aren't reflected in the online docs.

hadley:
Looks fine to me:

<img width="328" alt="Screenshot 2020-09-09 at 7 11 57 AM" src="https://user-images.githubusercontent.com/4196/92596647-c04e4980-f26b-11ea-85d3-bd238241cfa3.png">

mmuurr:
Ahh ... when navigate [tidyr.tidyverse.org](https://tidyr.tidyverse.org) manually and click through the links to a reference page (e.g. "Reference" in nav bar, then the "Nest and unnest" entry) I get the current version. But when I execute a [Google search](https://www.google.com/search?q=tidyr+unnest) (e.g. "[tidyr unnest](https://www.google.com/search?q=tidyr+unnest)") and click on Google's first result, I get the older version (v0.8.3), which appears to be a page I can't manually nav to anymore ... is the website a 'union' of updates? And perhaps an outdated page is still be crawled/indexed by Google?

Here's the exact URL: [https://tidyr.tidyverse.org/reference/unnest.html](https://tidyr.tidyverse.org/reference/unnest.html)

hadley:
Ah ok; we're tracking that underlying problem here: https://github.com/r-lib/pkgdown/issues/1394

PR #1048, opened 2020-09-25
Delete unnest.html as it is a 2 year old version

assignUser:
#1042 #1047


hadley:
It never occurred to me to do this directly 😆

PR #1050, opened 2020-09-28
Update pivot-wide.R

### LINKED ISSUE #1049: Documentation on 'values_fn' argument of 'pivot_wider' seems to be wrong
In the documentation of `pivot_wider`, the `values_fn` argument is described as follows.

> Optionally, a function applied to the `value` in each cell in the output. You will typically use this when the combination of `id_cols` and `value` column does not uniquely identify an observation.

I think it should be

> Optionally, a function applied to the `value` in each cell in the output. You will typically use this when the combination of `id_cols` and `name` column does not uniquely identify an observation.

junghoon-shin:
Closes #1049 

Corrected wrong description on the argument `values_fn` in the `pivot_wider` documentation.

### DIFF
--- a/R/pivot-wide.R
+++ b/R/pivot-wide.R
@@ -44,12 +44,13 @@
 #'
 #'   This can be a named list if you want to apply different fill values to
 #'   different value columns.
-#' @param values_fn Optionally, a function applied to the `value` in each cell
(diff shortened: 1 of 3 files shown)


DavisVaughan:
Thanks!

Issue #982, opened 2020-06-08
fill() should handle NaN


ArtemSokolov:
Dear tidyverse developers,

In the current implementation, `fill()` ignores `NaN`. My current workaround has been a (rather strange-looking) use of `replace_na()` to replace `NaN` with `NA` before passing the values to `fill()`:

```r
X <- tibble::tibble(x = c(1,0,3,0),
                    y = c(2,0,2,0),
                    z = x/y)

tidyr::fill(X, z)
#  # A tibble: 4 x 3
#        x     y     z
#    <dbl> <dbl> <dbl>
#  1     1     2   0.5
#  2     0     0 NaN
#  3     3     2   1.5
#  4     0     0 NaN

tidyr::replace_na(X, list(z=NA)) %>%
    tidyr::fill(z)
# # A tibble: 4 x 3
 #       x     y     z
 #   <dbl> <dbl> <dbl>
 # 1     1     2   0.5
 # 2     0     0   0.5
 # 3     3     2   1.5
 # 4     0     0   1.5
```

I believe it would be warranted to have `fill()` treat `NaN` and `NA` equivalently, because 1) it is already done so by `replace_na()`, and 2) `is.na(NaN)` is `TRUE`.

PR #1056, opened 2020-10-25
fill also fills NaN

ben-schwen:
Fixes issue #982 

`fill` should also handle `NaN` values like `replace_na` already does.

Current version:
```
df <- tibble(x = c(1, 2), y = c(1, NaN))
fill(df, y)
# A tibble: 2 x 2
      x     y
  <dbl> <dbl>
1     1     1
2     2   NaN
```

Updated in this PR:
```
df <- tibble(x = c(1, 2), y = c(1, NaN))
fill(df, y)
# A tibble: 2 x 2
      x     y
  <dbl> <dbl>
1     1     1
2     2     1
```

### DIFF
--- a/src/fill.cpp
+++ b/src/fill.cpp
@@ -40,7 +40,7 @@ SEXP fillDown(SEXP x) {
     double lastVal = xin[0];
 
     for (int i = 0; i < n; ++i) {
-      if (!ISNA(xin[i]))
+      if (!ISNAN(xin[i]))
         lastVal = xin[i];
       xout[i] = lastVal;
     }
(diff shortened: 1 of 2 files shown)


hadley:
We'll fix this in a different way, by switching to `vctrs::vec_fill_missing()`, which already supports filling `NaN`:

``` r
 vctrs::vec_fill_missing(c(1, NaN))
#> [1] 1 1
```

<sup>Created on 2021-02-18 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

hadley:
But thanks for the PR — I really appreciate that you care enough about this issue to work on a fix 😄

PR #1065, opened 2020-11-24
Examples for separate(): improve column names

kcarnold:
The examples used the same letters (a, b, and c) for both the data and the column names. This changes the examples to use x, y, and z instead.

(I have not tested this change.)


hadley:
Thanks!

PR #1070, opened 2020-12-07
uncount() small improvements

### LINKED ISSUE #1069: uncount() should error for non-integer weights
`uncount()` currently works with non-integer weights where it probably should not. Also the error message for negative weights is quite unclear.

``` r
library(tidyr)

# no error
tibble(x = c("a", "b"), n = c(1, 0.5)) %>% 
  uncount(n)
#> # A tibble: 1 x 1
#>   x    
#>   <chr>
#> 1 a

# unclear error message
tibble(x = c("a", "b"), n = c(1, -1)) %>% 
  uncount(n)
#> Error in rep(seq_nrow(data), w): invalid 'times' argument
```

<sup>Created on 2020-12-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

mgirlich:
Fixes #1069.

* clearer error message for negative weights
* error for non-integer weights
* use `vctrs` methods
* fix documentation of `.remove` parameter
* ~doesn't drop grouping variable (same behaviour as in `dplyr`)~ grouping variables are still dropped as there is no best answer what the behaviour should be


R/uncount.R
```diff
@@ -23,15 +24,18 @@
...
+  if (any(w < 0)) {
+    abort("all elements of `weights` must be >= 0")
   }
 
-  out <- data[rep(seq_nrow(data), w), , drop = FALSE]
+  out <- vec_slice(data, vec_rep_each(vec_seq_along(data), w))
 
-  if (.remove && quo_is_symbol(weights_quo)) {
-    out[[as_string(get_expr(weights_quo))]] <- NULL
+  weights_is_col <- quo_is_symbol(weights_quo) &&
```
hadley, line 35: I don't really like the idea of making it easier to use variables outside of the data frame. I get that the current error message isn't very informative (and making it better would be a reasonable amount of work), but I'd prefer not to go in this direction.

R/uncount.R
```diff
@@ -23,15 +24,18 @@
...
+    quo_name(weights_quo) %in% colnames(data)
+  if (.remove && weights_is_col) {
+    out <- dplyr::select(out, -!!as_name(weights_quo))
```
hadley, line 38: Reasoning through what `as_name(quo)` is tricky — can you please keep the previous stringification? And then do `-all_of(weights_chr)` or similar?

hadley (commented):
@DavisVaughan can you please take a look at the vctrs code?

R/uncount.R
mgirlich: Good point. I just realised that it is much easier to use `any_of()` as I don't need `quo_name(weights_quo) %in% colnames(data)` then.

R/uncount.R
mgirlich: I am not sure I really understand. The behaviour regarding variables outside of the data frame is the same as before. I thought I need the flag `weights_is_col` for `dplyr::select()` but you pointed me in the direction of `any_of()` which is way simpler anyways.

Example of current behaviour
``` r
library(tidyr)

df <- tibble(x = c("a", "b"), n = c(1, 2))
m <- 2
uncount(df, m)
#> # A tibble: 4 x 2
#>   x         n
#>   <chr> <dbl>
#> 1 a         1
#> 2 a         1
#> 3 b         2
#> 4 b         2
```

<sup>Created on 2020-12-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

hadley: Yeah, sorry, that was a red herring.

tests/testthat/test-uncount.R
```diff
@@ -22,7 +22,33 @@ test_that("expands constants and expressions", {
...
-  expect_error(uncount(df, "W"), "must evaluate to a numeric vector")
+  expect_error(uncount(df, "W"), class = "vctrs_error_incompatible_type")
+})
+
+test_that("works with 0 weights", {
+  df <- tibble(x = 1:2, w = c(0, 1))
+  expect_equal(uncount(df, w), tibble(x = 2))
+})
+
+test_that("errors on negative weights", {
```
DavisVaughan, line 51: `vec_rep_each()` will already error on negative `times`. We should probably consider adding `x_arg` and `times_arg` arguments to `vec_rep_each()` so the message can be improved and used here. What you have now is probably fine in the meantime.

```r
> vec_rep_each(df, df$w)
Error: `times` must be a vector of positive numbers. Location 1 is negative.
```

Same with `NA` times

```r
> vec_rep_each(df, NA)
Error: `times` can't be missing. Location 1 is missing.
```

DavisVaughan (changes requested):

R/uncount.R
```diff
@@ -23,15 +24,16 @@
...
+  w <- vec_cast(w, integer(), x_arg = "weights")
 
-  if (!is.numeric(w)) {
-    stop("`weights` must evaluate to a numeric vector", call. = FALSE)
+  if (any(w < 0)) {
+    abort("all elements of `weights` must be >= 0")
   }
 
-  out <- data[rep(seq_nrow(data), w), , drop = FALSE]
+  out <- vec_slice(data, vec_rep_each(vec_seq_along(data), w))
```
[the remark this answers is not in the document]
mgirlich: true, very nice

tests/testthat/test-uncount.R
mgirlich: Yes, that would be nice. I added a reminder for that in the code.

tests/testthat/test-uncount.R
```diff
@@ -22,7 +22,33 @@ test_that("expands constants and expressions", {
...
+  df <- tibble(g = 1, x = 1, w = 1) %>% dplyr::group_by(g)
+  expect_equal(uncount(df, w), df %>% dplyr::select(-w))
+})
+
+test_that("doesn't remove grouping variable", {
+  df <- tibble(g = 1, x = 1, w = 1) %>% dplyr::group_by(g)
+
+  expect_message(uncount(df, g)) %>%
+    expect_equal(df)
+})
```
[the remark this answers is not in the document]
mgirlich: I added it to the documentation of `.remove`. Not sure whether you'd prefer it there or in the details section.

I am not totally sure whether this is the correct behaviour but to me it feels more consistent with `dplyr::select()`. The feeling of surprise originally came from something like in the following example:

``` r
library(tidyr)

df <- tibble(g = c("a", "a", "b"), x = 1, w = c(1, 2, 1)) %>% 
  dplyr::group_by(g, w)

uncount(df, w)
#> # A tibble: 4 x 2
#> # Groups:   g [2]
#>   g         x
#>   <chr> <dbl>
#> 1 a         1
#> 2 a         1
#> 3 a         1
#> 4 b         1
uncount(df, w, .remove = FALSE)
#> # A tibble: 4 x 3
#> # Groups:   g, w [3]
#>   g         x     w
#>   <chr> <dbl> <dbl>
#> 1 a         1     1
#> 2 a         1     2
#> 3 a         1     2
#> 4 b         1     1
uncount(df, w + 0)
#> # A tibble: 4 x 3
#> # Groups:   g, w [3]
#>   g         x     w
#>   <chr> <dbl> <dbl>
#> 1 a         1     1
#> 2 a         1     2
#> 3 a         1     2
#> 4 b         1     1
```

<sup>Created on 2020-12-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

@hadley what's your opinion on that?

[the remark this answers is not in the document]
hadley: I think uncounting by a grouping variable is going to be relatively uncommon, and there's no clear best answer (since the two most applicable rules clash), so I think that argues for keeping the behaviour as is.

hadley:
Thanks!

PR #1072, opened 2020-12-07
Fix broken `\link{}` in doc

salim-b:
Before, the affected line in the documentation read

> like `year = 2010:2020` or `year = \link{full_seq}(year,1)`.

(`\link{}` inside a Markdown code span doesn't work for obvious reasons)


hadley:
Sorry I didn't spot this before I did it myself!

PR #1077, opened 2020-12-09
Capitalize Billboard as proper noun

erictleung:
Live site shows hanging comma and uncapitalized noun
https://tidyr.tidyverse.org/reference/billboard.html


hadley:
Thanks!

Issue #980, opened 2020-06-03
names_ptypes produces rlang errors under R 4.0


JaneWall:
Code that previously worked now produces an error for the names_ptypes 👍 
Error: Can't convert <character> to <integer>.
Run `rlang::last_error()` to see where the error occurred.

---
If I comment out the names_ptypes line and add a mutate(week = as.integer(week)), it works fine.

```r
billboard %>% 
  pivot_longer(
    cols = starts_with("wk"), 
    names_to = "week", 
    names_prefix = "wk",
    names_ptypes = list(week = integer()),
    values_to = "rank",
    values_drop_na = TRUE,
  )
```

batpigandme:
Reprex with backtrace (since the specific error seems to be that it can't convert character to integer):
``` r
library(tidyverse)

billboard %>% 
  pivot_longer(
    cols = starts_with("wk"), 
    names_to = "week", 
    names_prefix = "wk",
    names_ptypes = list(week = integer()),
    values_to = "rank",
    values_drop_na = TRUE,
  )
#> Error: Can't convert <character> to <integer>.
#> Backtrace:
#>      █
#>   1. ├─`%>%`(...)
#>   2. │ ├─base::withVisible(eval(quote(`_fseq`(`_lhs`)), env, env))
#>   3. │ └─base::eval(quote(`_fseq`(`_lhs`)), env, env)
#>   4. │   └─base::eval(quote(`_fseq`(`_lhs`)), env, env)
#>   5. │     └─`_fseq`(`_lhs`)
#>   6. │       └─magrittr::freduce(value, `_function_list`)
#>   7. │         ├─base::withVisible(function_list[[k]](value))
#>   8. │         └─function_list[[k]](value)
#>   9. │           ├─tidyr::pivot_longer(...)
#>  10. │           └─tidyr:::pivot_longer.data.frame(...)
#>  11. │             └─tidyr::build_longer_spec(...)
#>  12. │               └─vctrs::vec_cast(names[[col]], names_ptypes[[col]])
#>  13. └─vctrs::vec_default_cast(...)
#>  14.   └─vctrs::stop_incompatible_cast(...)
#>  15.     └─vctrs::stop_incompatible_type(...)
#>  16.       └─vctrs:::stop_incompatible(...)
#>  17.         └─vctrs:::stop_vctrs(...)
```

<sup>Created on 2020-06-03 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

[...]

PR #1088, opened 2021-01-20
Clarify names_ptypes vs. names_transform, 

mine-cetinkaya-rundel:
This PR addresses #980, though I'm not sure if it closes it.

It clarifies that `names_ptypes` is used to _confirm_ the type of a variable vs. `names_transform` is used to _change_ it. I think for most situations `names_transform` is what the user wants, not `names_ptypes`, so the paragraph about "if you don't do anything, these are the default types you'll get" is moved under `names_transform`.


hadley:
Thanks!

PR #1095, opened 2021-02-18
Allow pivot_longer's values_transform argument to take a single function

### LINKED ISSUE #1035: values_transform should accept a single function
I would expect that providing a single function to `values_transform` in `pivot_longer()` would apply to each column selected for pivoting (x and y in my example). This would be consistent to the recent change in `pivot_wider()`, from the News: 

> arguments values_fn and values_fill can now be single values; you now only need
to use a named list if you want to use different values for different value columns (#739, #746). 
They also get improved errors if they're not of the expected type.

```r
library(tidyr)
packageVersion("tidyr")
#> [1] '1.1.0'

df <- data.frame(x = 1:5, y = letters[1:5], stringsAsFactors = FALSE)

pivot_longer(df, c(x, y), values_transform = as.character)
#> Error: Can't combine `x` <integer> and `y` <character>.
```
If I do `pivot_longer(df, c(x, y), values_transform = list(value = as.character))`, it works as expected. 

If you agree this is a desired change, I would be happy to submit a PR, and I will also create a similar issue for `values_ptypes` (where I always face the same confusion).

smingerson:
Closes #1035.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,9 +12,9 @@
 
 * `separate_rows()` works for factor columns (@mgirlich, #1058).
 
-* `pivot_longer()` argument `values_transform` can now be a single 
-  function; you now only need to use a named list if you want to use different 
-  functions for different value columns (@smingerson, #1035). 
+* `pivot_longer()` arguments `names_transform`,`values_transform` can now be a
+  single function; you now only need to use a named list if you want to use
+  different functions for different value columns (@smingerson, #1035).
 
 
 # tidyr 1.1.2
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,11 @@
 
 * `separate_rows()` works for factor columns (@mgirlich, #1058).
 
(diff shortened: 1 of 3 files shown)


hadley (commented):
Looks good! Do the docs need an update too?

smingerson:
Yes, they do. Here I was, so impressed with myself for remembering a news item! The `values_transform` description is shared with `names_transform`. Should I add similar functionality to `names_transform` and update the docs, or split the documentation? I think it makes sense for both to allow functions, just seems a lot less useful for names.

hadley:
Yeah, I think it would be easier to be consistent.

smingerson:
Should be ready for another look when you have time.

hadley:
@DavisVaughan can you finish off the review please?

DavisVaughan:
Closed in favor of broader changes added in https://github.com/tidyverse/tidyr/pull/1287, I mentioned @smingerson in the NEWS bullet

PR #1101, opened 2021-03-03
More generic functions

### LINKED ISSUE #1071: more functions should be generic
while bringing some functions from `tidyr` to `dbplyr` I noticed that a couple of `tidyr` functions should probably be generic:

* [ ] `uncount()`

the pivot helpers would make sense as the `dbplyr` implemention of `pivot_*()` uses them anyway
* [ ] `build_longer_spec()`/`build_wider_spec()`
* [ ] `pivot_longer_spec()`/`pivot_wider_spec()`

data frame column helpers would make sense as nested tables in SQL are becoming more common (e.g. see https://github.com/tidyverse/dbplyr/issues/515)
* [ ] `pack()`/`unpack()`
* [ ] `unnest_longer()`/`unnest_wider()`

list column helpers: not sure if they are needed but for completeness I also listed them here:
* [ ] `chop()`/`unchop()`
* [ ] `hoist()`

mgirlich:
Fixes #1071 .


R/chop.R
```diff
@@ -89,7 +96,13 @@ chop <- function(data, cols) {
 
 #' @export
 #' @rdname chop
-unchop <- function(data, cols, keep_empty = FALSE, ptype = NULL) {
+unchop <- function(data, cols, keep_empty = FALSE, ptype = NULL, ...) {
```
hadley, line 99: In general, I think it's better to put the dots before the first optional argument. It's possible that a few people might have relied on partial or position based matching, but `check_dots_used()` will catch that.

hadley (commented):
Looks good — one comment about the position of `...` and then can you please add a bullet to NEWS?

R/chop.R
hadley: https://design.tidyverse.org/dots-position.html

mgirlich: In `build_wider_spec()` I put the dots after `names_from` and `values_from` as they felt like main arguments and that's also how they were used in the test *names_glue affects output names*

PR #1126, opened 2021-06-11
Speed up `unchop()`

mgirlich:
``` r
library(tidyr)
set.seed(1)

n <- 100e3
df <- tibble(
  a = purrr::map(1:n, ~ rlang::seq2(1, sample(0:10, 1))),
  y = 1:n
)

bench::mark(
  drop_empty = unchop(df, a),
  keep_empty = unchop(df, a, keep_empty = TRUE),
  check = FALSE
)

# this PR
#> # A tibble: 2 x 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 drop_empty   72.2ms   72.9ms      13.7    25.4MB     34.3
#> 2 keep_empty   81.4ms   81.4ms      12.3      29MB     86.0

# CRAN
#> # A tibble: 2 x 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 drop_empty 642.71ms 642.71ms     1.56     13.1MB     21.8
#> 2 keep_empty    1.02s    1.02s     0.979    13.9MB     22.5
```

<sup>Created on 2021-06-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

### DIFF
--- a/R/chop.R
+++ b/R/chop.R
@@ -95,23 +95,35 @@ unchop <- function(data, cols, keep_empty = FALSE, ptype = NULL) {
     return(data)
   }
 
+  col_sizes <- map(unclass(data)[cols], tidyr_sizes)
+
   if (keep_empty) {
-    for (col in cols) {
-      data[[col]][] <- map(data[[col]], init_col)
+    for (i in seq_along(cols)) {
+      col <- cols[[i]]
+
+      na_flag <- is.na(col_sizes[[i]])
+      data[[col]][na_flag] <- NA
+
+      empty_flag <- !na_flag & col_sizes[[i]] == 0
+      data[[col]][empty_flag] <- list(init_col(vec_c(!!!data[[col]][empty_flag])))
+
+      col_sizes[[i]][na_flag | empty_flag] <- 1L
     }
   }
 
   # In case `x` is a grouped data frame and any `cols` are lists,
   # in which case `[.grouped_df` will error
   cols <- new_data_frame(unclass(data)[cols])
+  out <- dplyr::select(data, -names(dplyr::all_of(cols)))
 
-  res <- df_unchop_info(cols, ptype)
+  res <- df_unchop_info(cols, ptype, col_sizes)
   new_cols <- res$val
   slice_loc <- res$loc
 
-  out <- vec_slice(data, slice_loc)
+  out <- vec_slice(out, slice_loc)
 
   out <- update_cols(out, new_cols)
+  out <- dplyr::relocate(out, colnames(data))
   reconstruct_tibble(data, out)
 }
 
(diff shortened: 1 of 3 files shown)


hadley:
How close a review do you think this needs? i.e. do you have any worries that this might change behaviour in some edge case? If so, I probably won't be able to review it until I'm next work on tidyr, since I'll need to load it all up into my brain again.

mgirlich:
There are quite a lot of changes, so a review would definitely make sense. But it's fine for me to wait until you have time to work again on tidyr.

R/nest.R
```diff
@@ -332,18 +332,9 @@ unnest.data.frame <- function(
...
 
-  if (nrow(data) == 0) {
-    for (col in names(cols)) {
-      data[[col]] <- as_empty_df(data[[col]], col = col)
-    }
-  } else {
-    for (col in names(cols)) {
-      data[[col]] <- map(data[[col]], as_df, col = col)
-    }
-  }
```
DavisVaughan, line 343: There are some weird edge cases that might break if we remove this. For example, having a mix of vectors and tibbles in the list-col you are trying to unnest currently works like this:

``` r
library(tidyr)

# outer column name of `x` is used to turn `1` into `data.frame(x = 1)`
# then the unchop+unpack is done
df <- tibble(x = list(1, tibble(y = 2)))
unnest(df, x)
#> # A tibble: 2 x 2
#>       x     y
#>   <dbl> <dbl>
#> 1     1    NA
#> 2    NA     2

# again we get `data.frame(x = 1)`, which then row binds with the tibble
# since it also uses `x` as the column name
df <- tibble(x = list(1, tibble(x = 1)))
unnest(df, x)
#> # A tibble: 2 x 1
#>       x
#>   <dbl>
#> 1     1
#> 2     1

# if the column types are different, that would be an error, which leads
# me to believe that this is a "risky" feature to be relying on
df <- tibble(x = list(1, tibble(x = "x")))
unnest(df, x)
#> Error: Can't combine `..1$x` <double> and `..2$x` <character>.
```

The first two would become errors if this code was removed.

Granted, this seems like strange behavior to rely on, so maybe we can just stop supporting it? FWIW `unnest_legacy()` errors in all of these cases, which makes me think that it would be alright to not support this.

Removing this bit of code does greatly improve performance, so it would be nice to do so if possible.

CC @hadley - not sure if you can think of any other edge cases where this code mattered off the top of your head.

hadley: I have zero recollection 😢 If removing it doesn't breaking any tests and has minimal impact on revdeps, I think it's fine to remove it. `unnest()`ing a mix of vectors and data frames doesn't seem terribly useful, and it does feel like it works by accident currently.

PR #1133, opened 2021-06-17
Speed up `unnest_longer()` and `unnest_wider()`

mgirlich:
Changes
* replace `map()` and `compact()` with `lapply()` and `vec_slice() + list_sizes()`
* replace `tibble()` with `new_data_frame()`

They should also benefit nicely from #1126.

## Benchmarks

``` r
library(purrr)
library(tidyr)

array_fields <- c("titles", "aliases", "allegiances", "books", 
                  "povBooks", "tvSeries", "playedBy")
got_chars2 <- map(repurrrsive::got_chars, 
                  ~map_at(.x, array_fields, ~as.list(.x)))

times <- tibble(
  method = character(),
  rows = integer(),
  time = numeric()
)

data <- tibble(data = rep(got_chars2, length.out = 5000))

bench::mark(
  unnest_longer = unnest_longer(data, data),
  unnest_wider = unnest_wider(data, data, simplify = TRUE),
  check = FALSE
)

# This PR
#> # A tibble: 2 x 6
#>   expression         min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>    <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 unnest_longer    192ms    199ms      4.98    5.55MB     16.6
#> 2 unnest_wider     911ms    911ms      1.10    14.6MB     24.1

# CRAN
#> # A tibble: 2 x 6
#>   expression         min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>    <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 unnest_longer    2.33s    2.33s     0.430      12MB     15.5
#> 2 unnest_wider     2.33s    2.33s     0.430    17.4MB     23.2
```

<sup>Created on 2021-06-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

### DIFF
--- a/R/rectangle.R
+++ b/R/rectangle.R
@@ -426,11 +426,11 @@ vec_to_wide <- function(x, col, names_sep = NULL) {
   }
 
   if (is.data.frame(x)) {
-    as_tibble(map(x, list))
+    as_tibble(lapply(x, list))
   } else if (vec_is(x)) {
     if (is.list(x)) {
-      x <- purrr::compact(x)
-      x <- map(x, list)
+      x <- vec_slice(x, list_sizes(x) != 0)
+      x <- lapply(x, list)
     } else {
       x <- as.list(x)
     }
--- a/R/rectangle.R
+++ b/R/rectangle.R
@@ -402,7 +402,7 @@ simplify_col <- function(x, nm, ptype = list(), transform = list(), simplify = F
     n <- map_int(x, vec_size)
   }
 
-  if (!all(n %in% c(0, 1))) {
+  if (any(n > 1)) {
     if (is.null(ptype)) {
       return(x)
     } else {
--- a/R/rectangle.R
+++ b/R/rectangle.R
@@ -377,7 +377,7 @@ simplify_col <- function(x, nm, ptype = list(), transform = list(), simplify = F
 
   # Don't simplify lists of lists, because that typically indicates that
   # there might be multiple values.
-  is_list <- map_lgl(x, is.list)
+  is_list <- vapply(x, is.list, logical(1L))
   if (any(is_list)) {
     if (is.null(ptype)) {
       return(x)
@@ -387,7 +387,7 @@ simplify_col <- function(x, nm, ptype = list(), transform = list(), simplify = F
   }
 
   # Don't try and simplify non-vectors
-  is_vec <- map_lgl(x, ~ vec_is(.x) || is.null(.x))
+  is_vec <- vapply(x, function(.x) vec_is(.x), logical(1L))
   if (any(!is_vec)) {
     if (is.null(ptype)) {
       return(x)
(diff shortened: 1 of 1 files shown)


R/rectangle.R
```diff
@@ -393,7 +396,12 @@ simplify_col <- function(x, nm, ptype = list(), transform = list(), simplify = F
     }
   }
 
-  n <- map_int(x, vec_size)
+  if (vec_is_list(x)) {
+    n <- list_sizes(x)
+  } else {
+    n <- map_int(x, vec_size)
```
hadley, line 402: I don't quite follow the logic here — if `x` is not a list then I'm guessing it's either an atomic vector or a data frame? So why do we need the map?

R/rectangle.R
```diff
@@ -426,15 +434,15 @@ vec_to_wide <- function(x, col, names_sep = NULL) {
   }
 
   if (is.data.frame(x)) {
-    as_tibble(map(x, list))
+    new_data_frame(lapply(x, list))
```
hadley, line 437: The coercion to tibble still happens, but elsewhere?

R/rectangle.R
```diff
@@ -426,15 +434,15 @@ vec_to_wide <- function(x, col, names_sep = NULL) {
...
   } else if (vec_is(x)) {
     if (is.list(x)) {
-      x <- purrr::compact(x)
-      x <- map(x, list)
+      x <- vec_slice(x, list_sizes(x) != 0)
```
hadley, line 440: Maybe worth pulling out into a local `compact` helper with a comment explaining the perf benefit? Or maybe vctr should provide `list_compact()`?

hadley (commented):
Thanks for working on this!

R/rectangle.R
mgirlich: Not directly but this doesn't matter. `vec_to_wide()` is only called from `unnest_wider()` and there the data frame is actually unpacked at the end.

If you are interested here a quick overview of `unnest_wider()`
```r
# make sure that `col` is a list-col of data frames
data[[col]] <- map(data[[col]], vec_to_wide)

# `unchop()` turns `col` into a df-col
data <- unchop(data, any_of(col), keep_empty = TRUE)

# simplify every column of the df-col `col`
data[[col]][] <- map(data[[col]], simplify_col)

# `unpack()` expands the df-col into individual columns
# -> it doesn't matter whether `col` is a data frame or a tibble
unpack(data, any_of(col), names_repair = names_repair)
```

R/rectangle.R
mgirlich: I pulled it into a compact helper. I also created a `vctrs` issue (https://github.com/r-lib/vctrs/issues/1395) but according to lionel it is out of scope for `vctrs`, at least for now.

PR #1137, opened 2021-07-13
Switch column order from pivot_longer() so indices are before values

JohannesNE:
This is a small thing, but I find it awkward that indices are after values in the output from pivot_longer().

If there is a good reason for this order, or you do not think this (potentially breaking) change is "worth it", just reject :)
I guess the main risk is that people may rely on column indices rather than names.


DavisVaughan:
I think we should probably keep this as is to prevent any potential breaking changes

Issue #1112, opened 2021-03-24
unnest column of data.frames


geotheory:
unnest() does not work with data.frame columns.

---

`unnest()` works with list column of nested data.frames, but fails with a _data.frame_ column for the same data.  Given that this is a common data format (e.g. typical `fromJSON()` output) it feels like an unnecessary trap for the unaware.

The documentation does specify "list-column" in the title of the documentation, but otherwise this distinction isn't column-type isn't discussed.  It feels to me that a column of nested data.frames is quite likely to be confused with a _list_ column of nested data.frames.

```r
require(jsonlite)
require(tidyr)
require(dplyr)

j = '{
  "observations": [{
    "id": "a",
    "data": {
      "count": 49,
      "max": 100
    }
  }, {
    "id": "b",
    "data": {
      "count": 93,
      "max": 120
    }
  }, {
    "id": "c",
    "data": {
      "count": 27,
      "max": 88
    }
  }]
}'

d = fromJSON(j)$observations %>% as_tibble()

glimpse(d)
#> Rows: 3
#> Columns: 2
#> $ id   <chr> "a", "b", "c"
#> $ data <df[,2]> <data.frame[3 x 2]>

d %>% unnest(data)
#> Error: Assigned data `map(data[[col]], as_df, col = col)` must be compatible with existing data.
#> x Existing data has 3 rows.
#> x Assigned data has 2 rows.
#> ℹ Only vectors of size 1 are recycled.

d %>% rowwise() %>% mutate(data = list(data)) %>% unnest(data)
#> # A tibble: 3 x 3
#>   id    count   max
#>   <chr> <int> <int>
#> 1 a        49   100
#> 2 b        93   120
#> 3 c        27    88
```

DavisVaughan:
The error happens here in `unnest()`:

https://github.com/tidyverse/tidyr/blob/2fd80d51351cef55cbd0d5ae8ddde58a8fb6be48/R/nest.R#L341

That can probably be fixed, but the bigger issue is that eventually this will call `unchop()`, and that won't support data frames until we have `vctrs::vec_slice2()`, which I noted here:

https://github.com/tidyverse/tidyr/blob/2fd80d51351cef55cbd0d5ae8ddde58a8fb6be48/R/chop.R#L122-L124

We could try to pull in the POC implementation of `vec_slice2()` from here, it doesn't seem too complicated but I'm not sure if it is ready for prime time or not. https://github.com/r-lib/vctrs/pull/1228

[...]

PR #1140, opened 2021-08-05
Improve `unchop()` performance

### LINKED ISSUE #1127: Improve `unchop()` and `unnest()` performance
`unnest()` and `unchop()` can be sped up quite a bit.

Using the same reprex as in #751

```r
library(tidyr)
n <- 100000

df <- tibble(
  g = 1:n,
  y = rep(list(tibble(x = 1:5)), n)
)

bench::mark(
  unnest(df, y),
  unnest_legacy(df, y),
  check = TRUE
)


# before
#> # A tibble: 2 x 13
#>   expression                min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
#>   <bch:expr>           <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>
#> 1 unnest(df, y)          90.6ms   96.3ms     10.1     2.85MB     20.3     6    12
#> 2 unnest_legacy(df, y)  134.8ms  141.2ms      7.08    4.15MB     23.0     4    13

# after
#> # A tibble: 2 x 13
#>   expression                min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
#>   <bch:expr>           <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>
#> 1 unnest(df, y)          30.2ms   33.8ms     29.6     4.26MB     13.8    15     7
#> 2 unnest_legacy(df, y)  120.6ms  124.1ms      7.66     3.6MB     26.8     4    14
```

### LINKED ISSUE #1128: `unchop()` doesn't work for data frame columns
``` r
library(tidyr)

df <- tibble(x = tibble(a = 1:2, b = "a"), y = list(3:4))
unchop(df, x)
#> Error: Can't combine `..1` <integer> and `..2` <character>.
```

<sup>Created on 2021-06-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

DavisVaughan:
Closes #1126 
Closes #1127 
Closes #1128 
A step towards #1112

@mgirlich I think you had some great ideas for improving `unchop()` in #1126! I've tweaked them a little bit and added them here, along with many additional edge case tests, and gave you credit in the NEWS file.

I haven't touched `unnest()` yet, I'll save that for another PR, but I think you are right that we could remove that `as_df()` code and just have `unpack()` only unpack the remaining df-cols.

The original plan for `df_unchop()` was to move it to C in vctrs, so I wrote it somewhat inefficiently with R level loops (which would have been pretty efficient at the C level). I'm no longer convinced that we should move it to vctrs because of how it peeks at the list-of ptype attribute (vctrs C code tries to be agnostic to these details, but that would be impossible here).

---

@mgirlich I've fixed a few edge cases that I discovered in yours, such as:

```r
# expect 0 row tibble with `x=integer()` column
df <- tibble(x = list(NULL, NULL))
unchop(df, x, ptype = tibble(x = integer()))
#> Error in x[needs_recycling] <- map2(x[needs_recycling], lengths_out[needs_recycling],  : 
#>   NAs are not allowed in subscripted assignments

# (probably similar to above)
df <- tibble(x = 1:2, y = list(NULL, NULL), z = list(NULL, NULL))
expect <- tibble(x = integer(), y = unspecified(), z = unspecified())
expect_identical(unchop(df, c(y, z)), expect)
#> Error in x[needs_recycling] <- map2(x[needs_recycling], lengths_out[needs_recycling],  : 
#>   NAs are not allowed in subscripted assignments

# this should return a column of double NAs, not logical NAs
df <- tibble(x = 1:2, y = list(integer(), double()))
unchop(df, y, keep_empty = TRUE)
#> # A tibble: 2 x 2
#>       x y    
#>   <int> <lgl>
#> 1     1 NA   
#> 2     2 NA   
```


R/chop.R
```diff
@@ -37,9 +37,9 @@
 #'   will be dropped from the output. If you want to preserve all rows,
 #'   use `keep_empty = TRUE` to replace size-0 elements with a single row
 #'   of missing values.
-#' @param ptype Optionally, supply a data frame prototype for the output `cols`,
-#'   overriding the default that will be guessed from the combination of
-#'   individual values.
+#' @param ptype Optionally, a named list of column name-prototype pairs to
+#'   coerce `cols` to, overriding the default that will be guessed from
+#'   combining the individual values.
```
DavisVaughan, line 42: One subtle change is that `ptype` no longer has to be a data frame. It is now just a named list where the names can match any names specified by `cols`. This matches `pivot_longer()`.

Previously we required a data frame here, but we had some strange behavior where we would cast the unchopped data frame to `ptype`, which could add columns that only ever existed in the ptype. That seemed very strange, and this feels like much better behavior for this argument.

R/chop.R
```diff
@@ -90,188 +90,253 @@ chop <- function(data, cols) {
...
-  }
+  sel <- tidyselect::eval_select(enquo(cols), data)
 
-  if (keep_empty) {
-    for (col in cols) {
-      data[[col]][] <- map(data[[col]], init_col)
-    }
+  if (is_null(ptype)) {
+    # Convert to form that `df_unchop()` requires
+    ptype <- list()
```
DavisVaughan, line 97: As an internal detail, I've made `df_unchop()` always take a list `ptype`, so we convert that here

R/chop.R
```diff
@@ -90,188 +90,253 @@ chop <- function(data, cols) {
...
+  names <- names(data)
+
+  # In case `data` is a grouped data frame and any `cols` are lists,
   # in which case `[.grouped_df` will error
-  cols <- new_data_frame(unclass(data)[cols])
+  out <- new_data_frame(data, n = size)
+  cols <- out[sel]
+
+  # Remove unchopped columns to avoid slicing them needlessly later
+  out[sel] <- NULL
```
DavisVaughan, line 109: @mgirlich had the nice idea that we should remove the columns that get unchopped up front. That way when we do `vec_slice(out, loc)` later on, we don't needlessly slice those columns (since we are going to immediately replace them with `update_cols()`)

R/chop.R
```diff
@@ -90,188 +90,253 @@ chop <- function(data, cols) {
...
 
-    for (j in seq_len_size) {
-      # TODO: col[[j]] -> vec_slice2(col, j)
-      elt <- col[[j]]
-      size <- sizes[[j]]
-
-      # Recycle each row element to the common size of that row
-      pieces[[j]] <- tidyr_recycle(elt, size)
+    if (!col_is_list) {
+      out_cols[[i]] <- vec_slice(col, out_loc)
```
DavisVaughan, line 211: Nice optimization also pointed out by @mgirlich that while unchopping non-list columns is kind of a silly thing for the user to do, it can be done a lot faster by just slicing the column. Previously I think we essentially converted it to a list, recycled each element, and then converted it back to a vector.

This will actually help performance with https://github.com/tidyverse/tidyr/issues/1112. The user is trying to `unnest()` a data frame column (really they should use `unpack()`, but `unnest()` should work too). The `unchop()` step of that will have them unchopping a df-col, which will be faster because of this optimization.

R/chop.R
```diff
@@ -90,188 +90,253 @@ chop <- function(data, cols) {
...
+      # This can happen when both of these are true:
+      # - `col` was an empty list(), or a list of all `NULL`s.
+      # - No ptype was specified for `col`, either by the user or by a list-of.
+      if (out_size != 0L) {
+        abort("Internal error: `NULL` column generated, but output size is not `0`.")
+      }
 
-  loc <- rep(seq_len_size, sizes)
+      col <- unspecified(0L)
+    }
```
DavisVaughan, line 236: This edge case is mainly a side effect of the fact that `vec_unchop(list())` and `vec_unchop(list(NULL))` return `NULL`. There isn't much we can do about this, and I feel like it is the right vctrs behavior, but we can't insert a `NULL` as a data frame column so the correct alternative is to insert an unspecified object. It can come up when the user tries to unchop `tibble(x = list())` or `tibble(x = list(NULL))`, where there is clearly no obvious type for `x`.

R/chop.R
```diff
@@ -90,188 +90,253 @@ chop <- function(data, cols) {
...
+  list(col = col, sizes = sizes, null = null)
 }
 
-tidyr_recycle <- function(x, size) {
-  if (is.null(x)) {
-    unspecified(size)
-  } else {
-    vec_recycle(x, size)
+unchop_sizes2 <- function(x, y) {
+  # Standard tidyverse recycling rules, just vectorized.
```
DavisVaughan, line 298: Something that is different between this implementation and the previous one (and @mgirlich's) is that we now use standard recycling rules when computing the majority of the common sizes for the rows. No more awkward special handling of `NA_integer_` sizes!

To be able to do this, we do a pre-adjustment to `NULL` elements in `unchop_col_info()`, and then a post-adjustment to entirely `NULL` rows in `unchop_finalize()`

tests/testthat/test-chop.R
```diff
@@ -100,33 +130,113 @@ test_that("correctly performs tidy recycling with size 1 inputs", {
...
+test_that("nonexistent `ptype` columns are ignored", {
   df <- tibble(x = 1, y = list(1, 2))
-  ptype <- tibble(y = numeric(), z = numeric())
-
-  expect <- tibble(x = c(1, 1), y = c(1, 2), z = c(NA_real_, NA_real_))
+  ptype <- list(y = numeric(), z = numeric())
 
-  expect_identical(unchop(df, y, ptype = ptype), expect)
+  expect_identical(unchop(df, y, ptype = ptype), unchop(df, y))
 })
```
DavisVaughan, line 138: This is the weird `ptype` behavior that we used to have. Notice that `z` was a column that only existed in the `ptype`, but that forced it to exist in the `unchop()` result as an all missing column. That seemed pretty silly, so I've removed that behavior.

mgirlich (commented):
Looks great to me. Very readable code. Thanks for improving my PR!

DavisVaughan:
No changes in revdeps (besides those caused by #1142)!

DavisVaughan:
@hadley, like #1126 this has significant changes, so I don't expect a full review until you get back to working on tidyr. No changes in revdeps though.

PR #1141, opened 2021-08-06
Improve `unnest()` performance

### LINKED ISSUE #1112: unnest column of data.frames

DavisVaughan:
Should be rebased on master after #1140 is merged. Then remove the `skips()` from the new tests.
Will benefit from the performance improvement of `unchop()` too.
Part of #1127 
Closes #1112 

This removes the `as_df()` and `as_empty_df()` helpers from `unnest()`. Their main purpose seemed to be ensuring that each element of the list-col was a data frame, so that `unchop()` would return a packed df-col that `unpack()` would unpack. This doesn't seem to be necessary. Instead, we just let `unchop()` unchop the list-cols (it "skips" over any non-list-cols) and then we tell `unpack()` the remaining df-cols to unpack.

As mentioned in https://github.com/tidyverse/tidyr/pull/1126#discussion_r683810282, there are a few edge cases that this would break. But they seemed to be off-label usage, only worked by accident, and `unnest_legacy()` didn't work that way either, so I _think_ we are okay. I'll run revdeps to be sure. I've also added a test of the new error behavior.


R/nest.R
```diff
@@ -329,20 +329,9 @@ unnest.data.frame <- function(
...
-    }
-  } else {
-    for (col in names(cols)) {
-      data[[col]] <- map(data[[col]], as_df, col = col)
-    }
-  }
-
   data <- unchop(data, any_of(cols), keep_empty = keep_empty, ptype = ptype)
+  cols <- cols[map_lgl(unclass(data)[cols], is.data.frame)]
   unpack(data, any_of(cols), names_sep = names_sep, names_repair = names_repair)
```
DavisVaughan, line 335: If `unpack()` was smart enough to "skip" any non df-cols (like how unchop "skips" non list-cols) then we wouldn't even need this filtering line! For example, this `unpack()` call _could_ return its input unchanged.

``` r
library(tidyr)

df <- tibble(x = 1L, y = 1)

# "skips" unchopping y
unchop(df, y)
#> # A tibble: 1 x 2
#>       x     y
#>   <int> <dbl>
#> 1     1     1

# requires df cols
unpack(df, y)
#> Error: `y` must be a data frame column
```

That would make `unnest()` truly an `unchop()` + `unpack()` without any intermediate adjustments, which theoretically is kind of nice.

tests/testthat/test-nest.R
```diff
@@ -233,15 +244,29 @@ test_that("can unnest empty data frame", {
...
+  skip("until rebased on #1140, which fixes `unchop()` with `list_of(NULL)` cols")
+  df <- tibble(x = 1L, y = list_of(NULL, .ptype = tibble(a = integer())))
+  out <- unnest(df, y)
+  expect_identical(out, tibble(x = integer(), a = integer()))
+})
+
+test_that("skips over vector columns", {
   df <- tibble(x = integer(), y = list())
-  expect_error(unnest(df, x), "list of vectors")
+  expect_identical(unnest(df, x), df)
```
DavisVaughan, line 269: This one used to be an error, but only in the size 0 case? i.e. on master:

``` r
library(tidyr)

df <- tibble(x = 1L, y = 1)
unnest(df, y)
#> # A tibble: 1 x 2
#>       x     y
#>   <int> <dbl>
#> 1     1     1

df <- tibble(x = integer(), y = double())
unnest(df, y)
#> Error: Input must be list of vectors
```

`unnest()` is generally pretty forgiving about trying to unnest a non list-col (which pretty much results in a no-op), so I think this should work.

R/nest.R
mgirlich: This would also be great to easily unpack all data frame columns `unpack(df, everything())` and in general less need to worry when using selection helpers

mgirlich:
LGTM! Looking forward to the speed improvements in `tidyr` 😄

PR #1143, opened 2021-08-06
Update GitHub actions

DavisVaughan:
- Updates the standard GitHub actions files
- Updates the devel dplyr GitHub actions file

I am hoping that updating the devel dplyr check file will fix errors like:
https://github.com/tidyverse/tidyr/pull/1141/checks?check_run_id=3263072379


hadley (approved):
Seems fine to me. I didn't read it closely; I assume if it works it's good to go.

PR #1148, opened 2021-08-13
update cheat sheet links

averiperny:
tidyr is now a separate cheat sheet, this updates the link and the thumbnail image.


hadley:
Thanks!

Issue #1185, opened 2021-11-02
Use `glue()` rather than `paste0()` for `names_sep` application


DavisVaughan:
To avoid bugs in the zero length recycling case, where `paste0()` doesn't follow tidyverse recycling rules

``` r
library(tidyr)

df <- tibble(y = list(1L, 1L))
unnest_wider(df, y, names_sep = "_")
#> # A tibble: 2 × 1
#>     y_1
#>   <int>
#> 1     1
#> 2     1

df <- tibble(y = list(1L, integer()))
unnest_wider(df, y, names_sep = "_")
#> Error in names(x) <- paste0(col, names_sep, index(x)): 'names' attribute [1] must be the same length as the vector [0]


x <- integer()

# doesn't recycle to size 0
paste0("y", "_", index(x))
#> [1] "y_"

names(x) <- paste0("y", "_", index(x))
#> Error in names(x) <- paste0("y", "_", index(x)): 'names' attribute [1] must be the same length as the vector [0]
```

Try to replace all usage of `paste0()` when used this way

Issue #835, opened 2019-12-13
FR: Rectangling a list according to a specification


mgirlich:
Some time ago I have written a function for rectangling a nested list according to a given specification. I basically like `hoist()` but I think that something like my list parser could also be a nice addition.

A specification basically has the following form:

``` r
lcols(
  `col_name = lcol_*(pluck_specification)`,
  `col_name = lcol_*(pluck_specification)`,
  ...
)
```

(I'm not that keen on the name but didn't think of anything better yet). The functions `lcol_*` are like follows

* `lcol_lgl(path)`: creates a logical column.
* similarly `lcol_int/dbl/chr/fct`, `dat` for date, `dtt` for datetime, `hms`, `guess`.
* `lcol_lst_of`(path, .ptype): creates a list-of-<ptype> column.
* `lcol_df(path, ...)` creates a dataframe column where `...` are `lcol_*` functions again.
* `lcol_df_lst` similar to `lcol_df` but creates a list of dataframe column.

These functions all have the arguments:
* `.req`: if it is `TRUE` and the path is missing an error is thrown (basically `chuck` is used instead of `pluck`).
* `.parser` can be used for a custom conversion function. This can be especially useful for date or datetime columns

For example to parse the `discog` data from `repurrrsive` this could like below:

``` r
recordlist <- repurrrsive::discog

col_specs <- lcols(
  lcol_int("id"),
  first_artist_name = lcol_chr(list("basic_information", "artists", 1, "name")),
  artists_df_lst = lcol_df_lst(
    c("basic_information", "artists"),
    lcol_chr("join"),
    lcol_chr("name"),
    lcol_chr("role"),
    lcol_int("id")
  ),
  rating_list = lcol_lst_of("rating", .ptype = ""),
  info = lcol_df(
    "basic_information",
    lcol_int("id"),
    lcol_chr("title")
  ),
)

list_parse(recordlist, col_specs)
#> # A tibble: 155 x 5
#>         id first_artist_name   artists_df_lst rating_list info$id $title        
#>      <int> <chr>               <list<df[,4]>> <list<chr>>   <int> <chr>         
#>  1  7.50e6 Mollot                     [1 × 4]         [1]  7.50e6 Demo          
#>  2  4.49e6 Una Bèstia Incontr…        [1 × 4]         [1]  4.49e6 Observant Com…
#>  3  9.83e6 S.H.I.T. (3)               [1 × 4]         [1]  9.83e6 I             
[...]

PR #1200, opened 2021-11-04
Rectangling function rewrite

### LINKED ISSUE #740: unnest multiple columns
See #737 for full example

Given the overall extreme flexibility of the new functions I was surprised that `unnest_wider` doesn't allow unnesting several columns.

``` r
library(tidyverse)
df2 <- tibble::tribble(
  ~id, ~var1,              ~Country,                        ~Sport,             ~Format,
  10L,  169L, c("Norway", "Sweden"),                        "Skii", c("Video", "Photo"),
  11L,  150L,               "Spain", c("Bike", "Soccer", "Basket"),             "Photo",
  12L,    0L,                 "USA",                    "Baseball",             "Video"
)

df2 %>%
  unnest_wider(Country, names_sep = "_") %>%
  unnest_wider(Sport, names_sep = "_") %>%
  unnest_wider(Format, names_sep = "_")
#> New names:
#> * `` -> ...1
#> * `` -> ...2
#> New names:
#> * `` -> ...1
#> New names:
#> * `` -> ...1
#> New names:
#> * `` -> ...1
#> New names:
#> * `` -> ...1
#> * `` -> ...2
#> * `` -> ...3
#> New names:
#> * `` -> ...1
#> New names:
#> * `` -> ...1
#> * `` -> ...2
#> New names:
#> * `` -> ...1
#> New names:
#> * `` -> ...1
#> # A tibble: 3 x 9
#>      id  var1 Country_...1 Country_...2 Sport_...1 Sport_...2 Sport_...3
#>   <int> <int> <chr>        <chr>        <chr>      <chr>      <chr>     
#> 1    10   169 Norway       Sweden       Skii       <NA>       <NA>      
#> 2    11   150 Spain        <NA>         Bike       Soccer     Basket    
#> 3    12     0 USA          <NA>         Baseball   <NA>       <NA>      
#> # ... with 2 more variables: Format_...1 <chr>, Format_...2 <chr>
```

If I didn't want to enumerate explicitly I could design a complex `reduce` call but it would be neat to be able to just call : 

```
df2 %>%
  unnest_wider(one_of("Country","Sport", "Format"), names_sep = "_")
```

In case this isn't acceptable, I think we could use a more helpful message than 

> Error in .subset2(x, i) : no such index at level 2

### LINKED ISSUE #996: hoist(): `.ptype` only respected if `.simplify = TRUE`
The `.ptype` argument has no effect for `.simplify = FALSE`

``` r
library(tidyverse)

df <- tibble(
  list_col = list(
    list(age = 1L),
    list(age = 2L)
  )
)

# every element of age is still an integer
df %>% 
  hoist(list_col,
    "age",
    .ptype = list(age = double()),
    .simplify = FALSE
  )
#> # A tibble: 2 x 1
#>   age      
#>   <list>   
#> 1 <int [1]>
#> 2 <int [1]>

df %>% 
  hoist(list_col,
    "age",
    .ptype = list(age = double()),
    .simplify = TRUE
  )
#> # A tibble: 2 x 1
#>     age
#>   <dbl>
#> 1     1
#> 2     2
```

<sup>Created on 2020-07-16 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

<details>

<summary>Session info</summary>

``` r
devtools::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value                       
#>  version  R version 4.0.1 (2020-06-06)
#>  os       macOS Catalina 10.15.5      
#>  system   x86_64, darwin17.0          
#>  ui       X11                         
#>  language (EN)                        
#>  collate  en_US.UTF-8                 
#>  ctype    en_US.UTF-8                 
#>  tz       UTC                         
#>  date     2020-07-16                  
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version    date       lib source                          
#>  assertthat    0.2.1      2019-03-21 [1] CRAN (R 4.0.0)                  
#>  backports     1.1.8      2020-06-17 [1] CRAN (R 4.0.1)                  
#>  blob          1.2.1      2020-01-20 [1] CRAN (R 4.0.0)                  
#>  broom         0.7.0      2020-07-09 [1] CRAN (R 4.0.1)                  
#>  callr         3.4.3      2020-03-28 [1] CRAN (R 4.0.0)                  
#>  cellranger    1.1.0      2016-07-27 [1] CRAN (R 4.0.0)                  
#>  cli           2.0.2      2020-02-28 [1] CRAN (R 4.0.0)                  
#>  colorspace    1.4-1      2019-03-18 [1] CRAN (R 4.0.0)                  
#>  crayon        1.3.4      2017-09-16 [1] CRAN (R 4.0.0)                  
#>  DBI           1.1.0      2019-12-15 [1] CRAN (R 4.0.0)                  
#>  dbplyr        1.4.4      2020-05-27 [1] CRAN (R 4.0.0)                  
#>  desc          1.2.0      2018-05-01 [1] CRAN (R 4.0.0)                  
#>  devtools      2.3.0      2020-04-10 [1] CRAN (R 4.0.0)                  
#>  digest        0.6.25     2020-02-23 [1] CRAN (R 4.0.0)                  
#>  dplyr       * 1.0.0      2020-05-29 [1] CRAN (R 4.0.0)                  
#>  ellipsis      0.3.1      2020-05-15 [1] CRAN (R 4.0.0)                  
#>  evaluate      0.14       2019-05-28 [1] CRAN (R 4.0.0)                  
#>  fansi         0.4.1      2020-01-08 [1] CRAN (R 4.0.0)                  
#>  forcats     * 0.5.0      2020-03-01 [1] CRAN (R 4.0.0)                  
#>  fs            1.4.2      2020-06-30 [1] CRAN (R 4.0.1)                  
#>  generics      0.0.2      2018-11-29 [1] CRAN (R 4.0.0)                  
#>  ggplot2     * 3.3.2      2020-06-19 [1] CRAN (R 4.0.1)                  
#>  glue          1.4.1      2020-05-13 [1] CRAN (R 4.0.0)                  
#>  gtable        0.3.0      2019-03-25 [1] CRAN (R 4.0.0)                  
#>  haven         2.3.1      2020-06-01 [1] CRAN (R 4.0.0)                  
#>  highr         0.8        2019-03-20 [1] CRAN (R 4.0.0)                  
#>  hms           0.5.3      2020-01-08 [1] CRAN (R 4.0.0)                  
#>  htmltools     0.5.0      2020-06-16 [1] CRAN (R 4.0.1)                  
#>  httr          1.4.1      2019-08-05 [1] CRAN (R 4.0.0)                  
#>  jsonlite      1.7.0      2020-06-25 [1] CRAN (R 4.0.1)                  
#>  knitr         1.29       2020-06-23 [1] CRAN (R 4.0.1)                  
#>  lifecycle     0.2.0      2020-03-06 [1] CRAN (R 4.0.0)                  
#>  lubridate     1.7.9      2020-06-08 [1] CRAN (R 4.0.1)                  
#>  magrittr      1.5        2014-11-22 [1] CRAN (R 4.0.0)                  
#>  memoise       1.1.0      2017-04-21 [1] CRAN (R 4.0.0)                  
#>  modelr        0.1.8      2020-05-19 [1] CRAN (R 4.0.0)                  
#>  munsell       0.5.0      2018-06-12 [1] CRAN (R 4.0.0)                  
#>  pillar        1.4.6      2020-07-10 [1] CRAN (R 4.0.1)                  
#>  pkgbuild      1.1.0      2020-07-13 [1] CRAN (R 4.0.1)                  
#>  pkgconfig     2.0.3      2019-09-22 [1] CRAN (R 4.0.0)                  
#>  pkgload       1.1.0      2020-05-29 [1] CRAN (R 4.0.0)                  
#>  prettyunits   1.1.1      2020-01-24 [1] CRAN (R 4.0.0)                  
#>  processx      3.4.3      2020-07-05 [1] CRAN (R 4.0.0)                  
#>  ps            1.3.3      2020-05-08 [1] CRAN (R 4.0.0)                  
#>  purrr       * 0.3.4      2020-04-17 [1] CRAN (R 4.0.1)                  
#>  R6            2.4.1      2019-11-12 [1] CRAN (R 4.0.0)                  
#>  Rcpp          1.0.5      2020-07-06 [1] CRAN (R 4.0.1)                  
#>  readr       * 1.3.1      2018-12-21 [1] CRAN (R 4.0.0)                  
#>  readxl        1.3.1      2019-03-13 [1] CRAN (R 4.0.0)                  
#>  remotes       2.1.1      2020-02-15 [1] CRAN (R 4.0.0)                  
#>  reprex        0.3.0      2019-05-16 [1] CRAN (R 4.0.0)                  
#>  rlang         0.4.7      2020-07-09 [1] CRAN (R 4.0.1)                  
#>  rmarkdown     2.3        2020-06-18 [1] CRAN (R 4.0.1)                  
#>  rprojroot     1.3-2      2018-01-03 [1] CRAN (R 4.0.0)                  
#>  rvest         0.3.5      2019-11-08 [1] CRAN (R 4.0.0)                  
#>  scales        1.1.1      2020-05-11 [1] CRAN (R 4.0.0)                  
#>  sessioninfo   1.1.1      2018-11-05 [1] CRAN (R 4.0.0)                  
#>  stringi       1.4.6      2020-02-17 [1] CRAN (R 4.0.0)                  
#>  stringr     * 1.4.0      2019-02-10 [1] CRAN (R 4.0.0)                  
#>  testthat      2.3.2      2020-03-02 [1] CRAN (R 4.0.0)                  
#>  tibble      * 3.0.3      2020-07-10 [1] CRAN (R 4.0.1)                  
#>  tidyr       * 1.1.0.9000 2020-07-15 [1] Github (tidyverse/tidyr@9c4f908)
#>  tidyselect    1.1.0      2020-05-11 [1] CRAN (R 4.0.0)                  
#>  tidyverse   * 1.3.0      2019-11-21 [1] CRAN (R 4.0.0)                  
#>  usethis       1.6.1      2020-04-29 [1] CRAN (R 4.0.0)                  
#>  utf8          1.1.4      2018-05-24 [1] CRAN (R 4.0.0)                  
#>  vctrs         0.3.2      2020-07-15 [1] CRAN (R 4.0.1)                  
#>  withr         2.2.0      2020-04-20 [1] CRAN (R 4.0.0)                  
#>  xfun          0.15       2020-06-21 [1] CRAN (R 4.0.1)                  
#>  xml2          1.3.2      2020-04-23 [1] CRAN (R 4.0.0)                  
#>  yaml          2.2.1      2020-02-01 [1] CRAN (R 4.0.0)                  
#> 
#> [1] /Library/Frameworks/R.framework/Versions/4.0/Resources/library
```

</details>

### LINKED ISSUE #998: hoist(): not possible to create a list_of column
It seems that it is currently not possible to create a `list_of` column.

``` r
library(tidyr)
df <- tibble(
  character = c("Toothless", "Dory"),
  metadata = list(
    list(colors = "black"),
    list(colors = "blue")
  )
)

# Goal: colors should be a `list_of<chr>`
df %>% 
  hoist(metadata,
    "colors",
    .ptype = list(colors = vctrs::list_of(.ptype = character()))
  )
#> Error: Can't convert <character> to <list_of<character>>.

# --> add explicit transformation function via `.transform`
df %>% 
  hoist(metadata,
    "colors",
    .ptype = list(colors = vctrs::list_of(.ptype = character())),
    .transform = list(
      colors = ~ {
        vctrs::list_of(!!!.x, .ptype = character())
      })
  )
#> Error: Can't simplfy 'colors'; contains a nested list

# --> simplify = FALSE doesn't combine to list_of
df %>% 
  hoist(metadata,
    "colors",
    .transform = list(
      colors = ~ {
        vctrs::list_of(!!!.x, .ptype = character())
      }),
    .simplify = FALSE
  )
#> # A tibble: 2 x 2
#>   character colors         
#>   <chr>     <list>         
#> 1 Toothless <list<chr> [1]>
#> 2 Dory      <list<chr> [1]>
```

<sup>Created on 2020-07-16 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### LINKED ISSUE #999: hoist(): .simplify = TRUE doesn't work for record style objects
``` r
library(tidyr)
x <- as.POSIXlt(Sys.time())
df <- tibble(
  metadata = list(
    list(time = x),
    list(time = x)
  )
)


df %>% 
  hoist(metadata,
    "time",
    .ptype = list(time = x)
  )
#> Error: Can't simplfy 'time'; contains a nested list
```

<sup>Created on 2020-07-16 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

It is actually explicitly checked in `tidyr:::simplify_col()` that each element must not be a list. For record style objects this check doesn't really make sense

```
simplify_col <- function(x, nm, ptype = list(), transform = list(), simplify = FALSE) {
   ...
  # Don't simplify lists of lists, because that typically indicates that
  # there might be multiple values.
  is_list <- map_lgl(x, is.list)
  if (any(is_list)) {
    if (is.null(ptype)) {
      return(x)
    } else {
      abort(glue("Can't simplfy '{nm}'; contains a nested list"))
    }
  }
  ...
}
```

### LINKED ISSUE #1000: Rethink simplify and ptype argument in hoist() and unnest_*()
After finding the issues `tidyr:::simplify_col()` (see #995, #996, #998, and #999) with the `simplify` and `ptype` argument I thought it might be worth to change the logic and simply get rid of the `simplify` argument. For example something like this might do the job

``` r
library(purrr)
library(vctrs)
library(rlang)

simplify_col <- function(x, nm, ptype = list(), transform = list()) {
  transform <- transform[[nm]]
  ptype <- ptype[[nm]]

  if (!is.null(transform)) {
    transform <- as_function(transform)
    x <- map(x, transform)
  }
  
  if (inherits(ptype, "AsIs")) {
    x
  } else if (is.null(ptype)) {
    tryCatch(
      vec_c(!!!x),
      vctrs_error_incompatible_type = function(e) x
    )
  } else {
    vec_c(!!!x, .ptype = ptype)
  }
}


# record style objects
x_rcrd <- rep(list(as.POSIXlt(Sys.time())), 2)
simplify_col(x_rcrd, nm = "rcrd", ptype = list(rcrd = x_rcrd[[1]]))
#> [1] "2020-07-16 09:31:21 UTC" "2020-07-16 09:31:21 UTC"
simplify_col(x_rcrd, nm = "rcrd", ptype = list(rcrd = I("")))
#> [[1]]
#> [1] "2020-07-16 09:31:21 UTC"
#> 
#> [[2]]
#> [1] "2020-07-16 09:31:21 UTC"

# list_of columns
x_list_of <- list("a", c("b", "c"))
simplify_col(
  x_list_of,
  nm = "list_of",
  ptype = list(list_of = list_of(.ptype = character())),
  transform = list(list_of = ~ list_of(.x, .ptype = character()))
)
#> <list_of<character>[2]>
#> [[1]]
#> [1] "a"
#> 
#> [[2]]
#> [1] "b" "c"
```

<sup>Created on 2020-07-16 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

This way:
* all prototypes work
* one got rid of an argument (and therefore the value of `simplify` cannot influence the action of `ptype` anymore)
* record style objects can be combined

In the case of `list_of` I don't really like casting every element to a `list_of` and then combining them afterwards while a simple `list_of(!!!x)` would be sufficient. Of course one could avoid this by giving control what should happen to the vector at the end (something like `transform_result`) but this would also make the user interface more complicated. For example something like

```r
simplify_col2 <- function(x, nm, ptype = list(), transform = list()) {
  transform <- transform[[nm]]
  ptype <- ptype[[nm]]

  if (!is.null(transform)) {
    transform <- as_function(transform)
    x <- map(x, transform)
  }
  
  if (is.function(ptype) || is_formula(ptype)) {
    ptype <- as_function(ptype)
    ptype(x)
  } else if (is.null(ptype)) {
    tryCatch(
      vec_c(!!!x),
      vctrs_error_incompatible_type = function(e) x
    )
  } else {
    vec_c(!!!x, .ptype = ptype)
  }
}
```

### LINKED ISSUE #1029: `unnest_longer()` should allow better handling of indices
1. The handling of indices in `unnest_longer()` is currently buggy

``` r
library(tidyr)
#> Warning: package 'tidyr' was built under R version 4.0.2
df <- tibble(
  x = 1:3,
  y = list(NULL, c(a = 1, b = 2), 4:5)
)

# with default arguments we get a `y_id` column
df %>% 
  unnest_longer(y)
#> # A tibble: 5 x 3
#>       x     y y_id 
#>   <int> <dbl> <chr>
#> 1     1    NA <NA> 
#> 2     2     1 a    
#> 3     2     2 b    
#> 4     3     4 <NA> 
#> 5     3     5 <NA>

# if we actually want to include the indices it doesn't work because not all elements of `y` are named
df %>% 
  unnest_longer(y, indices_include = TRUE)
#> Error: Can't combine `..1$y_id` <character> and `..3$y_id` <integer>.
df %>% 
  unnest_longer(y, indices_to = "y_id")
#> Error: Can't combine `..1$y_id` <character> and `..3$y_id` <integer>.
```

<sup>Created on 2020-08-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

<details>

<summary>Session info</summary>

``` r
devtools::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value                       
#>  version  R version 4.0.1 (2020-06-06)
#>  os       macOS Catalina 10.15.6      
#>  system   x86_64, darwin17.0          
#>  ui       X11                         
#>  language (EN)                        
#>  collate  en_US.UTF-8                 
#>  ctype    en_US.UTF-8                 
#>  tz       UTC                         
#>  date     2020-08-13                  
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version date       lib source        
#>  assertthat    0.2.1   2019-03-21 [1] CRAN (R 4.0.0)
#>  backports     1.1.8   2020-06-17 [1] CRAN (R 4.0.1)
#>  callr         3.4.3   2020-03-28 [1] CRAN (R 4.0.0)
#>  cli           2.0.2   2020-02-28 [1] CRAN (R 4.0.0)
#>  crayon        1.3.4   2017-09-16 [1] CRAN (R 4.0.0)
#>  desc          1.2.0   2018-05-01 [1] CRAN (R 4.0.0)
#>  devtools      2.3.1   2020-07-21 [1] CRAN (R 4.0.1)
#>  digest        0.6.25  2020-02-23 [1] CRAN (R 4.0.0)
#>  dplyr         1.0.1   2020-07-31 [1] CRAN (R 4.0.2)
#>  ellipsis      0.3.1   2020-05-15 [1] CRAN (R 4.0.0)
#>  evaluate      0.14    2019-05-28 [1] CRAN (R 4.0.0)
#>  fansi         0.4.1   2020-01-08 [1] CRAN (R 4.0.0)
#>  fs            1.5.0   2020-07-31 [1] CRAN (R 4.0.2)
#>  generics      0.0.2   2018-11-29 [1] CRAN (R 4.0.0)
#>  glue          1.4.1   2020-05-13 [1] CRAN (R 4.0.0)
#>  highr         0.8     2019-03-20 [1] CRAN (R 4.0.0)
#>  htmltools     0.5.0   2020-06-16 [1] CRAN (R 4.0.1)
#>  knitr         1.29    2020-06-23 [1] CRAN (R 4.0.1)
#>  lifecycle     0.2.0   2020-03-06 [1] CRAN (R 4.0.0)
#>  magrittr      1.5     2014-11-22 [1] CRAN (R 4.0.0)
#>  memoise       1.1.0   2017-04-21 [1] CRAN (R 4.0.0)
#>  pillar        1.4.6   2020-07-10 [1] CRAN (R 4.0.1)
#>  pkgbuild      1.1.0   2020-07-13 [1] CRAN (R 4.0.1)
#>  pkgconfig     2.0.3   2019-09-22 [1] CRAN (R 4.0.0)
#>  pkgload       1.1.0   2020-05-29 [1] CRAN (R 4.0.0)
#>  prettyunits   1.1.1   2020-01-24 [1] CRAN (R 4.0.0)
#>  processx      3.4.3   2020-07-05 [1] CRAN (R 4.0.0)
#>  ps            1.3.3   2020-05-08 [1] CRAN (R 4.0.2)
#>  purrr         0.3.4   2020-04-17 [1] CRAN (R 4.0.1)
#>  R6            2.4.1   2019-11-12 [1] CRAN (R 4.0.0)
#>  remotes       2.2.0   2020-07-21 [1] CRAN (R 4.0.1)
#>  rlang         0.4.7   2020-07-09 [1] CRAN (R 4.0.1)
#>  rmarkdown     2.3     2020-06-18 [1] CRAN (R 4.0.1)
#>  rprojroot     1.3-2   2018-01-03 [1] CRAN (R 4.0.0)
#>  sessioninfo   1.1.1   2018-11-05 [1] CRAN (R 4.0.0)
#>  stringi       1.4.6   2020-02-17 [1] CRAN (R 4.0.0)
#>  stringr       1.4.0   2019-02-10 [1] CRAN (R 4.0.0)
#>  testthat      2.3.2   2020-03-02 [1] CRAN (R 4.0.0)
#>  tibble        3.0.3   2020-07-10 [1] CRAN (R 4.0.1)
#>  tidyr       * 1.1.1   2020-07-31 [1] CRAN (R 4.0.2)
#>  tidyselect    1.1.0   2020-05-11 [1] CRAN (R 4.0.0)
#>  usethis       1.6.1   2020-04-29 [1] CRAN (R 4.0.0)
#>  utf8          1.1.4   2018-05-24 [1] CRAN (R 4.0.0)
#>  vctrs         0.3.2   2020-07-15 [1] CRAN (R 4.0.1)
#>  withr         2.2.0   2020-04-20 [1] CRAN (R 4.0.0)
#>  xfun          0.16    2020-07-24 [1] CRAN (R 4.0.2)
#>  yaml          2.2.1   2020-02-01 [1] CRAN (R 4.0.0)
#> 
#> [1] /Library/Frameworks/R.framework/Versions/4.0/Resources/library
```

</details>

2. The handling also doesn't feel right to me.
* to get the positions as index I have to unname all elements in the list column before to be sure that I actually get the positions.
* if I actually want the names and some elements do not have names then this is probably an error (or can easily lead to issues further down)
* if it is valid that some elements are named while others are not I probably want to get both, the positions and the names. But I'm not sure whether this is actually a realistic case.

Maybe the following logic works better:
* get rid of `indices_include`
* add argument `names_to`: A string giving the name of column which will contain the inner names of the values. Defaults to `NULL` (i.e. don't add a names column). Errors if not all elements have inner names.
* `indices_to`: A string giving the name of column which will contain the position of the values. Defaults to `NULL` (i.e. don't add an index column).

### LINKED ISSUE #1034: unnest_longer() doesn't work with single row nested data frames
I'm sorry, I'm new to tidyr package and I would be grateful if someone could explain me this behaviour. So, I've noticed that unnest_longer() is not able to unnest single row nested data frames as unnest() do. Is it expected to behave like this?

Here is a small example:

```R
library(tidyr)
library(magrittr)

nested_a <- tibble(A = 'a', B = 'b', C = 'c')
test_a <- tibble( col = list(nested_a) )

test_a %>% tidyr::unnest_longer(col)

#> Error: Assigned data `map2(...)` must be compatible with existing data.
#> x Existing data has 1 row.
#> x Assigned data has 3 rows.
#> i Row updates require a list value. Do you need `list()` or `as.list()`?
#> Run `rlang::last_error()` to see where the error occurred.
```

It works with unnest():

```R
test_a %>% tidyr::unnest(col)

#> # A tibble: 1 x 3
#>   A     B     C    
#>   <chr> <chr> <chr>
#> 1 a     b     c  
```

```R
Also with nested data frames with 2+ rows:

nested_b <- tibble( A = c('a', 'a'), B = c('b', 'b'), C = c('c', 'c') )
test_b <- tibble( col = list(nested_b))

test_b %<>% tidyr::unnest_longer(col)
test_b %>% unpack(col)

#> # A tibble: 2 x 3
#>   A     B     C    
#>   <chr> <chr> <chr>
#> 1 a     b     c    
#> 2 a     b     c 
```

### LINKED ISSUE #1044: Hoist and pluck return different values when extracting NULL list-elements
Hi,

Unless I am doing something wrong, it seems like `purrr::pluck()` and `tidyr::hoist()` return different values when subsetting a `NULL` list element. The reason I posted the issue here is because I think `pluck()` does the expected behaviour, meanwhile `hoist()` is doing some coercion. When `hoist`ing a `NULL` list-element it seems to be getting coerced to `NA`.
``` r
library(tidyverse)

foo <- function() stop("This won't run")

safe_foo <- safely(foo)

safe_foo()
#> $result
#> NULL
#> 
#> $error
#> <simpleError in .f(...): This won't run>

pluck(safe_foo(), "result")
#> NULL
pluck(safe_foo(), "error")
#> <simpleError in .f(...): This won't run>

tibble(a = 1:3) %>% 
  mutate(b = map(a, safe_foo)) %>% 
  hoist(b, "result", "error")
#> # A tibble: 3 x 3
#>       a result error     
#>   <int> <lgl>  <list>    
#> 1     1 NA     <smplErrr>
#> 2     2 NA     <smplErrr>
#> 3     3 NA     <smplErrr>
```

<sup>Created on 2020-09-09 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### LINKED ISSUE #1060: unnest_wider() and messages with unnamed lists
Suppose you have a list with unnamed elements in a data frame column and you `unnest_wider()` that column. 

``` r
library(tidyverse)

tibble(
  x = list(
    list(1,2,3),
    list(4,5,6),
    list(7,8,9)
  )
) %>% 
  unnest_wider(x)
#> New names:
#> * `` -> ...1
#> * `` -> ...2
#> * `` -> ...3
#> New names:
#> * `` -> ...1
#> * `` -> ...2
#> * `` -> ...3
#> New names:
#> * `` -> ...1
#> * `` -> ...2
#> * `` -> ...3
#> # A tibble: 3 x 3
#>    ...1  ...2  ...3
#>   <dbl> <dbl> <dbl>
#> 1     1     2     3
#> 2     4     5     6
#> 3     7     8     9
```

<sup>Created on 2020-11-09 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

The message about New names is repeated as many times as the number of rows, which seems undesirable. I think either hiding that message overall or showing it once would be preferable.

### LINKED ISSUE #1125: `unnest_wider()` drops column if it contains empty lists
Inconsistency in output format that came up when processing nested lists that might be empty:

``` r
library(tidyr)
df <- tibble(
  x = list(list(z = list(1)), list(z = list(2))), 
  y = 1:2
  )
# column z as expected
df %>% unnest_wider(x)
#> # A tibble: 2 x 2
#>   z              y
#>   <list>     <int>
#> 1 <list [1]>     1
#> 2 <list [1]>     2


df <- tibble(
  x = list(list(z = list()), list(z = list())), 
  y = 1:2
)
# column z not present
df %>% unnest_wider(x)
#> # A tibble: 2 x 1
#>       y
#>   <int>
#> 1     1
#> 2     2
```

<sup>Created on 2021-06-10 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

### LINKED ISSUE #1186: `unnest_wider()` fails with a list-col containing only `NULL`
``` r
library(tidyr)

# Unnest column with just list(NULL)
unnest_wider(tibble(x = list(NULL)), x)
#> Error in data[[col]][] <- map2(data[[col]], names(data[[col]]), simplify_col, : replacement has length zero
unnest_wider(tibble(x = list(NULL)), x, simplify = FALSE)
#> Error in data[[col]][] <- map2(data[[col]], names(data[[col]]), simplify_col, : replacement has length zero
```

Possibly some bad invariant in `[<-` from this bit `data[[col]][] <- map2`

### LINKED ISSUE #1187: `unnest_wider()` should respect the `ptype` of a `list_of()` in the 0-row case
``` r
library(tidyr)
library(vctrs)

df <- tibble(x = list_of(.ptype = tibble(a = 1, b = 1)))

# list-of type should be preserved
unnest_wider(df, x)
#> # A tibble: 0 × 0

df <- tibble(x = list_of(tibble(a = 1, b = 1)))

# should be a zero row version of this 
unnest_wider(df, x)
#> # A tibble: 1 × 2
#>       a     b
#>   <dbl> <dbl>
#> 1     1     1
```

<sup>Created on 2021-11-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### LINKED ISSUE #1188: `unnest_wider()` doesn't work with data frame columns
It isn't the intentional usage for this function, but it theoretically should work if we had a perfect implementation:

``` r
library(tidyr)

df <- tibble(a = 1:3, b = tibble(x = 1:3, y = 1:3))
df
#> # A tibble: 3 × 2
#>       a   b$x    $y
#>   <int> <int> <int>
#> 1     1     1     1
#> 2     2     2     2
#> 3     3     3     3

unnest_wider(df, b)
#> New names:
#> * `` -> ...1
#> * `` -> ...2
#> * `` -> ...3
#> New names:
#> * `` -> ...1
#> * `` -> ...2
#> * `` -> ...3
#> Error: Assigned data `map(data[[col]], vec_to_wide, col = col, names_sep = names_sep)` must be compatible with existing data.
#> x Existing data has 3 rows.
#> x Assigned data has 2 rows.
#> ℹ Only vectors of size 1 are recycled.

# Pretty sure it should look like:
df$b <- vctrs::vec_chop(df$b)
unnest_wider(df, b)
#> # A tibble: 3 × 3
#>       a     x     y
#>   <int> <int> <int>
#> 1     1     1     1
#> 2     2     2     2
#> 3     3     3     3

# Because that is sort of equivalent to what this does:
df$b <- 1:3
unnest_wider(df, b)
#> New names:
#> * `` -> ...1
#> New names:
#> * `` -> ...1
#> New names:
#> * `` -> ...1
#> # A tibble: 3 × 2
#>       a  ...1
#>   <int> <int>
#> 1     1     1
#> 2     2     2
#> 3     3     3
```

<sup>Created on 2021-11-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### LINKED ISSUE #1193: `unnest_longer()` can't unnest a `list(NULL)` column
``` r
library(tidyr)

df <- tibble(x = list(NULL))

unnest_longer(df, x)
#> Error in data[[col]][] <- map2(data[[col]], names(data[[col]]), simplify_col, : replacement has length zero
```

<sup>Created on 2021-11-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### LINKED ISSUE #1194: `unnest_longer()` issue with nested data frames - `indices_include` isn't utilized
``` r
library(tidyr)

df <- tibble(x = list(tibble(a = 1:2)))
df
#> # A tibble: 1 × 1
#>   x               
#>   <list>          
#> 1 <tibble [2 × 1]>

# Invariant looks right, should never add columns, only rows

# But why dont we get an index column?
unnest_longer(df, x, indices_include = TRUE)
#> # A tibble: 2 × 1
#>     x$a
#>   <int>
#> 1     1
#> 2     2
```

<sup>Created on 2021-11-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### LINKED ISSUE #1195: `unnest_longer()` issue with nested data frames - `values_to` isn't utilized
``` r
library(tidyr)

df <- tibble(x = list(tibble(a = 1:2)))
df
#> # A tibble: 1 × 1
#>   x               
#>   <list>          
#> 1 <tibble [2 × 1]>

# Why can't we override the column name?
unnest_longer(df, x, values_to = "col")
#> # A tibble: 2 × 1
#>     x$a
#>   <int>
#> 1     1
#> 2     2
```

<sup>Created on 2021-11-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### LINKED ISSUE #1196: `unnest_longer()` doesn't work with packed data frames
Again, from a purely theoretical standpoint it should work like unnest-longer-ing an integer column

``` r
library(tidyr)

df <- tibble(a = tibble(x = 1:2, y = 3:4, z = 5:6))
df
#> # A tibble: 2 × 1
#>     a$x    $y    $z
#>   <int> <int> <int>
#> 1     1     3     5
#> 2     2     4     6

unnest_longer(df, a)
#> Error: Assigned data `map(...)` must be compatible with existing data.
#> x Existing data has 2 rows.
#> x Assigned data has 3 rows.
#> ℹ Only vectors of size 1 are recycled.

# Should be similar to this (a no-op for the packed version, but could respect `values_to`)
df$a <- vctrs::vec_chop(df$a)
df
#> # A tibble: 2 × 1
#>   a               
#>   <list>          
#> 1 <tibble [1 × 3]>
#> 2 <tibble [1 × 3]>

unnest_longer(df, a)
#> # A tibble: 2 × 1
#>     a$x    $y    $z
#>   <int> <int> <int>
#> 1     1     3     5
#> 2     2     4     6

# Because it has the same invariants as this
# (i.e. data frame is a vector of rows, like this is a vector of integers)
df$a <- 1:2
df
#> # A tibble: 2 × 1
#>       a
#>   <int>
#> 1     1
#> 2     2

unnest_longer(df, a)
#> # A tibble: 2 × 1
#>       a
#>   <int>
#> 1     1
#> 2     2
```

<sup>Created on 2021-11-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### LINKED ISSUE #1198: What is the real purpose of `values_to` in `unnest_longer()`?
- Is it _just_ for renaming the column once it is unnested? Is that very useful?

- Would this be difficult to support if we could unnest multiple columns?

- Would we ever want to retain the original column if `values_to` was also used?

``` r
library(tidyr)

df <- tibble(
  films = list(
    c(
      "How to Train Your Dragon",
      "How to Train Your Dragon 2",
      "How to Train Your Dragon: The Hidden World"
    ),
    c(
      "Finding Nemo", 
      "Finding Dory"
    )
  )
)

df
#> # A tibble: 2 × 1
#>   films    
#>   <list>   
#> 1 <chr [3]>
#> 2 <chr [2]>

# Overwrites existing column
unnest_longer(df, films)
#> # A tibble: 5 × 1
#>   films                                     
#>   <chr>                                     
#> 1 How to Train Your Dragon                  
#> 2 How to Train Your Dragon 2                
#> 3 How to Train Your Dragon: The Hidden World
#> 4 Finding Nemo                              
#> 5 Finding Dory

# We don't keep the original column (would we ever want to?)
unnest_longer(df, films, values_to = "unnested")
#> # A tibble: 5 × 1
#>   unnested                                  
#>   <chr>                                     
#> 1 How to Train Your Dragon                  
#> 2 How to Train Your Dragon 2                
#> 3 How to Train Your Dragon: The Hidden World
#> 4 Finding Nemo                              
#> 5 Finding Dory
```

<sup>Created on 2021-11-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### LINKED ISSUE #1199: `unnest_longer()` drops unnested column if it is empty
Which goes against its invariant that it will never change the number of columns

``` r
library(tidyr)

# Should return 1 empty integer col
df <- tibble(x = integer())
unnest_longer(df, x)
#> # A tibble: 0 × 0

# Should return 1 empty logical col (really unspecified but we can't control that)
df <- tibble(x = list())
unnest_longer(df, x)
#> # A tibble: 0 × 0

# Should return 1 empty integer col
df <- tibble(x = vctrs::list_of(.ptype = integer()))
unnest_longer(df, x)
#> # A tibble: 0 × 0
```

<sup>Created on 2021-11-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### LINKED ISSUE #1201: `unnest_longer()` `indices_to` should be derived from `values_to`, not `col`
So that when you rename with `values_to`, you get an intelligent id column

``` r
library(tidyr)

df <- tibble(
  x = 1:2,
  y = list(c(a = 1, b = 2), c(a = 10, b = 11, c = 12))
)

unnest_longer(df, y, values_to = "xx")
#> # A tibble: 5 × 3
#>       x    xx y_id 
#>   <int> <dbl> <chr>
#> 1     1     1 a    
#> 2     1     2 b    
#> 3     2    10 a    
#> 4     2    11 b    
#> 5     2    12 c
```

<sup>Created on 2021-11-05 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

DavisVaughan:
`unnest_longer()`:
Closes #1201
Closes #1199 
Closes #1198 
Closes #1196 
Closes #1195 
Closes #1194 
Closes #1193 
Closes #1034 
Closes #1029 
Closes #740 

`unnest_wider()`:
Closes #1188 
Closes #1187 
Closes #1186
Closes #1125 
Closes #1133 
Closes #1060 

`hoist()`:
Added test of current behavior for #1203 
Closes #1044 
Closes #1000 
Closes #999 
Closes #998 
Closes #996 

Other:
Part of #1185 

---

Affected revdeps:

- ffscrapr
  - Pre-emptive PR already merged https://github.com/ffverse/ffscrapr/pull/344

- wehoop
  - Lots of off-label usage of `unnest_wider()`. PR https://github.com/saiemgilani/wehoop/pull/14
  - Two tests explicitly check for a set of column names after all the unnesting is applied. Fixing the first part of https://github.com/tidyverse/tidyr/issues/1125#issuecomment-959802379 means that columns with all empty elements are now (correctly) retained, which means this set of column names should be updated. Luckily this test isn't tested on CRAN, so it should be easy. PR https://github.com/saiemgilani/wehoop/pull/15

Update: Both revdeps have already merged the PRs and have updated their packages on CRAN. So there are now no affected revdeps.

---

This PR completely rewrites the rectangling functions. They are now much more consistent in a variety of edge cases, and are often much faster than they used to be. I anticipate very little breaking changes, except in some extreme off-label usage situations. I will leave in-line comments, but I think this generally should be reviewed as if these were completely new functions.

Here is the most interesting performance difference I found so far, which comes from https://stackoverflow.com/questions/68897169/any-speedier-alternatives-to-tidyrunnest-longer-when-dealing-with-nested-nam. Mainly this comes from getting rid of `tibble()` in favor of `new_data_frame()`. I think #1133 also identified this. (It is also worth noting that in this case a simple `unnest()` is even faster. Like, in the ~200ms range. But it is nice to see this improvement).

```r
library(tidyr)

elt <- 1:5
col <- rep(list(elt), 200000L)

df <- tibble(col = col)
df
#> # A tibble: 200,000 × 1
#>    col      
#>    <list>   
#>  1 <int [5]>
#>  2 <int [5]>
#>  3 <int [5]>
#>  4 <int [5]>
#>  5 <int [5]>
#>  6 <int [5]>
#>  7 <int [5]>
#>  8 <int [5]>
#>  9 <int [5]>
#> 10 <int [5]>
#> # … with 199,990 more rows

# CRAN
bench::mark(unnest_longer(df, col))
#> # A tibble: 1 × 6
#>   expression                  min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>             <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 unnest_longer(df, col)    1.14m    1.14m    0.0146     292MB     2.09

# This PR
bench::mark(unnest_longer(df, col))
#> # A tibble: 1 × 6
#>   expression                  min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>             <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 unnest_longer(df, col)    2.45s    2.45s     0.408    39.6MB     4.90
```

### DIFF
--- a/R/rectangle.R
+++ b/R/rectangle.R
@@ -396,8 +396,8 @@ col_to_long <- function(col,
 
   if (!vec_is_list(col)) {
     ptype <- vec_ptype(col)
-    col <- tidyr_chop(col)
-    col <- tidyr_temporary_new_list_of(col, ptype = ptype)
+    col <- vec_chop(col)
+    col <- new_list_of(col, ptype = ptype)
   }
 
   info <- collect_indices_info(col, indices_include)
@@ -596,8 +596,8 @@ col_to_wide <- function(col, name, strict, names_sep) {
 
   if (!vec_is_list(col)) {
     ptype <- vec_ptype(col)
-    col <- tidyr_chop(col)
-    col <- tidyr_temporary_new_list_of(col, ptype = ptype)
+    col <- vec_chop(col)
+    col <- new_list_of(col, ptype = ptype)
   }
 
   # If we don't have a list_of, then a `NULL` `col_ptype` will get converted to
@@ -655,18 +655,26 @@ elt_to_wide <- function(x, name, strict, names_sep) {
     # Extremely special case for data.frames,
     # which we want to treat like lists where we know the type of each element
     x <- tidyr_new_list(x)
-    ptypes <- map(x, vec_ptype)
-    x <- tidyr_chop(x)
-    x <- map2(x, ptypes, new_list_of)
-  } else if (!strict && vec_is_list(x)) {
-    empty <- list_sizes(x) == 0L
-    if (any(empty)) {
-      # Can't use vec_assign(), see https://github.com/r-lib/vctrs/issues/1424
-      x[empty] <- list(NULL)
-    }
-    x <- tidyr_chop(x)
+    x <- map(x, list_of)
   } else {
-    x <- tidyr_chop(x)
+    if (!strict && vec_is_list(x)) {
+      empty <- list_sizes(x) == 0L
+      if (any(empty)) {
+        # Can't use vec_assign(), see https://github.com/r-lib/vctrs/issues/1424
+        x[empty] <- list(NULL)
+      }
+    }
+
+    names <- vec_names(x)
+
+    if (is.null(names)) {
+      x <- vec_chop(x)
+    } else {
+      # Promote names to column names
+      x <- vec_set_names(x, NULL)
+      x <- vec_chop(x)
+      x <- vec_set_names(x, names)
+    }
   }
 
   if (!is.null(names_sep)) {
--- a/R/rectangle.R
+++ b/R/rectangle.R
@@ -397,7 +397,7 @@ col_to_long <- function(col,
   if (!vec_is_list(col)) {
     ptype <- vec_ptype(col)
     col <- tidyr_chop(col)
-    col <- new_list_of(col, ptype = ptype)
+    col <- tidyr_temporary_new_list_of(col, ptype = ptype)
   }
 
   info <- collect_indices_info(col, indices_include)
@@ -435,7 +435,7 @@ col_to_long <- function(col,
   ptype <- vec_ptype_common(elt_ptype, !!!col)
   col <- vec_cast_common(!!!col, .to = ptype)
 
-  col <- new_list_of(col, ptype = ptype)
+  col <- tidyr_temporary_new_list_of(col, ptype = ptype)
 
   col
 }
@@ -597,7 +597,7 @@ col_to_wide <- function(col, name, strict, names_sep) {
   if (!vec_is_list(col)) {
     ptype <- vec_ptype(col)
     col <- tidyr_chop(col)
-    col <- new_list_of(col, ptype = ptype)
+    col <- tidyr_temporary_new_list_of(col, ptype = ptype)
   }
 
   # If we don't have a list_of, then a `NULL` `col_ptype` will get converted to
@@ -625,7 +625,7 @@ col_to_wide <- function(col, name, strict, names_sep) {
     out <- vec_cast_common(!!!out, .to = ptype)
   }
 
-  out <- new_list_of(out, ptype = ptype)
+  out <- tidyr_temporary_new_list_of(out, ptype = ptype)
 
   out
 }
@@ -803,7 +803,7 @@ col_simplify <- function(x,
   if (!is.null(ptype)) {
     x <- tidyr_new_list(x)
     x <- vec_cast_common(!!!x, .to = ptype)
-    x <- new_list_of(x, ptype = ptype)
+    x <- tidyr_temporary_new_list_of(x, ptype = ptype)
   }
 
   if (!simplify) {
@@ -851,4 +851,23 @@ col_simplify <- function(x,
     vec_unchop(x_scalars, ptype = x_ptype),
     vctrs_error_incompatible_type = function(e) x
   )
+}
+
+tidyr_temporary_new_list_of <- function(x, ptype) {
+  # TODO: Remove me and replace with `new_list_of()` when
+  # https://github.com/r-lib/vctrs/pull/784 is merged / fixed.
+
+  # This is an unfortunate hack required because vctr types currently
+  # can't have `""` names, and that can easily come up when unnesting a
+  # mix of named and unnamed elements.
+
+  # Most of the time, in the places we use this helper the names
+  # weren't needed at all, so it is safe to remove them. Any place
+  # where this isn't the case has a test and a linked open issue.
+  names <- names(x)
+  if (!is.null(names) && any(names == "")) {
+    x <- unname(x)
+  }
+
+  new_list_of(x, ptype = ptype)
 }
--- a/R/rectangle.R
+++ b/R/rectangle.R
@@ -742,9 +742,23 @@ df_simplify <- function(x,
   if (!vec_is_list(ptype)) {
     abort("`ptype` must be a list.")
   }
-  if (!vec_is_list(ptype)) {
+  if (length(ptype) > 0L && !is_named(ptype)) {
+    abort("All elements of `ptype` must be named.")
+  }
+  if (vec_duplicate_any(names(ptype))) {
+    abort("The names of `ptype` must be unique.")
+  }
+
+  if (!vec_is_list(transform)) {
     abort("`transform` must be a list.")
   }
+  if (length(transform) > 0L && !is_named(transform)) {
+    abort("All elements of `transform` must be named.")
+  }
+  if (vec_duplicate_any(names(transform))) {
+    abort("The names of `transform` must be unique.")
+  }
+
   if (!is_bool(simplify)) {
     abort("`simplify` must be a single `TRUE` or `FALSE`.")
   }
--- a/R/rectangle.R
+++ b/R/rectangle.R
@@ -43,7 +43,9 @@
 #'   For `hoist()` and `unnest_auto()`, this must identify a single column.
 #'
 #'   For `unnest_wider()` and `unnest_longer()`, you can use tidyselect to
-#'   select multiple columns to unnest simultaneously.
+#'   select multiple columns to unnest simultaneously. When using
+#'   `unnest_longer()` with multiple columns, values across columns that
+#'   originated from the same row are recycled to a common size.
 #' @param ... Components of `.col` to turn into columns in the form
 #'   `col_name = "pluck_specification"`. You can pluck by name with a character
 #'   vector, by position with an integer vector, or with a combination of the
--- a/R/rectangle.R
+++ b/R/rectangle.R
@@ -60,8 +60,8 @@
 #'   check that each element has the type you expect when simplifying.
 #'
 #'   If a `ptype` has been specified, but `simplify = FALSE` or simplification
-#'   isn't possible, then a list-of column will be returned and each element
-#'   will have type `ptype`.
+#'   isn't possible, then a [list-of][vctrs::list_of()] column will be returned
+#'   and each element will have type `ptype`.
 #' @param .transform,transform Optionally, a named list of transformation
 #'   functions applied to each component. Use this function if you want
 #'   to transform or parse individual elements as they are extracted.
--- a/R/rectangle.R
+++ b/R/rectangle.R
@@ -12,6 +12,10 @@
 #'
 #' Learn more in `vignette("rectangle")`.
 #'
+#' @details
+#' When both `ptype` and `transform` are supplied, the `transform` is applied
+#' before the `ptype`.
+#'
 #' @section Unnest variants:
 #'
 #' The three `unnest()` functions differ in how they change the shape of the
@@ -19,25 +23,31 @@
 #'
 #' * `unnest_wider()` preserves the rows, but changes the columns.
 #' * `unnest_longer()` preserves the columns, but changes the rows
-#' * `unnest()` can change both rows and columns.
+#' * [unnest()] can change both rows and columns.
 #'
 #' These principles guide their behaviour when they are called with a
 #' non-primary data type. For example, if you `unnest_wider()` a list of data
 #' frames, the number of rows must be preserved, so each column is turned into
 #' a list column of length one. Or if you `unnest_longer()` a list of data
-#' frame, the number of columns must be preserved so it creates a packed
+#' frames, the number of columns must be preserved so it creates a packed
 #' column. I'm not sure how if these behaviours are useful in practice, but
 #' they are theoretically pleasing.
 #'
 #' @section `unnest_auto()` heuristics:
 #' `unnest_auto()` inspects the inner names of the list-col:
-#' * If all elements are unnamed, it uses `unnest_longer()`
+#' * If all elements are unnamed, it uses
+#'   `unnest_longer(indices_include = FALSE)`.
 #' * If all elements are named, and there's at least one name in
-#'   common acros all components, it uses `unnest_wider()`
+#'   common across all components, it uses `unnest_wider()`.
 #' * Otherwise, it falls back to `unnest_longer(indices_include = TRUE)`.
 #'
 #' @param .data,data A data frame.
 #' @param .col,col List-column to extract components from.
+#'
+#'   For `hoist()` and `unnest_auto()`, this must identify a single column.
+#'
+#'   For `unnest_wider()` and `unnest_longer()`, you can use tidyselect to
+#'   select multiple columns to unnest simultaneously.
 #' @param ... Components of `.col` to turn into columns in the form
 #'   `col_name = "pluck_specification"`. You can pluck by name with a character
 #'   vector, by position with an integer vector, or with a combination of the
@@ -48,15 +58,21 @@
 #'   single string you can choose to omit the name, i.e. `hoist(df, col, "x")`
 #'   is short-hand for `hoist(df, col, x = "x")`.
 #' @param .simplify,simplify If `TRUE`, will attempt to simplify lists of
-#'   length-1 vectors to an atomic vector
-#' @param .ptype,ptype Optionally, a named list of prototypes declaring the desired
-#'   output type of each component. Use this argument if you want to check each
-#'   element has the types you expect when simplifying.
-#' @param .transform,transform Optionally, a named list of transformation functions
-#'   applied to each component. Use this function if you want transform or
-#'   parse individual elements as they are hoisted.
+#'   length-1 vectors to an atomic vector.
+#' @param .ptype,ptype Optionally, a named list of prototypes declaring the
+#'   desired output type of each component. Use this argument if you want to
+#'   check that each element has the type you expect when simplifying.
+#'
+#'   If a `ptype` has been specified, but `simplify = FALSE` or simplification
+#'   isn't possible, then a list-of column will be returned and each element
+#'   will have type `ptype`.
+#' @param .transform,transform Optionally, a named list of transformation
+#'   functions applied to each component. Use this function if you want
+#'   to transform or parse individual elements as they are extracted.
 #' @param .remove If `TRUE`, the default, will remove extracted components
-#'   from `.col`. This ensures that each value lives only in one place.
+#'   from `.col`. This ensures that each value lives only in one place. If all
+#'   components are removed from `.col`, then `.col` will be removed from the
+#'   result entirely.
 #' @examples
 #' df <- tibble(
 #'   character = c("Toothless", "Dory"),
@@ -113,13 +129,52 @@
 #' df %>% unnest_wider(y)
 #' df %>% unnest_longer(y)
 #'
+#' # Both unnest_wider() and unnest_longer() allow you to unnest multiple
+#' # columns at once. This is particularly useful with unnest_longer(), where
+#' # unnesting sequentially would generate a cartesian product of the rows.
+#' df <- tibble(
+#'   x = 1:2,
+#'   y = list(1:2, 3:4),
+#'   z = list(5:6, 7:8)
+#' )
+#' unnest_longer(df, c(y, z))
+#' unnest_longer(unnest_longer(df, y), z)
+#'
+#' # With JSON, it is common for empty elements to be represented by `list()`
+#' # rather then their typed equivalent, like `integer()`
+#' json <- list(
+#'   list(x = 1:2, y = 1:2),
+#'   list(x = list(), y = 3:4),
+#'   list(x = 3L, y = list())
+#' )
+#' df <- tibble(json = json)
+#'
+#' # The defaults of `unnest_wider()` treat empty types (like `list()`) as `NULL`.
+#' # This chains nicely into `unnest_longer()`.
+#' wide <- unnest_wider(df, json)
+#' wide
+#'
+#' unnest_longer(wide, c(x, y))
+#'
+#' # To instead enforce strict vctrs typing rules, use `strict`
+#' wide_strict <- unnest_wider(df, json, strict = TRUE)
+#' wide_strict
+#'
+#' try(unnest_longer(wide_strict, c(x, y)))
 #' @export hoist
-hoist <- function(.data, .col, ..., .remove = TRUE, .simplify = TRUE, .ptype = list(), .transform = list()) {
+hoist <- function(.data,
+                  .col,
+                  ...,
+                  .remove = TRUE,
+                  .simplify = TRUE,
+                  .ptype = list(),
+                  .transform = list()) {
   check_present(.col)
-  .col <- tidyselect::vars_pull(names(.data), !!enquo(.col))
+  .col <- tidyselect::vars_pull(names(.data), {{.col}})
+
   x <- .data[[.col]]
-  if (!is.list(x)) {
-    abort("`col` must be a list-column")
+  if (!vec_is_list(x)) {
+    abort("`.col` must identify a list-column.")
   }
 
   pluckers <- check_pluckers(...)
@@ -128,20 +183,20 @@ hoist <- function(.data, .col, ..., .remove = TRUE, .simplify = TRUE, .ptype = l
   # the lookup has a large performance impact:
   # https://github.com/tidyverse/tidyr/issues/1001
   pluck <- purrr::pluck
-  new_cols <- map(pluckers, function(idx) {
+  cols <- map(pluckers, function(idx) {
     map(x, ~ pluck(.x, !!!idx))
   })
-  new_cols <- map2(
-    new_cols,
-    names(new_cols),
-    simplify_col,
-    simplify = .simplify,
+  cols <- new_data_frame(cols, n = vec_size(.data))
+
+  cols <- df_simplify(
+    cols,
     ptype = .ptype,
-    transform = .transform
+    transform = .transform,
+    simplify = .simplify
   )
 
   # Place new columns before old column
-  out <- append_df(.data, new_cols, after = match(.col, names(.data)) - 1L)
+  out <- append_df(.data, cols, after = match(.col, names(.data)) - 1L)
 
   if (!.remove) {
     return(out)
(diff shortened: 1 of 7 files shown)


R/rectangle.R
```diff
@@ -167,104 +222,461 @@ hoist <- function(.data, .col, ..., .remove = TRUE, .simplify = TRUE, .ptype = l
...
+  is_string <- map_lgl(pluckers, ~is.character(.x) && length(.x) == 1)
   auto_name <- names2(pluckers) == "" & is_string
 
   if (any(auto_name)) {
     names(pluckers)[auto_name] <- unlist(pluckers[auto_name])
   }
 
-  if (!is_named(pluckers)) {
-    stop("All elements of `...` must be named", call. = FALSE)
+  if (length(pluckers) > 0 && !is_named(pluckers)) {
```
DavisVaughan, line 232: Would be `is_named2()` if we had dev rlang

R/rectangle.R
```diff
@@ -167,104 +222,461 @@ hoist <- function(.data, .col, ..., .remove = TRUE, .simplify = TRUE, .ptype = l
...
 
   if (vec_duplicate_any(names(pluckers))) {
-    abort("The names of `...` must be unique")
+    abort("The names of `...` must be unique.")
   }
 
+  # Standardize all pluckers to lists for splicing into `pluck()`
+  # and for easier handling in `strike()`
+  is_not_list <- !map_lgl(pluckers, vec_is_list)
+  pluckers[is_not_list] <- map(pluckers[is_not_list], vec_chop)
```
DavisVaughan, line 243: It is generally easier to transform a mix of plucker types like: `hoist(df, col, x = "x", y = c(1, 2), z = list("foo", "bar"))` into their equivalent list type, like `hoist(df, col, x = list("x"), y = list(1, 2), z = list("foo", "bar"))`

R/rectangle.R
```diff
@@ -167,104 +222,461 @@ hoist <- function(.data, .col, ..., .remove = TRUE, .simplify = TRUE, .ptype = l
...
                           ptype = list(),
-                          transform = list()
-                          ) {
-
+                          transform = list()) {
   check_present(col)
-  col <- tidyselect::vars_pull(names(data), !!enquo(col))
+  # TODO: Use `allow_rename = FALSE`.
+  # Requires https://github.com/r-lib/tidyselect/issues/225.
+  cols <- tidyselect::eval_select(enquo(col), data)
```
DavisVaughan, line 320: For now we just rely on the assumption that the user didn't do any renaming, which should be okay for now

R/rectangle.R
```diff
@@ -167,104 +222,461 @@ hoist <- function(.data, .col, ..., .remove = TRUE, .simplify = TRUE, .ptype = l
...
+                        indices_to,
+                        indices_include) {
+  if (is.null(col)) {
+    abort(glue("Invalid `NULL` column detected for column `{name}`."))
   }
 
-  data[[col]] <- map(
-    data[[col]], vec_to_long,
-    col = col,
+  if (!vec_is_list(col)) {
```
DavisVaughan, line 394: General idea is to get the column into a list or list-of form as fast as possible, no matter the original type

R/rectangle.R
```diff
@@ -167,104 +222,461 @@ hoist <- function(.data, .col, ..., .remove = TRUE, .simplify = TRUE, .ptype = l
...
   )
+  elt_ptype <- vec_ptype(elt_ptype)
+
+  # Avoid expensive dispatch from `[[.list_of`, and allow for `[[<-`
+  col <- tidyr_new_list(col)
+
+  for (i in seq_along(col)) {
+    col[[i]] <- elt_to_long(
+      x = col[[i]],
+      index = indices[[i]],
```
DavisVaughan, line 424: Can't use `map2()` here because `indices` might be `NULL`, but `NULL[[i]]` is valid and just returns `NULL` so the for loop form worked well

R/rectangle.R
```diff
@@ -167,104 +222,461 @@ hoist <- function(.data, .col, ..., .remove = TRUE, .simplify = TRUE, .ptype = l
...
+  out <- tidyr_new_list(col)
+  out <- map(out, elt_to_wide, name = name, strict = strict, names_sep = names_sep)
+
+  # In the sole case of a list_of<data_frame>, we can be sure that the
+  # elements of `out` will all be of the same type. Otherwise,
+  # - If `col` isn't a list-of, we don't know the element type.
+  # - If `col` is a list-of but not one with a data frame ptype, then we
+  #   don't actually know if all elements have the same ptype, because the
+  #   number of resulting columns per element depends on that element's size.
+  has_identical_elements <- is_list_of(col) && is.data.frame(col_ptype)
```
DavisVaughan, line 616: Mainly we get a performance benefit from this

R/rectangle.R
```diff
@@ -167,104 +222,461 @@ hoist <- function(.data, .col, ..., .remove = TRUE, .simplify = TRUE, .ptype = l
...
+    abort(glue("Column `{name}` must contain a list of vectors."))
+  }
+
+  if (is.data.frame(x)) {
+    # Extremely special case for data.frames,
+    # which we want to treat like lists where we know the type of each element
+    x <- tidyr_new_list(x)
+    ptypes <- map(x, vec_ptype)
+    x <- tidyr_chop(x)
+    x <- map2(x, ptypes, new_list_of)
```
DavisVaughan, line 657: I actually don't think the data frame behavior of `unnest_wider()` completely follows the semantics that I would expect. But what you have implemented is probably the most practically useful thing for it to do.

Here is what I would have expected (yes, it is weird):

``` r
library(tibble)

df <- tibble(x = list(tibble(a = 1:2, b = 2:3, c = 3:4)))
df
#> # A tibble: 1 × 1
#>   x               
#>   <list>          
#> 1 <tibble [2 × 3]>

df$x
#> [[1]]
#> # A tibble: 2 × 3
#>       a     b     c
#>   <int> <int> <int>
#> 1     1     2     3
#> 2     2     3     4

# unnest_wider(df, x)
# - The size of each element of `df$x` determines the number of columns.
#   2 rows (size) = 2 new columns
# - The `vec_names()` of the element determine the column names, but tibbles
#   don't use rownames...
tibble(...1 = df$x[[1]][1,], ...2 = df$x[[1]][2,])
#> # A tibble: 1 × 2
#>   ...1$a    $b    $c ...2$a    $b    $c
#>    <int> <int> <int>  <int> <int> <int>
#> 1      1     2     3      2     3     4

# This is more useful if you had a data frame with row names
df <- tibble(x = list(data.frame(a = 1:2, b = 2:3, c = 3:4, row.names = c("r1", "r2"))))
df$x
#> [[1]]
#>    a b c
#> r1 1 2 3
#> r2 2 3 4

# Then you'd get:
# unnest_wider(df, x)
tibble(r1 = df$x[[1]][1,], r2 = df$x[[1]][2,])
#> # A tibble: 1 × 2
#>    r1$a    $b    $c  r2$a    $b    $c
#>   <int> <int> <int> <int> <int> <int>
#> 1     1     2     3     2     3     4
```

<sup>Created on 2021-11-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/rectangle.R
```diff
@@ -317,152 +729,123 @@ guess_dir <- function(x, col) {
...
-  if (any(!is_vec)) {
-    if (is.null(ptype)) {
-      return(x)
-    } else {
-      abort(glue("Can't simplfy '{nm}'; contains a non-vector"))
-    }
+  if (!is.null(ptype)) {
+    x <- tidyr_new_list(x)
+    x <- vec_cast_common(!!!x, .to = ptype)
+    x <- new_list_of(x, ptype = ptype)
```
DavisVaughan, line 803: I think this is pretty slick. It means that if you can't simplify (or choose not to), then you can create a list-of of the non-simplified results by supplying a ptype for that component. This was requested in https://github.com/tidyverse/tidyr/issues/998

``` r
library(tidyr)

df <- tibble(
  x = list(
    list(a = 1:2),
    list(a = 1)
  )
)

df
#> # A tibble: 2 × 1
#>   x               
#>   <list>          
#> 1 <named list [1]>
#> 2 <named list [1]>

unnest_wider(df, x)
#> # A tibble: 2 × 1
#>   a        
#>   <list>   
#> 1 <int [2]>
#> 2 <dbl [1]>

unnest_wider(df, x, ptype = list(a = integer()))
#> # A tibble: 2 × 1
#>             a
#>   <list<int>>
#> 1         [2]
#> 2         [1]
```

<sup>Created on 2021-11-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/rectangle.R
```diff
@@ -317,152 +729,123 @@ guess_dir <- function(x, col) {
...
   }
 
-  n <- map_int(x, vec_size)
-  if (!all(n %in% c(0, 1))) {
-    if (is.null(ptype)) {
-      return(x)
-    } else {
-      abort(glue("Can't simplfy '{nm}'; elements have length > 1"))
-    }
+  if (!simplify) {
```
DavisVaughan, line 806: Note that we now apply the ptype and transform before returning if we don't want to simplify. We also apply the transform _then_ the ptype if they both are given, which I'm fairly confident is correct.

R/rectangle.R
```diff
@@ -317,152 +729,123 @@ guess_dir <- function(x, col) {
...
-        !!values_to := x,
-        !!indices_to := index(x)
-      )
-    } else {
-      tibble(!!values_to := x)
-    }
-  } else {
-    stop("Input must be list of vectors", call. = FALSE)
+  # Ensure empty elements are filled in with their correct size 1 equivalent
+  info <- list_init_empty(x, null = TRUE, typed = TRUE)
```
DavisVaughan, line 831: A more correct version of the previous `x[n == 0] <- list(NA)` behavior

R/rectangle.R
```diff
@@ -317,152 +729,123 @@ guess_dir <- function(x, col) {
...
-}
 
-index <- function(x) {
-  names(x) %||% seq_along(x)
+  x_scalars <- info$x
+  x_ptype <- list_of_ptype(x)
+
+  # Assume that if combining fails, then we want to return the object
+  # after the `ptype` and `transform` have been applied, but before the
+  # empty element filling was applied
```
DavisVaughan, line 846: i.e. compare CRAN against dev here:

``` r
library(tidyr)

df <- tibble(
  x = list(
    list(a = NULL),
    list(a = 1L),
    list(a = "x")
  )
)

df
#> # A tibble: 3 × 1
#>   x               
#>   <list>          
#> 1 <named list [1]>
#> 2 <named list [1]>
#> 3 <named list [1]>

# CRAN behavior (the NULL is replaced by NA)
unnest_wider(df, x)
#> # A tibble: 3 × 1
#>   a        
#>   <list>   
#> 1 <lgl [1]>
#> 2 <int [1]>
#> 3 <chr [1]>
unnest_wider(df, x)$a[[1]]
#> [1] NA

# Dev behavior
unnest_wider(df, x)
#> # A tibble: 3 × 1
#>   a        
#>   <list>   
#> 1 <NULL>   
#> 2 <int [1]>
#> 3 <chr [1]>
unnest_wider(df, x)$a[[1]]
#> NULL
```

R/rectangle.R
```diff
@@ -48,15 +58,21 @@
...
-#' @param .transform,transform Optionally, a named list of transformation functions
-#'   applied to each component. Use this function if you want transform or
-#'   parse individual elements as they are hoisted.
+#'   length-1 vectors to an atomic vector.
+#' @param .ptype,ptype Optionally, a named list of prototypes declaring the
+#'   desired output type of each component. Use this argument if you want to
+#'   check that each element has the type you expect when simplifying.
+#'
+#'   If a `ptype` has been specified, but `simplify = FALSE` or simplification
+#'   isn't possible, then a list-of column will be returned and each element
```
hadley, line 67: Link to `list_of`?

R/rectangle.R
```diff
@@ -12,32 +12,42 @@
...
+#'   common across all components, it uses `unnest_wider()`.
 #' * Otherwise, it falls back to `unnest_longer(indices_include = TRUE)`.
 #'
 #' @param .data,data A data frame.
 #' @param .col,col List-column to extract components from.
+#'
+#'   For `hoist()` and `unnest_auto()`, this must identify a single column.
+#'
+#'   For `unnest_wider()` and `unnest_longer()`, you can use tidyselect to
+#'   select multiple columns to unnest simultaneously.
```
hadley, line 50: Mention recycling for `unnest_longer()`?

R/rectangle.R
```diff
@@ -167,104 +222,461 @@ hoist <- function(.data, .col, ..., .remove = TRUE, .simplify = TRUE, .ptype = l
...
+
 #' @export
 #' @rdname hoist
-#' @param values_to Name of column to store vector values. Defaults to `col`.
-#' @param indices_to A string giving the name of column which will contain the
-#'   inner names or position (if not named) of the values. Defaults to `col`
-#'   with `_id` suffix
-#' @param indices_include Add an index column? Defaults to `TRUE` when `col`
-#'   has inner names.
+#' @param values_to A single string representing the column name to store the
```
hadley, line 291:
```suggestion
#' @param values_to A string giving the column name (or names) to store the
```

R/rectangle.R
```diff
@@ -167,104 +222,461 @@ hoist <- function(.data, .col, ..., .remove = TRUE, .simplify = TRUE, .ptype = l
...
+#'   unnested values in. If multiple columns are specified in `col`, this can
+#'   also be a glue string containing `"{col}"` to provide a template for the
+#'   column names. If `NULL`, defaults to `"{col}"`.
```
hadley, line 294: Maybe: The default, `NULL`, gives the output columns the same names as the input columns?

R/rectangle.R
```diff
@@ -167,104 +222,461 @@ hoist <- function(.data, .col, ..., .remove = TRUE, .simplify = TRUE, .ptype = l
...
+#' @param indices_to A string giving the name of the column which will contain
```
hadley, line 295:
```suggestion
#' @param indices_to A string giving the column name to store
```

To match `values_to`

R/rectangle.R
```diff
@@ -167,104 +222,461 @@ hoist <- function(.data, .col, ..., .remove = TRUE, .simplify = TRUE, .ptype = l
...
+#'   the inner names or positions (if not named) of the values. If multiple
+#'   columns are specified in `col`, this can also be a glue string containing
+#'   `"{col}"` to provide a template for the column names. If `NULL`, defaults
+#'   to `values_to` suffixed with `"_id"`.
```
hadley, line 299:
```suggestion
#'   to `"{col}_id"`.
```

R/rectangle.R
```diff
@@ -167,104 +222,461 @@ hoist <- function(.data, .col, ..., .remove = TRUE, .simplify = TRUE, .ptype = l
...
+#' @param indices_include A single logical value specifying whether or not to
+#'   add an index column. If any value has inner names, the index column will be
+#'   a character vector of those names, otherwise it will be an integer vector
+#'   of positions. If `NULL`, defaults to `TRUE` if any value has inner names
+#'   or if `indices_to` is provided.
+#'
+#'   If `indices_to` is provided, then `indices_include` can't be `FALSE`.
```
hadley, line 306:
```suggestion
#'   If `indices_to` is provided, then `indices_include` must not be `FALSE`.
```

R/rectangle.R
hadley: That is cool!

R/rectangle.R
hadley: Yeah, that looks much better.

hadley (commented):
First pass, just through the docs. I'll try and find the time to play with the code a bit too.

R/rectangle.R
DavisVaughan: I actually do mean `values_to` here, because it might be set to something other than `"{col}"`, like `"{col}2"`, making the `indices_to` columns default to `"{col}2_id"`

mgirlich:
So far I only had a look at `unnest_longer()` and `df_unchop()`. The code looks good to me and easy to read. Great work!

Two things I found:
1. Partial inner names cause an error

``` r
devtools::load_all("~/GitHub/tidyr/")
#> ℹ Loading tidyr
tibble(y = list(a = 1:2, 3:4)) %>% 
  unnest_longer(y, indices_include = FALSE)
#> Error: If any elements of `.data` are named, all must be named
```

2. The names of the `ptype` and `transform` argument should be checked with `vec_as_names(repair = "check_unique")`

``` r
devtools::load_all("~/GitHub/tidyr/")
#> ℹ Loading tidyr
# two ptypes for x
tibble(x = integer()) %>% 
  unnest_longer(
    c(x),
    ptype = list(x = double(), x = integer())
  )
#> # A tibble: 0 × 1
#> # … with 1 variable: x <dbl>

# unnamed list
tibble(x = integer()) %>% 
  unnest_longer(
    c(x),
    ptype = list(double())
  )
#> # A tibble: 0 × 1
#> # … with 1 variable: x <int>

# maybe warn/inform about unused ptypes?
tibble(x = integer()) %>% 
  unnest_longer(
    c(x),
    ptype = list(y = "a")
  )
#> # A tibble: 0 × 1
#> # … with 1 variable: x <int>
```

<sup>Created on 2021-11-10 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

mgirlich:
For `unnest_wider()`

1. Add a test with a data frame column and `names_sep`
```r
tibble(
  y = tibble(a = 1, b = 11)
) %>% 
  unnest_wider(y, names_sep = "_")
```

2. Add a test for a column that is not a list and not a data frame

```r
tibble(
  y = c(a = 1, b = 11)
) %>% 
  unnest_wider(y, names_sep = "_")
```

DavisVaughan:
Thanks @mgirlich, very helpful!

1. This is due to an annoying list-of bug that would be fixed by https://github.com/r-lib/vctrs/pull/784. I'll try and get that in the next vctrs release, it has affected me many times now. I've worked around it the best I can with `tidyr_temporary_new_list_of()`. There is one edge case that it affects, which I opened an issue for #1212.

2. I added more input validation here, but I think unused ptype/transform elements are going to have to go by silently. I don't think it is worth it to check for these.

3. I think your intuition is right. When converting a non-list to a list, there are two things you can do with the names. You can move them to the outer list (this is what we currently do, and what CRAN tidyr does but doesn't have tests for) or you can keep them on the inner elements after converting to a list. The latter aligns with your suggestion, and I now feel that that is correct because it uses more standard vctrs tooling (i.e. `vec_chop()` rather than `tidyr_chop()`) and it generally results in more expected output, even though these are non-primary data types. Below are the two options. I've switched over to the `equivalent2` approach you suggested.

``` r
library(tidyr)
library(vctrs)

# Say we have this:
df <- tibble(x = c(a = 1L, b = 2L))

# Names are moved to outer list (tidyr_chop())
equivalent1 <- tibble(x = list_of(a = 1L, b = 2L))

# Names are kept on inner elements (vec_chop())
equivalent2 <- tibble(x = list_of(c(a = 1L), c(b = 2L)))

unnest_wider(equivalent1, x)
#> New names:
#> * `` -> ...1
#> New names:
#> * `` -> ...1
#> # A tibble: 2 × 1
#>    ...1
#>   <int>
#> 1     1
#> 2     2

unnest_wider(equivalent2, x)
#> # A tibble: 2 × 2
#>       a     b
#>   <int> <int>
#> 1     1    NA
#> 2    NA     2

longer1 <- unnest_longer(equivalent1, x)
longer1
#> # A tibble: 2 × 1
#>       x
#>   <int>
#> 1     1
#> 2     2

longer1$x
#> [1] 1 2

longer2 <- unnest_longer(equivalent2, x)
longer2
#> # A tibble: 2 × 2
#>       x x_id 
#>   <int> <chr>
#> 1     1 a    
#> 2     2 b

longer2$x
#> a b 
#> 1 2
```

4. I've added those two tests, thanks!

mgirlich:
> I added more input validation here, but I think unused ptype/transform elements are going to have to go by silently. I don't think it is worth it to check for these.

Unused `ptype`/`transform` element might just be a typo. On the other hand in `unnest_wider()` unused `ptype` elements might be used for something different:

Currently, `unnest_wider()` isn't stable regarding the shape of the output:

``` r
library(tidyr)
df <- tibble(
  x = 1:2,
  y = list(
    list(a = 1),
    list(a = 2, b = 2)
  )
)

df %>% unnest_wider(y)
#> # A tibble: 2 × 3
#>       x     a     b
#>   <int> <dbl> <dbl>
#> 1     1     1    NA
#> 2     2     2     2

df %>% dplyr::slice(1) %>% unnest_wider(y)
#> # A tibble: 1 × 2
#>       x     a
#>   <int> <dbl>
#> 1     1     1
```

<sup>Created on 2021-11-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

The `ptype` argument can be used to ensure the types of the columns. Maybe one could also use it to ensure to also have the columns `x`, `a`, and `b` in the second case:

```r
df %>% dplyr::slice(1) %>% unnest_wider(y, ptype = list(a = double(), b = double()))
#> # A tibble: 1 × 2
#>       x     a
#>   <int> <dbl>
#> 1     1     1

# maybe this should output
#> # A tibble: 1 × 3
#>       x     a     b
#>   <int> <dbl> <dbl>
#> 1     1     1    NA
```

mgirlich:
I am still not entirely sure about the handling of indices as they are not type stable:

``` r
devtools::load_all("~/GitHub/tidyr/")
#> ℹ Loading tidyr
df <- tibble(x = list(c(1, 2), c(a = 3)))

df %>% 
  unnest_longer(x, indices_include = TRUE)
#> # A tibble: 3 × 2
#>       x x_id 
#>   <dbl> <chr>
#> 1     1 <NA> 
#> 2     2 <NA> 
#> 3     3 a

df %>% 
  dplyr::slice(1) %>% 
  unnest_longer(x, indices_include = TRUE)
#> # A tibble: 2 × 2
#>       x  x_id
#>   <dbl> <int>
#> 1     1     1
#> 2     2     2
```

<sup>Created on 2021-11-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I am not sure whether this is a common scenario but it feels like this goes against the idea of the shape and type predictability in the tidyverse.

DavisVaughan:
> Maybe one could also use it to ensure to also have the columns x, a, and b in the second case:

I think that would stray pretty far away from how `ptype` is used elsewhere in tidyr. But that could be the place of a "spec" object, like mentioned in https://github.com/tidyverse/tidyr/issues/835. I doubt we add that in this release, but that seems like the right place for this idea

> I am still not entirely sure about the handling of indices as they are not type stable:

Yea I think we are probably stuck with this at this point for backwards compatibility's sake. I agree that some alternative API might have been cleaner / more type stable (like `names_to` and `indices_to` you mentioned in https://github.com/tidyverse/tidyr/issues/1029), but at this point it is probably slightly better to be backwards compatible.

DavisVaughan:
@mgirlich would you like to review any further?

At this point I am ready to merge. No revdeps are affected now that wehoop and ffscrapr have already updated their packages on CRAN.

DavisVaughan:
> the handling of indices as they are not type stable

If we are really looking for some way to justify this, Hadley reminded me that one way to look at this is that an _index_ could be an integer or character vector, so it can be seen as a union type of integer+character. And since `x[<index>]` works correctly with integer or character vectors, it should be fine for most usage. Obviously not a perfect argument, but that + preserving backwards compatibility is enough to convince me to keep this as is.

mgirlich:
A couple more things then I think it is ready to merge

1. An early return to `col_to_wide()` for data frames would be great. Adding this

```r
  if (is.data.frame(col)) {
    out <- col

    if (!is.null(names_sep)) {
      outer <- name
      inner <- colnames(out)
      names(out) <- apply_names_sep(outer, inner, names_sep)
    }

    return(out)
  }
```

drastically improves the performance

```r
set.seed(1)

n <- 10e3
df <- tibble(
  x = tibble(a = sample(1:n))
)
bench::mark(unnest_wider(df, x))


# Current
# A tibble: 1 × 13
  expression               min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
  <bch:expr>          <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>
1 unnest_wider(df, x)    2.71s    2.71s     0.370     1.5GB     17.0     1    46

# With early return
# A tibble: 1 × 13
  expression               min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
  <bch:expr>          <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>
1 unnest_wider(df, x)   2.04ms   2.27ms      421.    1.93MB     31.4   134    10  
```

2. Similarly an early return to `col_to_long()` for non-list vectors would be great. This passes the tests and is way faster

```r
  if (!vec_is_list(col)) {
    out <- tibble("{name}" := col)

    if (is_false(indices_include)) {
      return(out)
    }

    indices <- vec_names(col)
    all_unnamed <- is.null(indices)

    if (is.null(indices_include) && all_unnamed) {
      return(out)
    }

    if (all_unnamed) {
      out[[indices_to]] <- vec_rep(1L, vec_size(col))
    } else {
      out[[indices_to]] <- indices
    }

    return(out)
  }
```

```r
set.seed(1)

n <- 10e3
df <- tibble(
  x = tibble(a = sample(1:n))
)
bench::mark(unnest_longer(df, x))

# Current
# A tibble: 1 × 13
  expression                min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
  <bch:expr>           <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>
1 unnest_longer(df, x)    212ms    266ms      3.76     3.6MB     9.40     2     5

# With early return
# A tibble: 1 × 13
  expression                min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
  <bch:expr>           <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>
1 unnest_longer(df, x)   2.74ms   3.15ms      316.    2.19MB     38.1    91    11
```

3. Can we get rid of the duplicated `New names` notification?
``` r
devtools::load_all("~/GitHub/tidyr/")
#> ℹ Loading tidyr

a <- set_names(1:3, c("x", NA, ""))
# get rid of duplicated new names notification?
tibble(a) %>% 
  unnest_wider(a)
#> New names:
#> * `` -> ...1
#> New names:
#> * `` -> ...1
#> # A tibble: 3 × 2
#>       x  ...1
#>   <int> <int>
#> 1     1    NA
#> 2    NA     2
#> 3    NA     3
```

<sup>Created on 2021-11-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

4. In `unnest_wider()` the `names_sep` doesn't work well with `NA`

``` r
devtools::load_all("~/GitHub/tidyr/")
#> ℹ Loading tidyr

a <- set_names(1:3, c("x", NA, ""))
# "a_NA" name?
tibble(a) %>% 
  unnest_wider(a, names_sep = "_")
#> # A tibble: 3 × 3
#>     a_x  a_NA    a_
#>   <int> <int> <int>
#> 1     1    NA    NA
#> 2    NA     2    NA
#> 3    NA    NA     3
```

<sup>Created on 2021-11-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

DavisVaughan:
1. This changes the behavior when `simplify = FALSE`. In general, I am a big fan of the fact that `col_to_wide()` now always returns a list-of column where each element is a tibble. I'd like to avoid changing that to special case non-primary data types. I did add a test for this now though, so thanks!

``` r
library(tidyr)

df <- tibble(x = tibble(a = 1:2))

# This PR
unnest_wider(df, x, simplify = FALSE)
#> # A tibble: 2 × 1
#>             a
#>   <list<int>>
#> 1         [1]
#> 2         [1]

# With the mentioned change
unnest_wider(df, x, simplify = FALSE)
#> # A tibble: 2 × 1
#>       a
#>   <int>
#> 1     1
#> 2     2
```

2. I can't find any fault with this one, I think it would work, but I am still a bit hesitant about branching too far away from the main code path for this special case. It is already a pretty complex operation, so I think I'd rather keep it simple here (It took me a long time to verify that it wouldn't change anything). I am optimistic that since this is not the intended use case for `unnest_longer()` anyways, that it won't affect very much actual code.

3. This is a vector special case of #1060. I think that we are going to have to keep these duplicated names messages with unnamed / partially named lists, as otherwise if we stopped after the first name repair message then you could be fooled into thinking you only have 1 element without names. Non-lists obviously get converted to lists with `vec_chop()`, so possibly we could repair names early on before using `vec_chop()`, but I'm not sure it is worth it. Again, with this being the non-primary data type, I'm mainly happy with it returning a consistent and theoretically correct result with the least amount of variation from the main code path as is possible.

4. Yea, I think this one is a bug, in particular, these two cases should return the same data frame structure (even if the column names are a little different)

``` r
library(tidyr)
library(rlang)

col <- list(
  set_names(1, "a"),
  set_names(1, NA_character_),
  set_names(1:2, c("", ""))
)

df <- tibble(col = col)

# Should be the same
unnest_wider(df, col)
#> New names:
#> * `` -> ...1
#> New names:
#> * `` -> ...1
#> * `` -> ...2
#> # A tibble: 3 × 3
#>       a  ...1  ...2
#>   <dbl> <dbl> <int>
#> 1     1    NA    NA
#> 2    NA     1    NA
#> 3    NA     1     2
unnest_wider(df, col, names_sep = "_")
#> New names:
#> * col_ -> col_...1
#> * col_ -> col_...2
#> # A tibble: 3 × 4
#>   col_a col_NA col_...1 col_...2
#>   <dbl>  <dbl>    <int>    <int>
#> 1     1     NA       NA       NA
#> 2    NA      1       NA       NA
#> 3    NA     NA        1        2
```

``` r
library(tidyr)
library(rlang)

col <- list(
  set_names(1, "a"),
  set_names(1, NA_character_),
  set_names(1, "")
)

df <- tibble(col = col)

# Should be the same
unnest_wider(df, col)
#> New names:
#> * `` -> ...1
#> New names:
#> * `` -> ...1
#> # A tibble: 3 × 2
#>       a  ...1
#>   <dbl> <dbl>
#> 1     1    NA
#> 2    NA     1
#> 3    NA     1
unnest_wider(df, col, names_sep = "_")
#> # A tibble: 3 × 3
#>   col_a col_NA  col_
#>   <dbl>  <dbl> <dbl>
#> 1     1     NA    NA
#> 2    NA      1    NA
#> 3    NA     NA     1
```

At the very least this means we should do minimal name repair before applying `names_sep`, but really I think we should do `"unique"` name repair to ensure that these return consistent results (i.e. should probably get `"...1"` for the first suffix).

``` r
library(tidyr)
library(rlang)

col1 <- list(set_names(1, ""))
df1 <- tibble(col = col1)

col2 <- list(set_names(1:2, c("", "")))
df2 <- tibble(col = col2)

unnest_wider(df1, col, names_sep = "_")
#> # A tibble: 1 × 1
#>    col_
#>   <dbl>
#> 1     1

unnest_wider(df2, col, names_sep = "_")
#> New names:
#> * col_ -> col_...1
#> * col_ -> col_...2
#> # A tibble: 1 × 2
#>   col_...1 col_...2
#>      <int>    <int>
#> 1        1        2
```

PR #1204, opened 2021-11-09
Avoid computing on `.data` in the `nest()` generic

### LINKED ISSUE #1134: `nest()` generic might error for unnamed dots
Before calling the method corresponding to the class the `nest()` generic tries to name any unnamed dots. This doesn't work for lazy tibbles:

``` r
library(dtplyr)
library(tidyr)

dt <- lazy_dt(tibble(x = c(1, 2, 1), y = c("a", "a", "b")))
dt %>% nest(y)
#> Error: `x` must be a vector, not a <dtplyr_step_first/dtplyr_step> object.
```

<sup>Created on 2021-07-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

The issue is the following line

```r
col_names <- names(tidyselect::eval_select(expr(c(...)), .data))
```

The question is how to handle this?

* Should each method implement this functionality itself? Probably not.
* Should `nest()` get the underlying tibble of `.data`? But what about predicates (column types)? See https://github.com/r-lib/tidyselect/issues/226.
* Should `eval_select()` be able to handle this?

### LINKED ISSUE #1175: Mixing named / unnamed dots in `nest()` can cause issues
``` r
library(tidyr)

# nest() generic should probably filter for only the unnamed inputs
nest(mtcars, foo = cyl, mpg)
#> Warning: All elements of `...` must be named.
#> Did you want `data = c(foo, mpg)`?
#> Error: Can't subset columns that don't exist.
#> x Column `foo` doesn't exist.

# woops, nested disp and mpg
nest(mtcars, disp = cyl, mpg)
#> Warning: All elements of `...` must be named.
#> Did you want `data = c(disp, mpg)`?
#> # A tibble: 32 × 10
#>      cyl    hp  drat    wt  qsec    vs    am  gear  carb data            
#>    <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <list>          
#>  1     6   110  3.9   2.62  16.5     0     1     4     4 <tibble [1 × 2]>
#>  2     6   110  3.9   2.88  17.0     0     1     4     4 <tibble [1 × 2]>
#>  3     4    93  3.85  2.32  18.6     1     1     4     1 <tibble [1 × 2]>
#>  4     6   110  3.08  3.22  19.4     1     0     3     1 <tibble [1 × 2]>
#>  5     8   175  3.15  3.44  17.0     0     0     3     2 <tibble [1 × 2]>
#>  6     6   105  2.76  3.46  20.2     1     0     3     1 <tibble [1 × 2]>
#>  7     8   245  3.21  3.57  15.8     0     0     3     4 <tibble [1 × 2]>
#>  8     4    62  3.69  3.19  20       1     0     4     2 <tibble [1 × 2]>
#>  9     4    95  3.92  3.15  22.9     1     0     4     2 <tibble [1 × 2]>
#> 10     6   123  3.92  3.44  18.3     1     0     4     4 <tibble [1 × 2]>
#> # … with 22 more rows
```

Or maybe it should just error if there is a mix, since this compatibility layer was really about catching the old `nest()` behavior where only unnamed dots were passed through

DavisVaughan:
Closes #1134 
Closes #1175 

We don't really need to do the full `eval_select()` call here, we just need to create a nice-ish label for the warning and rework the unnamed dots into a `c(...)` format to pass on to another call to `nest()`.

This should avoid any issues with dbplyr / dtplyr methods.

Also fixed #1175, where a mix of named/unnamed dots could cause some issues.

### DIFF
--- a/tests/testthat/test-nest.R
+++ b/tests/testthat/test-nest.R
@@ -315,6 +315,18 @@ test_that("warn about old style interface", {
   expect_named(out, c("x", "data"))
 })
 
+test_that("only warn about unnamed inputs (#1175)", {
+  df <- tibble(x = 1:3, y = 1:3, z = 1:3)
+  expect_warning(out <- nest(df, x, y, foo = z), "data = c(x, y)", fixed = TRUE)
+  expect_named(out, c("foo", "data"))
+})
+
+test_that("unnamed expressions are kept in the warning", {
+  df <- tibble(x = 1:3, z = 1:3)
+  expect_warning(out <- nest(df, x, starts_with("z")), 'data = c(x, starts_with("z"))', fixed = TRUE)
+  expect_named(out, "data")
+})
+
 test_that("can control output column name", {
   df <- tibble(x = c(1, 1, 1), y = 1:3)
   expect_warning(out <- nest(df, y, .key = "y"), "y = c(y)", fixed = TRUE)
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # tidyr (development version)
 
+* The `nest()` generic now avoids computing on `.data`, making it more
+  compatible with lazy tibbles (#1134).
+  
 * `chop()` now works correctly with data frames with 0 rows (#1206).
 
 * The `cols` argument of `chop()` is no longer optional. This matches the
--- a/R/nest.R
+++ b/R/nest.R
@@ -121,15 +121,22 @@
 nest <- function(.data, ..., .names_sep = NULL, .key = deprecated()) {
   cols <- enquos(...)
 
-  if (any(names2(cols) == "")) {
-    col_names <- names(tidyselect::eval_select(expr(c(...)), .data))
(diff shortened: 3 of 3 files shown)


tests/testthat/test-nest.R
```diff
@@ -315,6 +315,18 @@ test_that("warn about old style interface", {
   expect_named(out, c("x", "data"))
 })
 
+test_that("only warn about unnamed inputs (#1175)", {
+  df <- tibble(x = 1:3, y = 1:3, z = 1:3)
+  expect_warning(out <- nest(df, x, y, foo = z), "data = c(x, y)", fixed = TRUE)
```
hadley, line 320: Would these be better as snapshot tests?

DavisVaughan: Yes, but there are quite a few in this file so I will do a separate PR and move them all over to `expect_snapshot()`

PR #1209, opened 2021-11-09
Implement minimally invasive change to return list-ofs from `nest()`

DavisVaughan:
Just to assess revdeps

### DIFF
--- a/R/nest.R
+++ b/R/nest.R
@@ -181,6 +181,9 @@ nest.tbl_df <- function(.data, ..., .names_sep = NULL, .key = deprecated()) {
     out <- map(cols, ~ vec_split(set_names(.data[.x], names(.x)), keys)$val)
   }
 
+  ptypes <- map(cols, ~ vec_ptype(.data[.x]))
+  out <- map2(out, ptypes, new_list_of)
+
   vec_cbind(u_keys, new_data_frame(out, n = nrow(u_keys)))
 }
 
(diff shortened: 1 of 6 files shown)


DavisVaughan:
5 revdeps affected

## ggseq

- Boils down to the fact that we need a `vec_cast.list.list_of()` cast method for explicitly casting from list-of to list. They call `vec_cast(x, list())` directly
- Easy fix, and is something I was going to do anyways

## inspectdf

- They call `nest()`, and then call `replace(df, is.na(df), 0)` on that entire data frame (which has a bunch of different column types). This seems like a bug on their end, and it just happens to work right now by chance. 
- It won't work now because the `[<-` method of a list-of type is strict and won't let you replace with `0`

## isoreader

- Boils down to us needing a `vec_ptype2.vctrs_list_of.list()` method that is lax, and allows you to combine those two types together, returning a bare list. Essentially, its https://github.com/r-lib/vctrs/pull/1231, which is more proof that we should make the coercion methods lax

## PKNCA

- There is a strict test that checks the results after calling `nest()`. It ends up checking the attributes, which are now different since the nested list-of column will have a `ptype` attribute.
- Should be fixable with a more focused test

## ssimparser

- Calls `nest()`, then loops over the nested column in a rather odd way, and tries to replace each element using `[[<-` with an object of a different type (not allowed, since the list-of `[[<-` method is strict)
- Could be easily fixed with a simple `mutate(df, nested_col = lapply(nested_col, fn))` over the column instead

DavisVaughan:
Probably going to delay this until we decide on r-lib/vctrs#1231

PR #1218, opened 2021-11-12
Utilize `pack()` and `chop()` directly in `nest()`

DavisVaughan:
This PR makes `nest()` utilize `pack()` and `chop()` directly, like how `unnest()` uses `unpack()` and `unchop()`.

- We don't want `nest()` to return list-of cols, so we turn them back into list cols before returning
- `pack()` and `chop()` handle `reconstruct_tibble()` so we don't need that here
- The grouped-df method is now simpler, because that exact reconstruction is handled in `pack()` and `chop()`'s calls to `reconstruct_tibble()`

No changes to worse in revdeps


hadley (approved):
Looks like a nice simplification!

PR #1235, opened 2021-11-23
Update `wrap_error_names()` to override vctrs name repair `arg`

### LINKED ISSUE #987: Feature request: Add detail to error messages with `names_repair`
When using `pivot_wider()`, I accidentally created duplicate column names.  That was not my intent, but the error message provided made it take a bit more effort to find what the underlying problem was.

It would help if `names_repair` provided specific error messages indicating what was bad about the names (and possibly how to fix it).

For the toy example below, it would be helpful if the message were something like:

```
Error: Failed to create output due to bad names:  The name "A" is duplicated.
* Choose another strategy with `names_repair`
```

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
pivot_wider(
  data.frame(A=1, B="A", C=3),
  names_from="B",
  values_from="C"
)
#> Error in pivot_wider(data.frame(A = 1, B = "A", C = 3), names_from = "B", : could not find function "pivot_wider"
```

<sup>Created on 2020-06-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

DavisVaughan:
Closes #987

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # tidyr (development version)
 
+* `pivot_wider()` and `pivot_longer()` now generate more informative
+  errors related to name repair (#987).
+
 * The `values_fn` argument of `pivot_wider()` now correctly allows anonymous
   functions (#1114).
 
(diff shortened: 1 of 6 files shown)


billdenney:
Thanks!

PR #1246, opened 2021-11-30
Update snapshot tests with dev rlang and dev testthat

DavisVaughan:
- First commit moves to dev rlang
- Second commit then adds dev testthat on top of that

### DIFF
--- a/tests/testthat/_snaps/chop.md
+++ b/tests/testthat/_snaps/chop.md
@@ -4,7 +4,7 @@
       (expect_error(chop(df)))
     Output
       <error/rlang_error>
-      Argument `cols` is missing with no default
+      Error in `check_present()`: Argument `cols` is missing with no default
 
 # the ptype must be a list
 
@@ -12,7 +12,7 @@
       (expect_error(unchop(mtcars, mpg, ptype = 1)))
     Output
       <error/rlang_error>
-      `ptype` must be a named list.
+      Error in `df_unchop()`: `ptype` must be a named list.
 
 # incompatible sizes are caught
 
@@ -20,7 +20,7 @@
       (expect_error(unchop(df, c(x, y))))
     Output
       <error/rlang_error>
-      In row 1, can't recycle input of size 2 to size 3.
+      Error in `fn()`: In row 1, can't recycle input of size 2 to size 3.
 
 # empty typed inputs are considered in common size, but NULLs aren't
 
@@ -28,5 +28,5 @@
       (expect_error(unchop(df, c(x, y))))
     Output
       <error/rlang_error>
-      In row 1, can't recycle input of size 0 to size 2.
+      Error in `fn()`: In row 1, can't recycle input of size 0 to size 2.
 
--- a/tests/testthat/_snaps/pivot-wide.md
+++ b/tests/testthat/_snaps/pivot-wide.md
@@ -13,7 +13,7 @@
 
     Code
       out <- pivot_wider(df, names_from = key, values_from = val, names_repair = "unique")
-    Message <simpleMessage>
+    Message
       New names:
       * a -> a...1
       * a -> a...2
@@ -67,8 +67,8 @@
 
     Code
       pv <- pivot_wider(df, names_from = key, values_from = val)
-    Warning <rlang_warning>
-      Values from `val` are not uniquely identified; output will contain list-cols.
+    Condition
+      Warning: Values from `val` are not uniquely identified; output will contain list-cols.
       * Use `values_fn = list` to suppress this warning.
       * Use `values_fn = length` to identify where the duplicates arise.
       * Use `values_fn = {summary_fun}` to summarise duplicates.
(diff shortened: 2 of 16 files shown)


tests/testthat/_snaps/chop.md
```diff
@@ -4,29 +4,29 @@
...
+      Error in `df_unchop()`: `ptype` must be a named list.
 
 # incompatible sizes are caught
 
     Code
       (expect_error(unchop(df, c(x, y))))
     Output
       <error/rlang_error>
-      In row 1, can't recycle input of size 2 to size 3.
+      Error in `fn()`: In row 1, can't recycle input of size 2 to size 3.
```
DavisVaughan, line 23: This is unfortunate, it comes from using `purrr::reduce()` so I think it is out of our control

tests/testthat/_snaps/pivot-wide.md
```diff
@@ -98,5 +98,5 @@
       (expect_error(pivot_wider(df, values_fn = 1)))
     Output
       <error/rlang_error>
-      Can't convert a double vector to function
+      Error in `map()`: Can't convert `.x[[i]]`, a double vector, to a function.
```
DavisVaughan, line 101: Interesting change here

tests/testthat/_snaps/rectangle.md
```diff
@@ -4,79 +4,79 @@
...
+      Error in `hoist()`: `.data` must be a data frame.
 
 # unnest_wider - bad inputs generate errors
 
     Code
       (expect_error(unnest_wider(df, y)))
     Output
       <error/rlang_error>
-      Column `y` must contain a list of vectors.
+      Error in `.f()`: Column `y` must contain a list of vectors.
```
DavisVaughan, line 57: Comes from doing a `map(out, elt_to_wide)` where `elt_to_wide()` looks like `.f()`

DavisVaughan (commented):
Ultimately I don't think this looks too bad. There are a few `stop_vctrs()` calls being shown, but I don't think it is so many that I'd want to spend a bunch of time right now updating them to be more informative if a future vctrs release is going to fix that.

Many of the calls already reference an internal tidyr function, or the top level tidyr verb, which is pretty good

tests/testthat/_snaps/chop.md
lionel-: We could probably pass a lambda to avoid this.

tests/testthat/_snaps/pivot.md
```diff
@@ -15,21 +15,21 @@
       (expect_error(check_spec(df)))
     Output
       <error/rlang_error>
-      The `.name` column must be a character vector.
+      Error in `check_spec()`: The `.name` column must be a character vector.
```
lionel-, line 18: This should set the error call.

tests/testthat/_snaps/chop.md
```diff
@@ -4,29 +4,29 @@
       (expect_error(chop(df)))
     Output
       <error/rlang_error>
-      Argument `cols` is missing with no default
+      Error in `check_present()`: Argument `cols` is missing with no default
```
lionel-, line 7: This should set the error call, or maybe use the rlang function.

tests/testthat/_snaps/pivot-wide.md
```diff
@@ -52,23 +52,23 @@
...
+      Error in `build_wider_spec()`: `values_from` must select at least one column.
 
 # `values_fn` emits an informative error when it doesn't result in unique values (#1238)
 
     Code
       (expect_error(pivot_wider(df, values_fn = list(value = ~.x))))
     Output
       <error/rlang_error>
-      Applying `values_fn` to `value` must result in a single summary value per key.
+      Error in `vals_dedup()`: Applying `values_fn` to `value` must result in a single summary value per key.
```
lionel-, line 63: This should set the error call.

tests/testthat/_snaps/rectangle.md
```diff
@@ -4,79 +4,79 @@
...
     Output
       <error/rlang_error>
-      `.col` must identify a list-column.
+      Error in `hoist()`: `.col` must identify a list-column.
     Code
       (expect_error(df %>% hoist(x, 1)))
     Output
       <error/rlang_error>
-      All elements of `...` must be named.
+      Error in `check_pluckers()`: All elements of `...` must be named.
```
lionel-, line 28: This should set the error call.

tests/testthat/_snaps/rectangle.md
```diff
@@ -107,23 +107,23 @@
...
+      Error in `unnest_wider()`: `data` must be a data frame.
 
 # unnest_longer - bad inputs generate errors
 
     Code
       (expect_error(unnest_longer(df, y)))
     Output
       <error/rlang_error>
-      Column `y` must contain a list of vectors.
+      Error in `elt_to_long()`: Column `y` must contain a list of vectors.
```
lionel-, line 126: This should set the error call.

tests/testthat/_snaps/replace_na.md
```diff
@@ -4,15 +4,15 @@
       (expect_error(replace_na(1, 1:10)))
     Output
       <error/rlang_error>
-      Replacement for `data` is length 10, not length 1.
+      Error in `check_replacement()`: Replacement for `data` is length 10, not length 1.
```
lionel-, line 7: This should set the error call.

lionel- (commented):
I agree it doesn't look too bad overall. We can already improve messages in many cases by setting the error call in internal helpers and passing lambdas to functionals.

PR #1271, opened 2021-12-14
Implement `names_vary` for `pivot_wider()`

### LINKED ISSUE #839: FR: order of columns resulting from pivot_wider

DavisVaughan:
Closes #839 

Implements `names_vary`, as described compactly by this comment https://github.com/tidyverse/tidyr/issues/839#issuecomment-993924527

The actual coding of this is fairly straightforward. It just swaps `vec_rep()` for `vec_rep_each()` and vice versa while creating the spec columns.

As shown in https://github.com/tidyverse/tidyr/issues/839#issuecomment-993932843, this can currently be accomplished using `arrange()` on a spec generated by `build_wider_spec()`. But this _feels_ like it might be worth adding an argument for, since it seems to come up fairly often (2 issues and at least 2 stack overflow questions).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # tidyr (development version)
 
+* `pivot_wider()` gains a new `names_vary` argument for controlling the
+  ordering when combining `names_from` values with `values_from` column names
+  (#839).
+
 * `pivot_wider()` now gives better advice about how to identify duplicates when
   values are not uniquely identified (#1113).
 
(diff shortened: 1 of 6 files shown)


DavisVaughan:
@hadley I'll leave this up to you as to whether you think this is the best way forward for that issue

hadley (approved):
I like it!

PR #1279, opened 2021-12-16
Remove maturing badges for `chop()` and `pack()`

### LINKED ISSUE #1086: Review use of lifecycle badges/signals

DavisVaughan:
Closes #1086 

Upon review:

Changed:
- `chop()` and `pack()` were maturing, but I think they are stable now

Unchanged:
- `spread()` is marked as superseded
- `gather()` is marked as superseded
- `nest_legacy()` and `unnest_legacy()` are marked as superseded
- `nest()` has two deprecated arguments
   - Unnamed `...`
   - `.key`
- `unnest()` has four deprecated arguments
   - `.drop` / `.preserve`
   -   `.id`
   - `.sep`
- We have deprecated lazy eval versions of the tidyr verbs

I doubt any of this should be changed right now

At first I thought that the lazy eval versions might could be removed, but you have this in the 1.0.0 NEWS:

> All other lazyeval functions have been formally deprecated, and will be made defunct in the next major release.

### DIFF
--- a/R/chop.R
+++ b/R/chop.R
@@ -1,8 +1,6 @@
 #' Chop and unchop
 #'
 #' @description
-#' `r lifecycle::badge("maturing")`
-#'
 #' Chopping and unchopping preserve the width of a data frame, changing its
 #' length. `chop()` makes `df` shorter by converting rows within each group
 #' into list-columns. `unchop()` makes `df` longer by expanding list-columns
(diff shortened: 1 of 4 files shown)


hadley (approved):
I think I probably meant minor release, since major releases are so rare. If removing them doesn't break any revdeps, I think it's fine to delete the lazyeval methods.

PR #1277, opened 2021-12-16
Implement `id_expand` and `names_expand` for `pivot_wider()`

DavisVaughan:
Closes #770 
Related to #560 

- `names_expand` calls `expand()` on the `names_from` columns while building out the spec. This should be rather cheap, as we don't have to expand the entire data frame.
- `id_expand` calls `complete()` and completes the `id_cols`. The data frame it completes contains the `id_cols` along with the `names_from` and `values_from` columns. I don't think there is any straightforward way to avoid calling `complete()` in favor of just calling `expand()` on the `id_cols`.
  - During the `complete()` stage, we fill in the previously implicit missing values in the expanded `values_from` columns

---

Is it worth it? I _think_ so, especially when you look at a few examples and consider the alternatives:
- It is very hard to reach the same efficiency as the approach implemented in this PR
- It is often hard to remember how to even get a "correct" result

Here is an example with `names_expand`:

``` r
library(tidyr)
weekdays <- c("Mon", "Tue", "Wed", "Thu", "Fri", "Sat", "Sun")

daily <- tibble(
  day = factor(c("Tue", "Thu", "Fri", "Mon"), levels = weekdays),
  value = c(2, 3, 1, 5)
)

daily
#> # A tibble: 4 × 2
#>   day   value
#>   <fct> <dbl>
#> 1 Tue       2
#> 2 Thu       3
#> 3 Fri       1
#> 4 Mon       5

# How else can you get this?
pivot_wider(daily, names_from = day, values_from = value, names_expand = TRUE)
#> # A tibble: 1 × 7
#>     Mon   Tue   Wed   Thu   Fri   Sat   Sun
#>   <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
#> 1     5     2    NA     3     1    NA    NA

# complete() first, then pivot.
# Note that this isn't efficient! When it adds a new row during
# the completing process, it adds NAs for all the other columns (i.e. `value`).
# All we have to do with `names_expand` is `expand(daily, day)`, which is less
# expensive
daily_comp <- complete(daily, day)
daily_comp
#> # A tibble: 7 × 2
#>   day   value
#>   <fct> <dbl>
#> 1 Mon       5
#> 2 Tue       2
#> 3 Wed      NA
#> 4 Thu       3
#> 5 Fri       1
#> 6 Sat      NA
#> 7 Sun      NA

pivot_wider(daily_comp, names_from = day, values_from = value)
#> # A tibble: 1 × 7
[...]

PR #1280, opened 2021-12-16
Implement `unused_fn` for summarizing unused columns

### LINKED ISSUE #990: Feature request: Can pivot_wider() have an option to preserve non-`id_cols` column by chopping into list columns?
(I asked this on [RStudio Community](https://community.rstudio.com/t/what-is-the-best-way-to-pivot-wider-and-summarise-at-the-same-time/71914) first, but couldn't get a nice answer. So I thought this is worth a feature request.)

At first, honestly I didn't feel it's useful that `pivot_wider()` creates list columns when there are any duplicated values. But, now I find it very useful, in that it allows us to delay the decision of how to handle the duplication. This request is basically about extending it to non-`id_cols` column.

Suppose I have this data:

``` r
library(tidyr)

d <- data.frame(
  id = rep(1:5, each = 3),
  update = c(as.Date("2020-07-01") + 1:15),
  key = rep(c("a", "b", "c"), 5),
  val = 1:15
)

d
#>    id     update key val
#> 1   1 2020-07-02   a   1
#> 2   1 2020-07-03   b   2
#> 3   1 2020-07-04   c   3
#> 4   2 2020-07-05   a   4
#> 5   2 2020-07-06   b   5
#> 6   2 2020-07-07   c   6
#> 7   3 2020-07-08   a   7
#> 8   3 2020-07-09   b   8
#> 9   3 2020-07-10   c   9
#> 10  4 2020-07-11   a  10
#> 11  4 2020-07-12   b  11
#> 12  4 2020-07-13   c  12
#> 13  5 2020-07-14   a  13
#> 14  5 2020-07-15   b  14
#> 15  5 2020-07-16   c  15
```

and want to

* pivot `key` to columns and
* get the latest `update` of each `id`

at the same time. Currently, I can achieve this by summarising `update` with `mutate()` beforehand:

``` r 
d %>% 
  # summarise `update` by using `mutate()`
  group_by(id) %>% 
  mutate(last_update = max(update)) %>% 
  ungroup() %>% 
  # choose `id` and `last_update` as key
  pivot_wider(c(id, last_update),
    names_from = key,
    values_from = val
  )
#> # A tibble: 5 x 5
#>      id last_update     a     b     c
#>   <int> <date>      <int> <int> <int>
#> 1     1 2020-07-04      1     2     3
#> 2     2 2020-07-07      4     5     6
#> 3     3 2020-07-10      7     8     9
#> 4     4 2020-07-13     10    11    12
#> 5     5 2020-07-16     13    14    15
```

This is fine. But, I wander if it's possible that `pivot_wider()` chops non-`id_cols` column into a list column instead of just dropping. If it does, I can take `max()`s of them. Here's a pseudo-code; `update` isn't specified either on `id_cols`, `names_from`, or `values_from`), so it gets dropped, but I want to preserve the column.

``` r
d %>% 
  pivot_wider(id,
    names_from = key,
    values_from = val
  ) %>%
  rowwise() %>%
  mutate(last_update = max(update), .keep = "unused")
```

Or, more simply, this:

``` r
d %>% 
  pivot_wider(id,
    names_from = key,
    values_from = val,
    values_fn = max
  )
```

DavisVaughan:
Closes #990 
Closes #1076 (CC @mgirlich I've built off your PR from there and mentioned you in the NEWS)

~_This means we have to tweak https://github.com/tidyverse/tidyr/pull/1277 to also include the columns identified by `unused_fn` when performing the `id_expand` `complete()` step._~ Done in this PR

I was initially a little resistant to this feature, but I've come around and actually think it is pretty neat and solves an otherwise clunky problem. @hadley what do you think? I know we are starting to add quite a few arguments to pivot-wider...

The only _slightly_ weird thing about it is that in order for `unused_fn` to be useful, you have to specify `id_cols` explicitly, otherwise all unspecified cols are treated as `id_cols`! I have mentioned this in the docs.

The original issue uses the following data, and wants to retain the maximum `update` date for each key while pivoting.

``` r
library(tidyr)

df <- tibble(
  id = rep(1:5, each = 3),
  update = as.Date("2020-07-01") + 1:15,
  key = rep(c("a", "b", "c"), times = 5),
  val = 1:15
)
df
#> # A tibble: 15 × 4
#>       id update     key     val
#>    <int> <date>     <chr> <int>
#>  1     1 2020-07-02 a         1
#>  2     1 2020-07-03 b         2
#>  3     1 2020-07-04 c         3
#>  4     2 2020-07-05 a         4
#>  5     2 2020-07-06 b         5
#>  6     2 2020-07-07 c         6
#>  7     3 2020-07-08 a         7
#>  8     3 2020-07-09 b         8
#>  9     3 2020-07-10 c         9
#> 10     4 2020-07-11 a        10
#> 11     4 2020-07-12 b        11
#> 12     4 2020-07-13 c        12
#> 13     5 2020-07-14 a        13
#> 14     5 2020-07-15 b        14
#> 15     5 2020-07-16 c        15

# Pivot on `name` with a key of `id`, but retain the most recent `update` date
pivot_wider(
  data = df, 
  id_cols = id, 
  names_from = key, 
  values_from = val, 
  unused_fn = list(update = max)
)
#> # A tibble: 5 × 5
#>      id     a     b     c update    
#>   <int> <int> <int> <int> <date>    
#> 1     1     1     2     3 2020-07-04
#> 2     2     4     5     6 2020-07-07
#> 3     3     7     8     9 2020-07-10
#> 4     4    10    11    12 2020-07-13
#> 5     5    13    14    15 2020-07-16

# Same as above, but delay the aggregation altogether, instead just chopping
# the `update` column into a list column
pivot_wider(
  data = df, 
  id_cols = id, 
  names_from = key, 
  values_from = val, 
  unused_fn = list(update = list)
)
#> # A tibble: 5 × 5
#>      id     a     b     c update    
#>   <int> <int> <int> <int> <list>    
#> 1     1     1     2     3 <date [3]>
#> 2     2     4     5     6 <date [3]>
#> 3     3     7     8     9 <date [3]>
#> 4     4    10    11    12 <date [3]>
#> 5     5    13    14    15 <date [3]>
```

<sup>Created on 2021-12-16 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/R/pivot-wide.R
+++ b/R/pivot-wide.R
@@ -151,38 +151,6 @@
 #'     values_from = breaks,
 #'     values_fn = ~mean(.x, na.rm = TRUE)
 #'   )
-#'
-#' # Unused columns that aren't involved in the pivoting process are dropped
-#' # from the result by default. If you want to keep them, you can summarize
-#' # them with `unused_fn`. In this case, we'd like to retain the most recent
-#' # `update` date while pivoting on `name`.
-#' df <- tibble(
-#'   id = rep(1:5, each = 3),
-#'   update = as.Date("2020-07-01") + 1:15,
-#'   name = rep(c("a", "b", "c"), times = 5),
-#'   value = 1:15
-#' )
-#' df
-#'
-#' # You have to explicitly specify `id_cols` here, which allows `update` to
-#' # be recognized as an unused column
-#' pivot_wider(
-#'   data = df,
-#'   id_cols = id,
-#'   names_from = name,
-#'   values_from = value,
-#'   unused_fn = max
-#' )
-#'
-#' # Alternatively, you can keep all of the data and opt out of
-#' # immediately summarizing by converting to a list
-#' pivot_wider(
-#'   data = df,
-#'   id_cols = id,
-#'   names_from = name,
-#'   values_from = value,
-#'   unused_fn = list
-#' )
 pivot_wider <- function(data,
                         id_cols = NULL,
                         id_expand = FALSE,
--- a/R/pivot-wide.R
+++ b/R/pivot-wide.R
@@ -74,6 +74,20 @@
 #'
 #'   This can be a named list if you want to apply different aggregations
 #'   to different `values_from` columns.
+#' @param unused_fn Optionally, a function applied to summarize the values from
+#'   the unused columns (i.e. columns not identified by `id_cols`,
+#'   `names_from`, or `values_from`).
+#'
+#'   The default drops all unused columns from the result.
+#'
+#'   This can be a named list if you want to apply different aggregations
+#'   to different unused columns.
+#'
+#'   `id_cols` must be supplied for `unused_fn` to be useful, since otherwise
+#'   all unspecified columns will be considered `id_cols`.
+#'
+#'   This is similar to grouping by the `id_cols` then summarizing the
+#'   unused columns using `unused_fn`.
 #' @param ... Additional arguments passed on to methods.
 #' @export
 #' @examples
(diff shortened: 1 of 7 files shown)


R/pivot-wide.R
```diff
@@ -125,6 +139,38 @@
...
+#' df
+#'
+#' # You have to explicitly specify `id_cols` here, which allows `update` to
+#' # be recognized as an unused column
+#' pivot_wider(
+#'   data = df,
+#'   id_cols = id,
+#'   names_from = name,
+#'   values_from = value,
+#'   unused_fn = max
```
hadley, line 162: I feel like `first()` would be another good example to show. It comes up when you're pivoting data that's implicitly a join between two tables (so a bunch of variables are just repeated).

hadley (approved):
Looks good! I think it's worth including this in the vignette too.

R/pivot-wide.R
DavisVaughan: Do you have an example? Why would the repeated variable not just be another `id_col`?

PR #1281, opened 2021-12-17
Advance all lazyeval functions to the defunct stage

DavisVaughan:


DavisVaughan:
- dartR failed because plotly uses `gather_()` https://github.com/plotly/plotly.R/issues/2088
- modeltime.resample failed because plotly uses `gather_()` https://github.com/plotly/plotly.R/issues/2088

- cogmapr is a known breakage from a different issue
- InjurySeverityScore failed because tidyr wasn't available? Not sure what happened there.

DavisVaughan:
I just noticed that not all the lazyeval verbs were soft deprecated (like `gather_()`). So maybe we should start by soft deprecating them all and then in a future release we can defunct them

PR #1300, opened 2022-01-10
Disallow access to group columns in `expand()` and `complete()`

### LINKED ISSUE #1299: Should `complete()` and `expand()` disallow accessing the group columns?
With #1289 we added a grouped-df method for `complete()`. Originally I made the decision to pass `cur_data_all()` through to each ungrouped call to `complete()` because it seemed like people were (nonsensically) completing on grouping variables. This actually causes issues!

https://github.com/tidyverse/tidyr/blob/c73cf216434ca5edb1b560000d5e5aeca3bf39ef/R/complete.R#L99-L108

I am thinking we should just pass `cur_data()` through instead, which would avoid the issue seen below.

This would result in a __breaking change__ that people would no longer be able to complete on a grouping variable (You'd get an error saying that the grouping variable isn't found). But I think that completing on a grouping variable is a bit useless, as the whole point is to complete _within_ that variable, so you shouldn't have access to it.

``` r
library(tidyr)

df <- tibble(
  item_id = c(1:2, 2, 3),
  value1 = c(1, NA, 3, 4),
  item_name = c("a", "a", "b", "b"),
  value2 = 4:7,
  group = c(1:2, 1, 2)
)
gdf <- group_by(df, group)
gdf
#> # A tibble: 4 × 5
#> # Groups:   group [2]
#>   item_id value1 item_name value2 group
#>     <dbl>  <dbl> <chr>      <int> <dbl>
#> 1       1      1 a              4     1
#> 2       2     NA a              5     2
#> 3       2      3 b              6     1
#> 4       3      4 b              7     2

# expand "within" each group. so the group value should be repeated.
expand(gdf, item_id, item_name)
#> # A tibble: 8 × 3
#> # Groups:   group [2]
#>   group item_id item_name
#>   <dbl>   <dbl> <chr>    
#> 1     1       1 a        
#> 2     1       1 b        
#> 3     1       2 a        
#> 4     1       2 b        
#> 5     2       2 a        
#> 6     2       2 b        
#> 7     2       3 a        
#> 8     2       3 b

# (the NAs in the 1 `group` should be 1, and should be 2 in the 2 `group`)
# we don't get that here because we passed the "full" data including grouping
# cols through to `complete.data.frame()`, which did a full-join resulting in
# missing values for those columns. So then `summarize()` thought it didn't
# need to add the grouping columns back on.
complete(gdf, item_id, item_name)
#> # A tibble: 8 × 5
#> # Groups:   group [3]
#>   group item_id item_name value1 value2
#>   <dbl>   <dbl> <chr>      <dbl>  <int>
#> 1     1       1 a              1      4
#> 2    NA       1 b             NA     NA
#> 3    NA       2 a             NA     NA
#> 4     1       2 b              3      6
#> 5     2       2 a             NA      5
#> 6    NA       2 b             NA     NA
#> 7    NA       3 a             NA     NA
#> 8     2       3 b              4      7
```

<sup>Created on 2022-01-10 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

DavisVaughan:
Closes #1299
Follow up to #1289 

In #1289 we added a grouped-df method for `complete()` that calls `complete()` "within" each group. This would send the entire group slice of data (including the grouping column) into the inner call to `complete()`. This turns out to be buggy, since it can generate missing values in the grouping columns if the within-group-expansion adds rows:

``` r
library(tidyr)

df <- tibble(
  g = c("x", "x", "y"),
  a = c(1, 2, 2),
  b = c(1, 2, 2)
)
gdf <- group_by(df, g)
gdf
#> # A tibble: 3 × 3
#> # Groups:   g [2]
#>   g         a     b
#>   <chr> <dbl> <dbl>
#> 1 x         1     1
#> 2 x         2     2
#> 3 y         2     2

# The `NA`s should be `"x"` values.
# Columns `a` and `b` should be completed "within" each group of `g`
complete(gdf, a, b)
#> # A tibble: 5 × 3
#> # Groups:   g [3]
#>   g         a     b
#>   <chr> <dbl> <dbl>
#> 1 x         1     1
#> 2 <NA>      1     2
#> 3 <NA>      2     1
#> 4 x         2     2
#> 5 y         2     2
```

We never saw this with `expand()`, because each call to `expand()` "within" each group would only return the expansion columns, then the outer `summarize()` would add the group columns back on.

``` r
expand(gdf, a, b)
#> # A tibble: 5 × 3
#> # Groups:   g [2]
#>   g         a     b
#>   <chr> <dbl> <dbl>
#> 1 x         1     1
#> 2 x         1     2
#> 3 x         2     1
#> 4 x         2     2
#> 5 y         2     2
```

Compare this with `complete()`, where each inner call to `complete()` "within" each group returns all the columns in the data frame. This includes the grouping columns if they are passed through, and `summarize()` won't overwrite those, which is why we ended up with the problem above.

The fix is to use `cur_data()` rather than `cur_data_all()`, forcing `summarize()` to handle the re-addition of any group columns. This means you can no longer attempt to "complete" or "expand" on a group column, but that was pretty much undefined behavior previously, since conceptually you should be completing/expanding "within" each group, meaning you don't have access to that group info.

### DIFF
--- a/tests/testthat/test-expand.R
+++ b/tests/testthat/test-expand.R
@@ -44,39 +44,67 @@ test_that("expand accepts expressions", {
   expect_equal(df, crossing(x = 1:3, y = 3:1))
 })
 
-test_that("expand respects groups", {
-  local_options(lifecycle_verbosity = "quiet")
-
+test_that("expand will expand within each group (#396)", {
   df <- tibble(
+    g = c("a", "b", "a"),
     a = c(1L, 1L, 2L),
-    b = c(1L, 2L, 1L),
-    c = c(2L, 1L, 1L)
-  )
-  out <- df %>% dplyr::group_by(a) %>% expand(b, c) %>% nest()
-
-  expect_equal(out$data[[1]], crossing(b = 1:2, c = 1:2))
-  expect_equal(out$data[[2]], tibble(b = 1L, c = 1L))
-})
-
-test_that("expand allows expanding on grouping variable (#396)", {
-  df <- tibble(
-    g = factor(c("a", "b", "b")),
-    a = c(1, 1, 3)
+    b = factor(c("a", "a", "b"), levels = c("a", "b", "c"))
   )
   gdf <- dplyr::group_by(df, g)
 
-  out <- expand(gdf, g, a)
+  out <- expand(gdf, a, b)
 
-  # Same as split by group, expand, combine.
-  # Note that this produces duplicate rows in the result. Not ideal, but
-  # would probably break revdeps if we removed this grouped-df method.
-  expect <- vec_rbind(
-    expand(df[1,], g, a),
-    expand(df[2:3,], g, a)
+  # Still grouped
+  expect_identical(dplyr::group_vars(out), "g")
+
+  out <- nest(out, data = -g)
+
+  expect_identical(out$data[[1]], crossing(a = 1:2, b = factor(levels = c("a", "b", "c"))))
+  expect_identical(out$data[[2]], crossing(a = 1L, b = factor(levels = c("a", "b", "c"))))
+})
+
+test_that("expand does not allow expansion on grouping variable (#1299)", {
+  df <- tibble(
+    g = "x",
+    a = 1L
   )
-  expect <- dplyr::group_by(expect, g)
+  gdf <- dplyr::group_by(df, g)
 
-  expect_identical(out, expect)
+  # This is a dplyr error that we don't own
+  expect_error(expand(gdf, g))
+})
+
+test_that("can use `.drop = FALSE` with expand (#1299)", {
+  levels <- c("a", "b", "c")
+
+  df <- tibble(
+    g = factor(c("a", "b", "a"), levels = levels),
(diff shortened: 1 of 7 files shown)


tests/testthat/test-expand.R
```diff
@@ -44,39 +44,67 @@ test_that("expand accepts expressions", {
...
+
+test_that("can use `.drop = FALSE` with expand (#1299)", {
+  levels <- c("a", "b", "c")
 
-  # Same as split by group, expand, combine.
-  # Note that this produces duplicate rows in the result. Not ideal, but
-  # would probably break revdeps if we removed this grouped-df method.
-  expect <- vec_rbind(
-    expand(df[1,], g, a),
-    expand(df[2:3,], g, a)
```
DavisVaughan, line 75: I also really like this change because it means that we can no longer produce duplicate rows in the result of `expand()` (which I previously mentioned should be an invariant of this function). This was only an issue if we allow expansion on a grouping column

hadley (approved):

PR #1319, opened 2022-02-09
let fill() handle data with a .direction column

tjmahr:
My data had a column named `.direction` so there was an environment-vs-data-mask ambiguity that caused fill() to fail. I added `!!` to force evaluation of the `.direction` argument.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # tidyr (development version)
 
+* `fill()` now works correctly when there is a column named `.direction` in
+  `data` (#1319, @tjmahr).
+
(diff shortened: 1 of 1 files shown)


tjmahr:
switching from `!! .direction` to `.env$.direction` based on the rlang docs

> Note that in most cases you don't need injection with !!. For instance, the [.data](https://rlang.r-lib.org/reference/dot-data.html) and [.env](https://rlang.r-lib.org/reference/dot-data.html) pronouns provide more intuitive alternatives to injecting a column name and injecting a value.

PR #1321, opened 2022-02-10
Improve `id_cols` OOB error message

### LINKED ISSUE #1318: new version breaks pivot_wider
Latest version of tidyr returns error about not being able to subset names that don't exist.

Reverting to ver. 1.1.4 makes the code work again.

I've no time to create a reprex, but here's essentially what breaks:

 ```
temp.pivot <- pivot_wider(
    output.df,
    id_cols = "boom",
    values_from = c(paste0(r,"some_suffix_name")), 
    names_from = "boom"
  )
```

### PR DESCRIPTION
**Pull request by DavisVaughan**
Closes #1318 

This PR improves on the `id_cols` error message that occurs when a column selected by `id_cols` was also selected by `names_from` or `values_from` (which meant that it was already removed at this point, resulting in the OOB error).

``` r
library(tidyr)

df <- tibble(name = c("x", "y"), value = c(1, 2))

# `name` and `value` already removed from `df`, so evaluating the `id_cols = name` tidyselect gives this error:
pivot_wider(df, id_cols = name, names_from = name, values_from = value)

# before
#> Error in `stop_subscript()`:
#> ! Can't subset columns that don't exist.
#> x Column `name` doesn't exist.

# after
#> Error in `select_wider_id_cols()`:
#> ! `id_cols` can't select a column already selected by `names_from` or `values_from`.
#> x Column `name` has already been selected.
```

@lionel- does this usage of `try_fetch()` look right to you? In particular, I have used both `abort(parent = NA)` and `cnd_signal()` here, and I am not 100% confident of either one.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 # tidyr (development version)
 
-* Improved the error message returned when a column selected by `names_from` or
-  `values_from` is also selected by `id_cols` (#1318).
+* Improved the error message returned by `pivot_wider()` when a column selected
+  by `names_from` or `values_from` is also selected by `id_cols` (#1318).
 
 # tidyr 1.2.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # tidyr (development version)
 
+* Improved the error message returned when a column selected by `names_from` or
+  `values_from` is also selected by `id_cols` (#1318).
+
 # tidyr 1.2.0
 
 ## Pivoting
--- a/R/pivot-wide.R
+++ b/R/pivot-wide.R
@@ -574,7 +574,8 @@ rethrow_id_cols_oob <- function(cnd, non_id_cols) {
   if (i %in% non_id_cols) {
     stop_id_cols_oob(i)
   } else {
-    cnd_signal(cnd)
+    # Zap this special handler, throw the normal condition
+    zap()
   }
 }
 
--- a/R/pivot-wide.R
+++ b/R/pivot-wide.R
@@ -582,7 +582,7 @@ rethrow_id_cols_oob <- function(cnd, non_id_cols) {
 stop_id_cols_oob <- function(i) {
   message <- c(
     "`id_cols` can't select a column already selected by `names_from` or `values_from`.",
-    x = glue("Column `{i}` has already been selected.")
+    i = glue("Column `{i}` has already been selected.")
   )
   abort(message, parent = NA)
 }
--- a/R/pivot-wide.R
+++ b/R/pivot-wide.R
@@ -546,16 +546,44 @@ select_wider_id_cols <- function(data,
                                  non_id_cols = character()) {
   id_cols <- enquo(id_cols)
 
-  # Remove known non-id-cols so they are never selected
+  # Remove known `non_id_cols` so they are never selected
   data <- data[setdiff(names(data), non_id_cols)]
 
   if (quo_is_null(id_cols)) {
-    names(data)
-  } else {
+    # Default selects everything in `data` after `non_id_cols` have been removed
+    return(names(data))
+  }
+
+  try_fetch(
     # TODO: Use `allow_rename = FALSE`.
     # Requires https://github.com/r-lib/tidyselect/issues/225.
-    names(tidyselect::eval_select(enquo(id_cols), data))
+    id_cols <- tidyselect::eval_select(enquo(id_cols), data),
+    vctrs_error_subscript_oob = function(cnd) rethrow_id_cols_oob(cnd, non_id_cols)
+  )
+
+  names(id_cols)
+}
+
+rethrow_id_cols_oob <- function(cnd, non_id_cols) {
+  i <- cnd[["i"]]
+
+  if (!is_string(i)) {
+    abort("`i` is expected to be a string.", .internal = TRUE)
   }
+
+  if (i %in% non_id_cols) {
+    stop_id_cols_oob(i)
+  } else {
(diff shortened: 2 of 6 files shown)


R/pivot-wide.R
```diff
@@ -542,16 +546,44 @@ select_wider_id_cols <- function(data,
...
+  i <- cnd[["i"]]
+
+  if (!is_string(i)) {
+    abort("`i` is expected to be a string.", .internal = TRUE)
   }
+
+  if (i %in% non_id_cols) {
+    stop_id_cols_oob(i)
+  } else {
+    cnd_signal(cnd)
```
DavisVaughan, line 577: Is this right? I want to allow the `cnd` to be signaled "like normal" here

R/pivot-wide.R
```diff
@@ -542,16 +546,44 @@ select_wider_id_cols <- function(data,
...
+  }
+}
+
+stop_id_cols_oob <- function(i) {
+  message <- c(
+    "`id_cols` can't select a column already selected by `names_from` or `values_from`.",
+    x = glue("Column `{i}` has already been selected.")
+  )
+  abort(message, parent = NA)
```
DavisVaughan, line 586: Here is where I think I am saying, "I want to fully control this error, even though it originally came from tidyselect"

R/pivot-wide.R
```diff
@@ -542,16 +546,44 @@ select_wider_id_cols <- function(data,
...
-    names(tidyselect::eval_select(enquo(id_cols), data))
+    id_cols <- tidyselect::eval_select(enquo(id_cols), data),
+    vctrs_error_subscript_oob = function(cnd) rethrow_id_cols_oob(cnd, non_id_cols)
+  )
+
+  names(id_cols)
+}
+
+rethrow_id_cols_oob <- function(cnd, non_id_cols) {
+  i <- cnd[["i"]]
```
DavisVaughan, line 568: Nicely, the vctrs error passes through a string `i` containing the problem column. I think this should be stable at this point, so I felt comfortable using it here

R/pivot-wide.R
```diff
@@ -542,16 +546,44 @@ select_wider_id_cols <- function(data,
                                  non_id_cols = character()) {
   id_cols <- enquo(id_cols)
 
-  # Remove known non-id-cols so they are never selected
+  # Remove known `non_id_cols` so they are never selected
   data <- data[setdiff(names(data), non_id_cols)]
```
hadley, line 550: Are you sure it wouldn't be better to leave data as is and then error if any of `non_id_cols` appear in `id_cols`?

R/pivot-wide.R
lionel-: This works but if you just want to let the cnd pass through unchanged you should `return(zap())`. Returning the zap sentinel in `try_fetch()` is equivalent to a return in `withCallingHandlers()`.

R/pivot-wide.R
lionel-: Regarding the message content, this should probably be an info bullet rather than a cross bullet. The latter typically have "can't" or "must" phrasing.

It would be nice to mention which of `names_from` and `values_from` contains the duplicate selection in the info bullet.

R/pivot-wide.R
lionel-: I think so. In general, internal condition data should be stored in a list in a `my_package$internal` field.

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # tidyr (development version)
 
+* Improved the error message returned when a column selected by `names_from` or
```
lionel-, line 3:
```suggestion
* Improved the error message returned by `pivot_wider()` when a column selected by `names_from` or
```

lionel- (commented):
Will also need to pass the `pivot_wider()` call once dev tidyselect is on CRAN.

R/pivot-wide.R
DavisVaughan: I wanted to ensure that using something like `id_cols = everything()` would exactly match the default `id_cols = NULL` behavior, i.e.:

``` r
library(tidyr)

df <- tibble(
  id = c("a", "a", "b"),
  name = c("x", "y", "x"), 
  value = c(1, 2, 2)
)
df
#> # A tibble: 3 × 3
#>   id    name  value
#>   <chr> <chr> <dbl>
#> 1 a     x         1
#> 2 a     y         2
#> 3 b     x         2

pivot_wider(
  df,
  id_cols = everything(), # <- exactly matches default `NULL` behavior. "everything" except names_from/values_from
  names_from = name,
  values_from = value
)
#> # A tibble: 2 × 3
#>   id        x     y
#>   <chr> <dbl> <dbl>
#> 1 a         1     2
#> 2 b         2    NA
```

This works in both tidyr 1.2.0 _and_ 1.1.4. It works in 1.1.4 because we called `eval_select()` on all of `data` as you are saying here, and then just did a `setdiff()` on the `non_id_cols`. But this resulted in the original confusing case where you could specify a particular column in both `id_cols` and `names/values_from` and it would get silently (and confusingly) removed from `id_cols`.

If we left `data` as is and errored on any `id_cols` that were also in `non_id_cols`, then we'd always error on the `id_cols = everything()` case, because it would always select something that overlaps with both `names_from` and `values_from`. Since this used to work in pre 1.2.0 tidyr, I'd be worried that it may break someone's code.

You could also imagine some usage of `starts_with()` or any other tidyselect helper that is dynamic being used for `id_cols`, and you could run into the same issue.

hadley: Got it!

Issue #1332, opened 2022-03-08
Bug in v1.2 : `complete()` errors when given the grouping variable of a grouped df


daattali:
The following code worked in version 1.1.3

```
df <- dplyr::tibble(
    x = c(1:3, 1:3),
    y = c(1, NA, 3, NA, 5, NA),
    z = c(1:6)
)
df <- dplyr::group_by(df, x)
tidyr::complete(df, x, y, z)
```

But in version 1.2 it breaks with this error:

```
Error in `dplyr::summarise()`:
! Problem while computing `..1 = complete(data = dplyr::cur_data(), ..., fill =
  fill, explicit = explicit)`.
i The error occurred in group 1: x = 1.
Caused by error:
! object 'x' not found
```

If you change the grouping variable to `y` instead of `x`, you'll get the same error but with "y". If you don't have the grouping variable in the `complete()` call, then it works.

The output to the first chunk should be

```
# A tibble: 12 x 3
# Groups:   x [3]
       x     y     z
   <int> <dbl> <int>
 1     1     1     1
 2     1     1     4
 3     1    NA     1
 4     1    NA     4
 5     2     5     2
 6     2     5     5
 7     2    NA     2
 8     2    NA     5
 9     3     3     3
10     3     3     6
11     3    NA     3
12     3    NA     6
```

DavisVaughan:
This is an intended change documented in the NEWS https://github.com/tidyverse/tidyr/blob/main/NEWS.md#grids

And in the Details section of `?complete`

> With grouped data frames, complete() operates within each group. Because of this, you cannot complete a grouping column.

---

It has always been intended that `complete()` should operate _within_ each group (which implies that you shouldn't have access to the groups themselves when specifying columns to complete on). `complete()` with grouped-dfs had been a little broken for quite some time (https://github.com/tidyverse/tidyr/issues/396 and https://github.com/tidyverse/tidyr/issues/966). The changes in 1.2.0 (this being one of them) made `complete()` more correct with grouped data frames.

Just don't complete on the group variable, and you get the result you expected.

The error message isn't the best here, but this one is particularly hard to improve on because of the way `complete()` works (sort of like `mutate()` where you can supply arbitrary expressions to `...`, rather than just selections, which make it hard for us to tell ahead of time if you specify a group column or not)

daattali:
Thanks for the explanation. Shouldn't this be documented as a breaking change, since code that worked previously is now erroring, even though only a minor version update occurred?

[...]

PR #1334, opened 2022-03-10
NEWS: mention breaking change related to #1332

daattali:
#1332 
I understand you won't realistically bump the version after-the-fact, but I think it's a good idea to modify the NEWS file. Previously [there was no mention of the inability to complete on a grouping](https://github.com/tidyverse/tidyr/blob/49753836537547511e331b1d55a44b3471b8f422/R/complete.R) so it's very possible others have done it and usually when code breaks you search for breaking changes to try to debug.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -21,6 +21,13 @@
   grouped data frame happens "within" each group and otherwise has the
   potential to produce erroneous results (#1299).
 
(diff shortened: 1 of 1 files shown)


DavisVaughan:
Thanks!

PR #1336, opened 2022-03-16
make replace_na work for matrices (again)

struckma:
Dear Hadley, dear other contributors,

I came across this issues running an internal legacy script calling `replace_na` against its specification for a matrix. Last year, this worked just fine with the implementation for vectors, but it stopped working now. I am not sure, in how far this would be desirable, but nevertheless, I think, if someone else has a similar problem, it should be shared somewhere. Since it is not exactly an issue, I have decided not to open an issue for this, but to open a pull request.

My solution would implement the generic `replace_na` also for matrices, and from there, it can basically use the default implementation, just the matrix-property is transiently dropped.

There are three considerations, that come to my mind:

  - My implementation should be more generic to work also for higher-dimensional arrays
  - using `as.matrix(as.vector(.))` is not the tidiest implementation I can think of, but it is really short and should work fast enough
  - The solution could break maybe newer code relying on the undocumented and (for my understanding) confusing behavior of the `replace_na` if called with a matrix in the current upstream-version (that obviously recycles the `missing`-vector instead of computing it element-wise for a matrix). However, since the function is not yet documented to support matrices at all, this may not really be an issue.

I would be happy, if you could consider my thoughts and maybe merge my code, but it would be perfectly fine and understandable for me, if you decide against.

Thank you

Stephan


DavisVaughan:
`replace_na()` _does_ work for matrices, just not in the way you are thinking.

With the upgrade of `replace_na()` to use vctrs, it now uses the vctrs definition of a missing value defined by `vctrs::vec_equal_na()`. For matrices and data frames, `vec_equal_na()` returns a logical vector with the same number of rows as the input and only returns `TRUE` for a particular row if _all_ of the columns in that row are `NA`.

If you want the old behavior I think you are going to have to call `as.vector()`, replace, then regenerate the matrix/array on the back side.

``` r
library(tidyr)

x <- matrix(
  c(1,  2,
    NA, 1,
    1,  NA,
    NA, NA),
  nrow = 4,
  byrow = TRUE
)

x
#>      [,1] [,2]
#> [1,]    1    2
#> [2,]   NA    1
#> [3,]    1   NA
#> [4,]   NA   NA

# Only the rows with all columns equal to NA are considered missing
vctrs::vec_equal_na(x)
#> [1] FALSE FALSE FALSE  TRUE

replace_na(x, replace = 99)
#>      [,1] [,2]
#> [1,]    1    2
#> [2,]   NA    1
#> [3,]    1   NA
#> [4,]   99   99

replace_na(x, replace = matrix(c(-99, 99), nrow = 1))
#>      [,1] [,2]
#> [1,]    1    2
#> [2,]   NA    1
#> [3,]    1   NA
#> [4,]  -99   99

# You should use this
array(replace_na(as.vector(x), replace = 99), dim = dim(x))
#>      [,1] [,2]
#> [1,]    1    2
#> [2,]   99    1
#> [3,]    1   99
#> [4,]   99   99
```

This definition of missingness for matrices and data frames is really nice when you consider data frame-columns and matrix-columns, so while I understand it may be a little confusing here, it is unlikely to change.

And since we want tidyr to be inline with how vctrs works, we are also unlikely to change `replace_na()` as well.

In the future, it is best to open a PR to discuss changes like this first, just so you don't end up doing unnecessary work! But we appreciate the PR!

PR #1340, opened 2022-04-01
Pass calls for improved error messages

### LINKED ISSUE #1313: Pass error call information through to `abort()`
To improve on the new rlang 1.0.0 error messages

Start from this PR, which captures the snapshot changes in 1.0.0 and shows where improvements could be made
https://github.com/tidyverse/tidyr/pull/1311

mgirlich:
Pass call to `abort()` and some vctrs functions for nicer error messages.
Fixes #1313.


hadley (approved):

DavisVaughan:
I'll look at these in some detail next week, thanks @mgirlich

hadley:
Let's wait a week or two for vctrs and tidyselect to be out, and then I'll finish off.

hadley:
I think this should be good for review. I worked through all the snapshot files so while there are probably more places that need updates, this feels like a good place to start.

R/expand.R
```diff
@@ -277,7 +277,7 @@ grid_dots <- function(..., `_data` = NULL) {
     }
 
     arg <- paste0("..", i)
-    vec_assert(dot, arg = arg)
+    vec_assert(dot, arg = arg, call = caller_env())
```
DavisVaughan, line 280: Probably worth adding `error_call = caller_env()` as an argument to `grid_dots()`?

R/chop.R
```diff
@@ -141,17 +141,17 @@ unchop <- function(data, cols, keep_empty = FALSE, ptype = NULL) {
 #   used to slice the data frame `x` was subset from to align it with `val`.
 # - `val` the unchopped data frame.
 
-df_unchop <- function(x, ..., ptype = NULL, keep_empty = FALSE) {
+df_unchop <- function(x, ..., ptype = NULL, keep_empty = FALSE, call = caller_env()) {
```
DavisVaughan, line 144: Our emerging convention when passing error calls through as arguments seems to be:

- Name them `call` and `arg` when the function's main purpose is for some kind of input checking, i.e. `vec_assert()`, `check_*()` functions, `abort()`, etc.

- Name them `error_call` and `error_arg` otherwise.

Since `df_unchop()`'s main purpose is unchopping, it should use `error_call`. The rest of the PR probably needs a pass to do this updating too. I'll try and mark them as I see them.

We have started doing this in vctrs, tidyselect, internally in dplyr, and I've done it in ivs, so probably best to take the time to be consistent here since call improvements is what this PR is all about.

R/chop.R
```diff
@@ -264,7 +264,7 @@ df_unchop <- function(x, ..., ptype = NULL, keep_empty = FALSE) {
   out
 }
 
-unchop_sizes2 <- function(x, y) {
+unchop_sizes2 <- function(x, y, call) {
```
DavisVaughan, line 267:
```suggestion
unchop_sizes2 <- function(x, y, error_call) {
```

R/extract.R
```diff
@@ -50,7 +50,7 @@ extract.data.frame <- function(data, col, into, regex = "([[:alnum:]]+)",
   reconstruct_tibble(data, out, if (remove) var else chr())
 }
 
-str_extract <- function(x, into, regex, convert = FALSE) {
+str_extract <- function(x, into, regex, convert = FALSE, call = caller_env()) {
```
DavisVaughan, line 53:
```suggestion
str_extract <- function(x, into, regex, convert = FALSE, error_call = caller_env()) {
```

R/hoist.R
```diff
@@ -133,7 +133,7 @@ hoist <- function(.data,
   out
 }
 
-check_pluckers <- function(...) {
+check_pluckers <- function(..., .call = caller_env()) {
```
DavisVaughan, line 136:
```suggestion
check_pluckers <- function(..., .error_call = caller_env()) {
```

R/pivot-wide.R
```diff
@@ -525,17 +526,19 @@ build_wider_spec <- function(data,
 build_wider_id_cols_expr <- function(data,
                                      id_cols = NULL,
                                      names_from = name,
-                                     values_from = value) {
+                                     values_from = value,
+                                     call = caller_env()) {
```
DavisVaughan, line 530:
```suggestion
                                     error_call = caller_env()) {
```

R/pivot-wide.R
```diff
@@ -525,17 +526,19 @@ build_wider_spec <- function(data,
...
+  names_from_cols <- names(tidyselect::eval_select(enquo(names_from), data, error_call = call))
+  values_from_cols <- names(tidyselect::eval_select(enquo(values_from), data, error_call = call))
 
   out <- select_wider_id_cols(
     data = data,
     id_cols = {{ id_cols }},
     names_from_cols = names_from_cols,
-    values_from_cols = values_from_cols
+    values_from_cols = values_from_cols,
+    call = caller_env()
```
DavisVaughan, line 541:
```suggestion
    call = error_call
```

?

R/pivot-wide.R
```diff
@@ -544,7 +547,8 @@ build_wider_id_cols_expr <- function(data,
 select_wider_id_cols <- function(data,
                                  id_cols = NULL,
                                  names_from_cols = character(),
-                                 values_from_cols = character()) {
+                                 values_from_cols = character(),
+                                 call = caller_env()) {
```
DavisVaughan, line 551:
```suggestion
                                 error_call = caller_env()) {
```

R/pivot-wide.R
```diff
@@ -558,38 +562,38 @@ select_wider_id_cols <- function(data,
...
-      rethrow_id_cols_oob(cnd, names_from_cols, values_from_cols)
+      rethrow_id_cols_oob(cnd, names_from_cols, values_from_cols, call)
     }
   )
 
   names(id_cols)
 }
 
-rethrow_id_cols_oob <- function(cnd, names_from_cols, values_from_cols) {
+rethrow_id_cols_oob <- function(cnd, names_from_cols, values_from_cols, call) {
```
DavisVaughan, line 574: `call` is fine here, main purpose is error related

R/replace_na.R
```diff
@@ -81,6 +81,6 @@ check_replacement <- function(x, var) {
   n <- vec_size(x)
 
   if (n != 1) {
-    abort(glue("Replacement for `{var}` is length {n}, not length 1."))
+    abort(glue("Replacement for `{var}` is length {n}, not length 1."), call = caller_env())
```
DavisVaughan, line 84: Worth adding `error_call = caller_env()` as an argument to `check_replacement()`, even thought it is only used in `replace_na()`.

I think adding the `error_call` argument should be the "go to" solution for most of these, because we never know how the helpers might be used in the future, and having 1 rule to follow means we don't have to think about it as much.

DavisVaughan (approved):
Looks really good! 90% of these comments are marking requests for `call` -> `error_call` updates based on some emerging conventions

R/hoist.R
hadley: According to your conventions this should be `.call`? Since this is a check function?

R/pivot.R
```diff
@@ -31,22 +31,22 @@
 #' check_pivot_spec(spec)
 check_pivot_spec <- function(spec) {
```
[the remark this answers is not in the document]
hadley: Yeah, it probably should

R/replace_na.R
hadley: Yeah, agreed. I just missed updating these.

R/hoist.R
DavisVaughan: Yea my bad

R/chop.R
lionel-: This means we need to rename a lot of already exported functions? Isn't it a bit late to change this now?

By the way I don't consider a dotted variant to represent a different name. I see it as different syntax for the same name.

hadley: Right, I don't think of them as different names either, but pragmatically you have to remember which of 4 options it is.

PR #1342, opened 2022-04-01
Remove unnecessary workarounds

### LINKED ISSUE #1212: Remove usage of `tidyr_temporary_new_list_of()`
Added in #1200 because of the issue outlined in https://github.com/r-lib/vctrs/pull/784

This will also fix this niche test added in #1200:

```r
test_that("can't currently retain names when simplification isn't done and a ptype is supplied if there are a mix of named/unnamed elements (#1212)", {
  df <- tibble(x = list(list(a = 1L), list(1L)))
  ptype <- list(x = integer())

  out <- unnest_longer(df, x, indices_include = TRUE, ptype = ptype, simplify = FALSE)

  # Ideally should be `c("a", "")`, but the call to
  # `tidyr_temporary_new_list_of()` in `col_simplify()` prevents this
  expect_named(out$x, NULL)
  expect_identical(out$x_id, c("a", NA))
})
```

mgirlich:
Some workarounds are now unnecessary thanks to newer versions of vctrs and rlang.
Fixes #1212.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -25,7 +25,7 @@ Imports:
     lifecycle,
     magrittr,
     purrr,
-    rlang (>= 1.0.1),
+    rlang (>= 1.0.2),
     tibble (>= 2.1.1),
     tidyselect (>= 1.1.0),
     utils,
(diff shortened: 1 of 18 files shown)


DavisVaughan:
Thanks!

PR #1346, opened 2022-04-08
Use `vec_assign()` instead of `vec_slice<-()`

mgirlich:
As this is generally preferred (see https://github.com/r-lib/vctrs/issues/1409).


DavisVaughan:
Thanks!

PR #1347, opened 2022-04-08
Implement `cols_vary` argument for `pivot_longer()`

DavisVaughan:
Closes #1312 

After looking at this a bit more, I've determined that this argument is mostly useful when you are pivoting every column in your dataset (both the `anscombe` dataset and my original example in #1312 do this). This is somewhat rare, so I'm happy that `cols_vary = "fastest"` was implicitly the default to begin with. That feels like the right choice VS what `gather()` does.

``` r
library(tidyr)

df <- tibble(
  start = as.Date(c("2019-01-01", "2019-01-02")),
  end = as.Date(c("2019-01-03", "2019-01-04"))
)
df
#> # A tibble: 2 × 2
#>   start      end       
#>   <date>     <date>    
#> 1 2019-01-01 2019-01-03
#> 2 2019-01-02 2019-01-04

pivot_longer(df, c(start, end))
#> # A tibble: 4 × 2
#>   name  value     
#>   <chr> <date>    
#> 1 start 2019-01-01
#> 2 end   2019-01-03
#> 3 start 2019-01-02
#> 4 end   2019-01-04

# Useful!
pivot_longer(df, c(start, end), cols_vary = "slowest")
#> # A tibble: 4 × 2
#>   name  value     
#>   <chr> <date>    
#> 1 start 2019-01-01
#> 2 start 2019-01-02
#> 3 end   2019-01-03
#> 4 end   2019-01-04


# The `religion` column naturally generates a "grouping" column that `income` is nested in
relig_income %>%
  pivot_longer(!religion, names_to = "income", values_to = "count")
#> # A tibble: 180 × 3
#>    religion income             count
#>    <chr>    <chr>              <dbl>
#>  1 Agnostic <$10k                 27
#>  2 Agnostic $10-20k               34
#>  3 Agnostic $20-30k               60
#>  4 Agnostic $30-40k               81
#>  5 Agnostic $40-50k               76
#>  6 Agnostic $50-75k              137
#>  7 Agnostic $75-100k             122
#>  8 Agnostic $100-150k            109
#>  9 Agnostic >150k                 84
#> 10 Agnostic Don't know/refused    96
#> # … with 170 more rows

# So `"slowest"` isn't as useful here
relig_income %>%
  pivot_longer(!religion, names_to = "income", values_to = "count", cols_vary = "slowest")
#> # A tibble: 180 × 3
#>    religion                income count
#>    <chr>                   <chr>  <dbl>
#>  1 Agnostic                <$10k     27
[...]

PR #1353, opened 2022-04-19
Adjust dots in pivoting functions

### LINKED ISSUE #1350: Consider moving the dots in `pivot_wider()` and `pivot_longer()`
We'd move them up front, like:

```r
pivot_longer(
  data,
  cols,
  ...,
  cols_vary = "fastest",
  names_to = "name",
  names_prefix = NULL,
  names_sep = NULL,
  names_pattern = NULL,
  names_ptypes = NULL,
  names_transform = NULL,
  names_repair = "check_unique",
  values_to = "value",
  values_drop_na = FALSE,
  values_ptypes = NULL,
  values_transform = NULL
)

pivot_wider(
  data,
  ...,
  id_cols = NULL,
  id_expand = FALSE,
  names_from = name,
  names_prefix = "",
  names_sep = "_",
  names_glue = NULL,
  names_sort = FALSE,
  names_vary = "fastest",
  names_expand = FALSE,
  names_repair = "check_unique",
  values_from = value,
  values_fill = NULL,
  values_fn = NULL,
  unused_fn = NULL
)
```

This has two big benefits:

- Allows us to add more optional arguments to the generic without fear of breaking packages that have written methods for these generics. As long as the optional arguments come after the `...` in the generic, then there won't be a CRAN check warning.

- Forces users to specify the optional arguments by name. This is nice because we can add new optional arguments in any order without fear of breaking any positional arguments. This is particularly important for package code, where reliance on positional argument ordering can cause a CRAN failure

See https://github.com/tidyverse/tidyr/pull/1347#issuecomment-1095147941 for more details and some motivation

This would be a breaking change for anyone who is currently relying on positional arguments. But the `cols_vary` addition is _already_ a breaking change for the same reason, so it seems like a nice time to change this too.

DavisVaughan:
Closes #1350 

- `pivot_longer()` and `pivot_wider()` dots have been moved to the front, after the required args but before any optional ones. `check_dots_used()` ensures that typos are caught.
- `pivot_longer_spec()`, `pivot_wider_spec()`, `build_longer_spec()`, and `build_wider_spec()` have all gained dots that are also placed up front. `check_dots_empty()` ensures that typos are caught.

This should make it much easier to extend these functions going forward. I see two big benefits:
- For developers: it should mean we can add arguments to the generic after the `...` and not automatically break any packages that have added S3 methods for these functions
- For users: it forces them to specify arguments by name, creating more robust and readable code. We will be able to add new arguments in any order we prefer and not have to worry about positional argument issues.

I'll run revdeps, I'm hoping it clears up the S3 related issues from here https://github.com/tidyverse/tidyr/pull/1347#issuecomment-1095147941

### DIFF
--- a/R/pivot-long.R
+++ b/R/pivot-long.R
@@ -140,6 +140,7 @@
 #'   )
 pivot_longer <- function(data,
                          cols,
+                         ...,
                          cols_vary = "fastest",
                          names_to = "name",
                          names_prefix = NULL,
@@ -151,8 +152,7 @@ pivot_longer <- function(data,
                          values_to = "value",
                          values_drop_na = FALSE,
                          values_ptypes = NULL,
-                         values_transform = NULL,
-                         ...) {
+                         values_transform = NULL) {
   check_dots_used()
   UseMethod("pivot_longer")
 }
@@ -160,6 +160,7 @@ pivot_longer <- function(data,
 #' @export
 pivot_longer.data.frame <- function(data,
                                     cols,
+                                    ...,
                                     cols_vary = "fastest",
                                     names_to = "name",
                                     names_prefix = NULL,
@@ -171,8 +172,7 @@ pivot_longer.data.frame <- function(data,
                                     values_to = "value",
                                     values_drop_na = FALSE,
                                     values_ptypes = NULL,
-                                    values_transform = NULL,
-                                    ...) {
+                                    values_transform = NULL) {
   spec <- build_longer_spec(
     data = data,
     cols = {{ cols }},
@@ -204,6 +204,7 @@ pivot_longer.data.frame <- function(data,
 #'
 #' @keywords internal
 #' @export
+#' @inheritParams rlang::args_dots_empty
 #' @inheritParams pivot_longer
 #' @param spec A specification data frame. This is useful for more complex
 #'  pivots because it gives you greater control on how metadata stored in the
@@ -238,11 +239,14 @@ pivot_longer.data.frame <- function(data,
 #' )
 pivot_longer_spec <- function(data,
                               spec,
+                              ...,
                               cols_vary = "fastest",
                               names_repair = "check_unique",
                               values_drop_na = FALSE,
                               values_ptypes = NULL,
                               values_transform = NULL) {
+  check_dots_empty0(...)
+
   spec <- check_pivot_spec(spec)
   spec <- deduplicate_spec(spec, data)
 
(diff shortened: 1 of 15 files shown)


DavisVaughan:
Breaks 11 packages, but does fix the 3 packages that had S3 consistency issues, which is good news.

The 11 breakages are mainly due to people doing `pivot_wider(df, id, <more named args>)`, i.e. they weren't specifying the `id_cols` argument by name, which is a little unfortunate.

I still think it is worth making patches for those packages and pushing through with this, they should be easy fixes. If we do this, future maintenance should be much easier and less prone to breaking packages.

DavisVaughan:
Maybe look for a fallback idea?

Check for a single unnamed argument to see if `id_cols` is being used without being named. Throw a warning and make that work anyways as a temporary fallback

DavisVaughan:
Revdeps of compat behavior running at `"04115383-fc69-415d-b648-4ffc8d9fb4c0"`

DavisVaughan:
Revdep summary looks good. @hadley I think this is ready for a review now.

Only 1 package has an issue related to this PR, {faux}.

## `pivot_longer()` dots
- faux needs to name args in `pivot_longer()` in the `continuous.Rmd` vignette

## `allow_rename = FALSE` tidy selection

- ggpubr needs to unname tidy selection in `...` before call to `unite()`
- metaconfoundr needs to unname tidy selection before call to `pivot_longer()`
- recipes needs to unname tidy selection before call to `drop_na()`

## stringr

- cmcR stringr failure
- mpwR stringr failure
- tidyfst stringr failure

## False alarms
- openalexR can't reproduce locally, my guess is there was a download failure

R/pivot-long.R
```diff
@@ -140,6 +140,7 @@
 #'   )
 pivot_longer <- function(data,
                          cols,
+                         ...,
```
DavisVaughan, line 143: In the `pivot_longer/wider()` docs these dots are documented as "passed on to methods"

But in the `build_*()` functions they must be empty

R/pivot-wide.R
```diff
@@ -608,6 +620,52 @@ stop_id_cols_oob <- function(i, arg, call) {
...
+
+  lifecycle::deprecate_warn(
+    when = "1.3.0",
+    what = I(cli::format_inline(
+      "Specifying the {.arg id_cols} argument by position"
+    )),
+    details = cli::format_inline(
+      "Please explicitly name {.arg id_cols}, like {.code id_cols = {id_cols}}."
+    ),
+    always = TRUE,
```
DavisVaughan, line 664: Fairly aggressive about this compatibility deprecation warning, which felt alright to me since it is an easy fix on the user side and we'd like to remove this as soon as we reasonably can

hadley (approved):

PR #1358, opened 2022-04-29
Make `uncount()` generic

mgirlich:
I guess that `uncount()` is more stable than the other functions mentioned in #1071. This would currently be useful for dtplyr  https://github.com/tidyverse/dtplyr/issues/352
@DavisVaughan Are you fine with making `uncount()` generic or do you think the interface is not yet stable (enough)?


R/uncount.R
```diff
@@ -21,7 +22,14 @@
...
 #' # Or expressions
 #' uncount(df, 2 / n)
-uncount <- function(data, weights, .remove = TRUE, .id = NULL) {
+uncount <- function(data, weights, ..., .remove = TRUE, .id = NULL) {
+  UseMethod("uncount")
+}
+
+#' @export
+uncount.data.frame <- function(data, weights, ..., .remove = TRUE, .id = NULL) {
+  check_dots_used()
```
DavisVaughan, line 31:
```suggestion
  check_dots_empty()
```

Since at this point we are checking for typos and don't expect anything in the dots

DavisVaughan (commented):
Looks good, I am fine with making `uncount()` generic, as it does seem more stable than the other functions mentioned in #1071, as you said. @hadley should weigh in about this though.

I agree with the placement of the dots. I am running revdeps now and will report back with the results. Hopefully no one was relying on positional args for `.remove` or `.id` here.

Could you also please add two NEWS bullets? One that mentions that `...` have been added between the required and optional arguments, and one that mentions that `uncount()` has become generic. Both should reference this PR.

R/uncount.R
hadley: You can't, in general, use `check_dots_empty()` in a method because it's possible that the method could have been called from another method that had additional arguments (i.e. via `NextMethod()`)

hadley (approved):
Seems fine to me.

R/uncount.R
DavisVaughan: Same problem with `check_dots_used()` in a method

``` r
library(rlang)

foo <- function(x, ...) {
  UseMethod("foo")
}
foo.integer <- function(x, ...) {
  check_dots_used()
  x
}
foo.subclass <- function(x, ..., optional = FALSE) {
  force(optional)
  NextMethod()
}

foo(1L)
#> [1] 1
foo(structure(1L, class = c("subclass", "integer")), optional = TRUE)
#> Error:
#> ! Arguments in `...` must be used.
#> ✖ Problematic argument:
#> • optional = optional
```

<sup>Created on 2022-05-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

DavisVaughan: And putting `check_dots_used()` in the generic has the issue that if a method adds a new tidyselect argument, then it automatically fails because it won't look "used" because of how `enquo()` works (i.e. this issue https://github.com/r-lib/rlang/issues/1251#issuecomment-901874849)

``` r
library(rlang)

foo <- function(x, ...) {
  check_dots_used()
  UseMethod("foo")
}
foo.data.frame <- function(x, ..., custom) {
  tidyselect::eval_select(enquo(custom), x)
}

foo(mtcars, custom = cyl)
#> Error:
#> ! Arguments in `...` must be used.
#> ✖ Problematic argument:
#> • custom = cyl
```

But hopefully that won't be the case for `uncount()`? So @mgirlich I think you can just move `check_dots_used()` into the generic and that should be good enough.

DavisVaughan:
Revdepcheck has no changes from this PR!

R/uncount.R
mgirlich: It seems unlikely to me that we need a new tidyselect argument for `uncount()`, so I moved `check_dots_used()` into the generic.

DavisVaughan:
Thanks @mgirlich !

PR #1359, opened 2022-05-09
add kevin as contributor

kevinushey:
Mainly to recognize the contribution via https://github.com/tidyverse/tidyr/blob/main/src/melt.cpp (originally pulled from https://github.com/hadley/reshape/blob/master/src/melt.cpp).


hadley (approved):

DavisVaughan:
Thanks @kevinushey!

PR #1372, opened 2022-06-10
Make pack() and unpack() generic

krlmlr:
Useful now that duckdb supports them with the DBI interface: https://github.com/duckdb/duckdb/pull/3495.


krlmlr:
I wonder if we should instead define new generics `build_pack_spec()` and `pack_spec()`, similarly for "unpack". The DuckDB implementation is copying a bit of code from tidyr: https://github.com/duckdb/duckdb/pull/3829.

olivroy:
I may be mistaken, but I think this would help `separate_wider_delim()` / `separate_wider_regex()` work with sf objects. Is there still plans to go in that direction? (https://github.com/r-spatial/sf/issues/2359)

PR #1373, opened 2022-06-14
Update spread.R

mdkrause:
Hello,
One should load tidyr and not dplyr.
Thank you

### DIFF
--- a/R/spread.R
+++ b/R/spread.R
@@ -35,7 +35,7 @@
 #'   by `"<key_name><sep><key_value>"`.
 #' @export
 #' @examples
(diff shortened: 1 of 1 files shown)


DavisVaughan:
We typically assume the user has called `library(tidyr)` already when writing examples for tidyr.

It is true that the `library(dplyr)` call isn't needed, it doesn't look like it is used for anything in the examples, but `spread()` and `gather()` are superseded and aren't under active development anymore, even for simple doc improvements like that, so I'm going to close this one.

Thank you for the PR though!

PR #1375, opened 2022-06-21
Rectangling reorganisation

hadley:
Now each function gets its own file and its own documentation.

The documentation (especially the examples) needs more work; I just did the minimal to disentangle.


R/unnest.R
```diff
@@ -0,0 +1,186 @@
+#' Unnest a 2d list-column into rows and columns
+#'
+#' Unnest expands a list-column containing 2d data structures (e.g. data frames)
+#' into rows and columns.
+#'
+#' @inheritParams unchop
+#' @inheritParams unpack
```
DavisVaughan, line 7: If you look at the `?unnest` page, the `...` docs are actually coming from `pack()`

I think you want to mention that `...` are deprecated, maybe you can pull from the current dots of `nest()`

R/unnest.R
```diff
@@ -0,0 +1,186 @@
...
+#'     tibble(a = 1, b = 2),
+#'     tibble(a = 1:3, b = 3:1, c = 4)
+#'   )
+#' )
+#' # unnest() recycles input rows for each row of the list-column
+#' # and adds a column for each column
+#' df %>% unnest(y)
+#'
+#' # input rows with 0 rows in the list-column will usually disappear,
+#' # but you keep (generating NAs) with keep_empty = TRUE:
```
DavisVaughan, line 42:
```suggestion
#' # but you can keep them (generating NAs) with `keep_empty = TRUE`:
```

R/unnest-longer.R
```diff
@@ -0,0 +1,286 @@
+#' Unnest a list-column into rows
+#'
+#' @description
+#' `unnest_longer()` turns each element of a list-column into a row. It
+#' is most naturally suited to list-columns where the elements are unnamed
+#' and the length varies from row-to-row.
```
DavisVaughan, line 6:
```suggestion
#' and the length of each element varies from row-to-row.
```

I felt like this small bit of disambiguation helped with how I read this sentence

R/unnest-longer.R
```diff
@@ -0,0 +1,286 @@
...
+#'   `"_id"`.
+#' @param indices_include A single logical value specifying whether or not to
+#'   add an index column. If any value has inner names, the index column will be
+#'   a character vector of those names, otherwise it will be an integer vector
+#'   of positions. If `NULL`, defaults to `TRUE` if any value has inner names
+#'   or if `indices_to` is provided.
+#'
+#'   If `indices_to` is provided, then `indices_include` must not be `FALSE`.
+#' @inheritParams hoist
+#' @inheritParams unnest
```
DavisVaughan, line 34: Would you mind putting these above the new params you document?

R/unnest-wider.R
```diff
@@ -0,0 +1,238 @@
+#' Unnest a list-column into columns
+#'
+#' @description
+#' `unnest_wider()` turns each element of a list-column into a column. It
+#' is most naturally suited to list-columns where every element is named,
+#' and the names are consistent from row-to-row.
```
DavisVaughan, line 6: Having these "it is most naturally suited for..." in both `unnest_longer()` and `unnest_wider()` is really nice. The descriptions here make it pretty clear how to decide between the two

R/unnest-wider.R
```diff
@@ -0,0 +1,238 @@
...
+#'   inner names will be automatically generated as an increasing sequence of
+#'   integers.
+#' @param strict A single logical specifying whether or not to apply strict
+#'   vctrs typing rules. If `FALSE`, typed empty values (like `list()` or
+#'   `integer()`) nested within list-columns will be treated like `NULL` and
+#'   will not contribute to the type of the unnested column. This is useful
+#'   when working with JSON, where empty values tend to lose their type
+#'   information and show up as `list()`.
+#' @family rectangling
+#' @inheritParams unnest_longer
```
DavisVaughan, line 26: Same comments about the tag ordering, in this case for both `@inheritParams` and `@export`.

I'm not sure if you really have any strong feelings about where these tags are generally placed, but in my head I have some rough expected order of `inheritParams -> params -> export -> examples`

R/unnest-wider.R
```diff
@@ -0,0 +1,238 @@
...
+#'
+#' # You can also widen unnamed list-cols:
+#' df <- tibble(
+#'   x = 1:3,
+#'   y = list(NULL, 1:3, 4:5)
+#' )
+#' # where you'll usually want to provide names_sep:
+#' df %>% unnest_wider(y, names_sep = "_")
+#'
+#' #' # The defaults of `unnest_wider()` treat empty types (like `list()`) as `NULL`.
```
DavisVaughan, line 64:
```suggestion
#' # The defaults of `unnest_wider()` treat empty types (like `list()`) as `NULL`.
```

tests/testthat/test-unnest.R
```diff
@@ -0,0 +1,289 @@
...
+  df <- dplyr::group_by(tibble(x = c(1, 1, 1), y = 1:3), x)
+  expect_snapshot(out <- nest(df, .key = "y"))
+  expect_named(out, c("x", "y"))
+})
+
+test_that(".key gets warning with new interface", {
+  df <- tibble(x = c(1, 1, 1), y = 1:3)
+  expect_snapshot(out <- nest(df, y = y, .key = "y"))
+  expect_named(df, c("x", "y"))
+})
```
DavisVaughan, line 247: This chunk should stay with `nest()` tests + regenerate snapshots in this file and in `test-nest.R`

DavisVaughan (commented):
I know you said that the docs probably need another pass, so feel free to skip some of these comments until then. But I figured I would go ahead and point out some issues so you have something to refer back to if you make more changes in another PR

R/unnest.R
```diff
@@ -0,0 +1,186 @@
+#' Unnest a 2d list-column into rows and columns
```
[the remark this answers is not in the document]
DavisVaughan: Might also want to update the Description section if you use this

R/unnest-wider.R
hadley: As you can tell, I just stick them anywhere 😁

hadley (commented):
That feedback was super useful, thanks!

PR #1405, opened 2022-10-18
Use allow_rename = FALSE

### LINKED ISSUE #1104: Error on named tidyselections in `pivot_longer()` and `pivot_wider()`
As the `cols` argument supports the tidy-select syntax I thought I could use it to rename a column. But this didn't work as expected and when using a named vector in `cols` we get:

* a unclear error message when using a data frame
* a better error message when using a data frame
* a clearly unwanted result when the name actually exists as a column

I think it would be nice to allow renaming via `cols`

``` r
library(tidyr)

# unclear error message for data frame
data.frame(x = 1, y = 2) %>% 
  pivot_longer(c(z = y))
#> Error in `[.data.frame`(data, cols): undefined columns selected

# better for tibble
tibble(x = 1, y = 2) %>% 
  pivot_longer(c(z = y))
#> Error: Can't subset columns that don't exist.
#> x Column `z` doesn't exist.

# when the name actually exists we get something quite unwanted
data.frame(x = 1, y = 2, z = 3) %>% 
  pivot_longer(c(z = y))
#> # A tibble: 1 x 4
#>       x     y name  value
#>   <dbl> <dbl> <chr> <dbl>
#> 1     1     2 z         3
```

<sup>Created on 2021-03-08 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

hadley:
Fixes #1104

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # tidyr (development version)
 
+* `chop()`, `drop_na()`, `fill()`, `pivot_longer()`, `pivot_wider()`, 
+  `separate_rows()`, `unite()`, `unnest_longer()`, and `unnest_wider()` now
+  error if you supply named selection (#1104).
+
 * `pivot_longer()` now throws a slightly better error message when
   `values_ptypes` or `names_ptypes` is provided and the coercion can't be made
   (#1364).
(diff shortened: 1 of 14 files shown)


DavisVaughan (approved):
Maybe a test for `pivot_longer()` and `pivot_wider()`? They seem to be the biggest culprits where this was coming up

PR #1412, opened 2022-10-20
Use the same sectioning approach everywhere

jennybc:
If we're going to use this as the poster child of example sectioning in R Packages, let's make it pretty :nail_care:

### DIFF
--- a/R/chop.R
+++ b/R/chop.R
@@ -40,7 +40,7 @@
 #'   can be supplied, which will be applied to all `cols`.
 #' @export
 #' @examples
-#' # Chop ======================================================================
+#' # Chop ----------------------------------------------------------------------
 #' df <- tibble(x = c(1, 1, 1, 2, 2, 3), y = 1:6, z = 6:1)
 #' # Note that we get one row of output for each unique combination of
 #' # non-chopped variables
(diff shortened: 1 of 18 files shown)


hadley:
But `=` instead of `-`? 😭 

(I'll fix the build failures)

jennybc:
I was just going with what was already here and what was already in the book.

https://github.com/tidyverse/tidyr/blob/4e59e04bdf69e64edc4ab4c46db18cb28fa3c1da/R/chop.R#L43-L61

https://github.com/hadley/r-pkgs/blob/218933dc33aa4fc883997a62cc488038269e5ea2/man.Rmd#L491-L513

I don't care either way. But I've changed to `-`. Do you want me to remove the mention of `=` from the book as well?

hadley:
Yes please, I definitely prefer `-`!

hadley:
I added section dividers in more places so there are better examples to draw upon than `chop()` (which is an esoteric function and a rare user of `====`).

* `complete()` got a drive by update to use the pipe
* all section dividers now stop at 80 characters

Let me know what you think!

jennybc:
> the `=` is the h1 and there `-` is the h2, but I don't usually use `=` unless there's a second level

I'm somewhat confused by this. So you're saying that, when we only have one level of sectioning, we use `-`. But if there are two levels, we use `=` at the h1 level and `-` at the h2 level. Do I have that right? I suppose this is because `=` is "bigger" than `-`.

If I'm following the logic, I think it's not worth it :grin:. I vote we always section with `-` and bring out `=` for the rare second level of sectioning.

(I also don't really care deeply about this. My only goal is to make sure whatever we feature in R Packages isn't weird or puzzling.)

hadley:
Agreed. I've simplified.

PR #1419, opened 2022-11-11
Fix `pivotting` -> `pivoting` + misc typos

mattroumaya:
Just caught this while reading the docs. Thank you!


DavisVaughan:
Pivotting is just the British English form of pivoting, so it isn't technically a typo. But we were pretty inconsistent about which we used so it seems reasonable to just standardize on one of them. Thanks!

Issue #1426, opened 2022-11-23
`pivot_wider()` is quite slow when `names_sep` is provided


mgirlich:
Most of the time is spent in `apply_names_sep()` because it uses `glue()`:

<img width="822" alt="image" src="https://user-images.githubusercontent.com/18258854/203495574-fb7c11fc-680b-43d2-8f4f-7bc2f061c9d0.png">

PR #1427, opened 2022-11-23
Speed up `pivot_wider(names_sep = ...)`

mgirlich:
Closes https://github.com/tidyverse/tidyr/issues/1426

I think the only reason to use `glue()` was the handling in the case of empty input. This is already covered in tests (`names_sep` works with empty elements (#1185)`).

``` r
library(tidyr)

df <- tibble(
  id = 1:10e3,
  x = list(1:3)
)

bench::mark(
  unnest_wider = unnest_wider(df, x, names_sep = "_")
)

# Before
#> # A tibble: 1 × 6
#>   expression        min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 unnest_wider    1.79s    1.79s     0.558    3.66MB     22.3

# After
#> # A tibble: 1 × 6
#>   expression        min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 unnest_wider    207ms    227ms      4.45    3.66MB     20.8
```

<sup>Created on 2022-11-23 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # tidyr (development version)
 
+* `pivot_wider()` is now faster when `names_sep` is provided (@mgirlich, #1426).
+
 * The `...` argument of both `pivot_longer()` and `pivot_wider()` has been
   moved to the front of the function signature, after the required arguments
   but before the optional ones. Additionally, `pivot_longer_spec()`,
--- a/R/utils.R
+++ b/R/utils.R
@@ -182,7 +182,11 @@ list_of_ptype <- function(x) {
 }
 
 apply_names_sep <- function(outer, inner, names_sep) {
-  as.character(glue("{outer}{names_sep}{inner}"))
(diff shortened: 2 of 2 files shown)


DavisVaughan:
Would you mind trying `tidyr:::vec_paste0()`? Is that too slow?

mgirlich:
Much better than before but still quite expensive

```
# A tibble: 1 × 6
 expression        min   median `itr/sec` mem_alloc `gc/sec`
 <bch:expr>   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
1 unnest_wider    288ms    318ms      3.15    6.18MB     17.3
```

Issue #1422, opened 2022-11-13
Require purrr 1.0.0 then unskip this `hoist()` test


DavisVaughan:
It skips on CRAN, but will start failing for us locally and on CI when purrr 1.0.0 is installed because that issue is fixed now. Should revisit that old issue to see if anything else needs to be updated.

https://github.com/tidyverse/tidyr/blob/91747952f10c961be747c0de1026d109c920e4fc/tests/testthat/test-hoist.R#L139-L155

Issue #1203, opened 2021-11-08
`hoist()` isn't type stable due to `purrr::pluck()` behavior with empty elements


DavisVaughan:
```r
library(tidyr)

col <- list(
  list(a = integer(), b = 1),
  list(a = integer(), b = 2)
)

df <- tibble(col = col)

# Resulting `a` column should be typed as <integer>
hoist(df, col, "a", "b")
#> # A tibble: 2 × 2
#>   a         b
#>   <lgl> <dbl>
#> 1 NA        1
#> 2 NA        2

# Partly because this doesn't return `integer()`
purrr::pluck(col, 1, "a")
#> NULL
```

See https://github.com/tidyverse/tidyr/issues/1044#issuecomment-954882201 for more information

I don't think there is anything we can do about this without a change in purrr.

hadley:
With dev purrr:

``` r
library(tidyr)

col <- list(
  list(a = integer(), b = 1),
  list(a = integer(), b = 2)
)

df <- tibble(col = col)

# Resulting `a` column should be typed as <integer>
hoist(df, col, "a", "b")
#> # A tibble: 2 × 2
#>       a     b
#>   <int> <dbl>
#> 1    NA     1
#> 2    NA     2
```

<sup>Created on 2022-10-18 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

DavisVaughan:
The bottom half of my comment here suggests we need to do something else too 
https://github.com/tidyverse/tidyr/issues/1044#issuecomment-954882201

DavisVaughan:
Oh it looks like I added `list_init_empty()` after that comment, and that fixes the suggestion I made in that comment

PR #1430, opened 2022-12-20
Updates for purrr/stringr/dplyr

DavisVaughan:
Closes https://github.com/tidyverse/tidyr/issues/1422

Updated and activated a test related to https://github.com/tidyverse/tidyr/issues/1203 now that `pluck()` retains empty elements correctly.

Updated snapshots, but these seem to be a little backwards. Really `unnest_wider()` calls `map()`, but the `.f` we call `map()` with accepts an `error_call` argument and that `error_call` points to `unnest_wider()` as the function to use in the errors. So it ends up looking like `map()` calls `unnest_wider()`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -28,7 +28,7 @@ Imports:
     magrittr,
     purrr (>= 1.0.0),
     rlang (>= 1.0.4),
-    stringr (>= 1.4.0.9000),
+    stringr (>= 1.5.0),
     tibble (>= 2.1.1),
     tidyselect (>= 1.2.0),
     utils,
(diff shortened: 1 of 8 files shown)


hadley:
Instead of passing `error_call` to `elt_to_wide()` maybe we should wrap the error?

hadley:
i.e. something like this very crude attempt:

```R
tryCatch(
    out <- map(out, \(x) elt_to_wide(x, name = name, strict = strict, names_sep = names_sep, error_call = NULL)),
    error = function(e) cli::cli_abort(
      "failed",
      parent = e,
      call = error_call
    )
  )
```

hadley:
I've co-opted this PR to include updates for stringr and dplyr too.

hadley:
Hmmmm, one simple fix for the error problem is to switch from `map()` to `lapply()`, which I've done here. Maybe we will need some way to turn off the error wrapping in purrr?

PR #1435, opened 2023-01-02
Replace `any(map_lgl(...))`

mgirlich:
I stumbled over the pattern `any(map_lgl(...))` and thought there must be something better. And with `every()`, `some()` and `none()` there is something nicer in `purrr`. Also this is a bit more efficient in some cases.

### DIFF
--- a/R/unnest-helper.R
+++ b/R/unnest-helper.R
@@ -90,7 +90,7 @@ col_simplify <- function(x,
 
   # Don't try and simplify non-vectors. list-of types always contain vectors.
   if (!is_list_of(x)) {
-    has_non_vector <- any(!map_lgl(x, ~ vec_is(.x) || is.null(.x)))
+    has_non_vector <- !all(map_lgl(x, ~ vec_is(.x) || is.null(.x)))
 
     if (has_non_vector) {
       return(x)
--- a/R/unnest-helper.R
+++ b/R/unnest-helper.R
@@ -82,7 +82,7 @@ col_simplify <- function(x,
   if (is_list_of(x)) {
     has_list_of_list <- vec_is_list(list_of_ptype(x))
   } else {
-    has_list_of_list <- some(x, vec_is_list)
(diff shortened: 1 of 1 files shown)


DavisVaughan:
I think `some()` and `every()` are useful if you expect to use their "early exit" feature fairly often, otherwise they are slower than `map_lgl()`'s implementation because it calls the `.f` in a C loop.

In this case, I don't expect to use the early exit very often. We:
- Don't typically expect list of lists
- Don't typically expect to have any non-vector types

So I think keeping the wrapper around `map_lgl()` will probably end up being faster in the average case

``` r
library(purrr)
library(vctrs)

x <- as.list(1:10000)

fn <- function(x) {
  vec_is(x) || is.null(x)
}
bench::mark(
  all(map_lgl(x, fn)),
  every(x, fn)
)
#> # A tibble: 2 × 6
#>   expression               min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>          <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 all(map_lgl(x, fn))   15.9ms   20.8ms      47.7    3.91MB     30.7
#> 2 every(x, fn)          42.6ms   43.1ms      23.2  103.95KB    104.

bench::mark(
  any(map_lgl(x, vec_is_list)),
  some(x, vec_is_list)
)
#> # A tibble: 2 × 6
#>   expression                        min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 any(map_lgl(x, vec_is_list))   4.63ms   5.55ms     180.    41.31KB     24.0
#> 2 some(x, vec_is_list)          26.32ms  27.23ms      36.4    8.02KB     57.2
```

I will change from `any(!fn())` to `!all(fn())` though, as that is more efficient

mgirlich:
Good catch! I didn't expect them to be much slower.

PR #1438, opened 2023-01-03
Document grouped data frame support

### LINKED ISSUE #952: Document which functions work with grouped data
On the `?fill` help page, there is no statement that `fill` knows to respect groups from `dplyr::group_by`. (The last example does demonstrate this.) Same on the `?expand` and `?complete` help pages, possibly other functions too. It would be great if there was a standard sentence or phrase that was part of the `Description` section for all non-`dplyr` functions that know what to do with a `grouped_df`. Perhaps "Respects tibble groupings".

I think this would both serve as a good reference (I constructed a little example before posting thing to check on `expand`... I was pretty sure it respected groups but needed to check to see), and also for those new users who read help pages it might help cement the concept that `group_by` isn't magic that works with every function, only specially constructed functions.

DavisVaughan:
Closes #952

There are 4 functions that support grouped data frames in a special way (beyond just reconstruction on the way out)
- `nest()` (method)
- `expand()` (method)
- `complete()` (method)
- `fill()` (`mutate_at()`)

`nest()`, `expand()`, and `complete()` all already documented their grouped df support, having both examples and a section specifically about grouped data frames.

I've standardized all of the functions to have a `Grouped data frames:` section, repurposing existing bits as appropriate (`nest()` already had one and has no changes)

### DIFF
--- a/R/fill.R
+++ b/R/fill.R
@@ -6,7 +6,12 @@
 #'
 #' Missing values are replaced in atomic vectors; `NULL`s are replaced in lists.
 #'
-#' @inheritParams gather
+#' @section Grouped data frames:
+#' With grouped data frames created by [dplyr::group_by()], `fill()` will be
+#' applied _within_ each group, meaning that it won't fill across group
+#' boundaries.
+#'
+#' @param data A data frame.
 #' @param ... <[`tidy-select`][tidyr_tidy_select]> Columns to fill.
 #' @param .direction Direction in which to fill missing values. Currently
 #'   either "down" (the default), "up", "downup" (i.e. first down and then up)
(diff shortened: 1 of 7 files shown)


R/fill.R
```diff
@@ -6,7 +6,12 @@
 #'
 #' Missing values are replaced in atomic vectors; `NULL`s are replaced in lists.
 #'
-#' @inheritParams gather
+#' @section Grouped data frames:
+#' With grouped data frames created by [dplyr::group_by()], `fill()` will be
+#' applied _within_ each group, meaning that it won't fill across group
+#' boundaries.
+#'
+#' @param data A data frame.
```
DavisVaughan, line 14: I also made this minor change to avoid inheriting `data` docs from the deprecated `gather()` function

hadley (approved):

Issue #1363, opened 2022-06-02
`unnest_longer()` uses a mix of `keep_empty = TRUE` and `keep_empty = FALSE`


DavisVaughan:
Currently it keeps `NULL` elements, but drops empty vectors like `integer()`.

It should work like `unchop(keep_empty = FALSE)` which drops both `NULL` and `integer()`.

This is heavily related to #1339. I imagine that to be able to add `keep_empty` to `unnest_longer()` we will have to resolve this. I think it requires changes to `tidyr:::elt_to_long()`'s handling of `NULL`. When `keep_empty = FALSE` (the default) it should probably use `unspecified(0L)` and an empty vector for the `index` if needed. The current behavior looks to be what should happen for `keep_empty = TRUE`. 

Fixing this is probably a small breaking change. But I think it is definitely worth it for package consistency.

``` r
library(tidyr)

df <- tibble(
  x = list(NULL, 1L, integer())
)

df
#> # A tibble: 3 × 1
#>   x        
#>   <list>   
#> 1 <NULL>   
#> 2 <int [1]>
#> 3 <int [0]>

# Inconsistent:
# Keeps the `NULL` as `NA`.
# Drops the `integer()`.
unnest_longer(df, x)
#> # A tibble: 2 × 1
#>       x
#>   <int>
#> 1    NA
#> 2     1

# Consistent with our definition of "empty".
# - Drops `NULL`
# - Drops `integer()` (size 0 vectors)
unchop(df, x)
#> # A tibble: 1 × 1
#>       x
#>   <int>
#> 1     1

# - Keeps `NULL` as NA
# - Keeps `integer()` as NA
unchop(df, x, keep_empty = TRUE)
#> # A tibble: 3 × 1
#>       x
#>   <int>
#> 1    NA
#> 2     1
#> 3    NA
```

<sup>Created on 2022-06-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Issue #1339, opened 2022-03-28
Should `unnest_longer()` have `keep_empty`?


DavisVaughan:
From https://community.rstudio.com/t/unnest-longer-drops-lists-rows-with-character-0/132748

Original example:

``` r
library(tidyverse)

my_df <- tibble(
  txt = c(
    "chestnut, pear, kiwi, peanut",
    "grapes, banana"
  )
)

#Extract all nuts
my_df <- my_df %>% 
  mutate(nuts = str_extract_all(txt, regex("\\w*nut\\w*"))) %>% 
  mutate(index = row_number(), .before=1)

#Row index 2 has nuts <chr [0]>
my_df
#> # A tibble: 2 × 3
#>   index txt                          nuts     
#>   <int> <chr>                        <list>   
#> 1     1 chestnut, pear, kiwi, peanut <chr [2]>
#> 2     2 grapes, banana               <chr [0]>

#unnest
my_df_long <- my_df %>% 
  unnest_longer(nuts, values_to = "nuts_long")

#Row index 2 is now missing
my_df_long
#> # A tibble: 2 × 3
#>   index txt                          nuts_long
#>   <int> <chr>                        <chr>    
#> 1     1 chestnut, pear, kiwi, peanut chestnut 
#> 2     1 chestnut, pear, kiwi, peanut peanut
```

<sup>Created on 2022-03-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Minimal reprex:

``` r
library(tidyverse)

df <- tibble(
  x = list("a", character())
)
df
#> # A tibble: 2 × 1
#>   x        
#>   <list>   
#> 1 <chr [1]>
#> 2 <chr [0]>

unnest_longer(df, x)
#> # A tibble: 1 × 1
#>   x    
#>   <chr>
#> 1 a

unnest(df, x, keep_empty = TRUE)
#> # A tibble: 2 × 1
#>   x    
#>   <chr>
#> 1 a    
#> 2 <NA>
```

<sup>Created on 2022-03-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

It may be as simple as passing `keep_empty` through to the `unchop()` call in `unnest_longer()`, but I'd need to think about it critically to make sure

hadley:
I just noticed this absence while writing for R4DS, so I definitely think we should add `keep_empty` support.

hadley:
[...]

Issue #1327, opened 2022-02-25
unnest_wider() fails when list columns have additional classes and missing values


allenbaron:
I recently upgraded from `1.1.4` to `1.2.0` and ran into an error with `unnest_wider()`. It now fails with a type conversion error when two occurrences overlap:

1. the list column to be unnested is composed of list(s) with at least one empty object (`NULL` or `list()`).
2. the list column to be unnested is composed of list(s) with an additional class.

``` r
library(tidyr)
#> Warning: package 'tidyr' was built under R version 4.1.2
library(tibble)

x <- list(a = 1, b = NULL)
df <- tibble::tibble(c = 1, d = list(x))
str(df)
#> tibble [1 × 2] (S3: tbl_df/tbl/data.frame)
#>  $ c: num 1
#>  $ d:List of 1
#>   ..$ :List of 2
#>   .. ..$ a: num 1
#>   .. ..$ b: NULL


# works
tidyr::unnest_wider(df, d)
#> # A tibble: 1 × 3
#>       c     a b    
#>   <dbl> <dbl> <lgl>
#> 1     1     1 NA


# does not work
class(x) <- c("fake", "list")
df <- tibble::tibble(c = 1, d = list(x))
tidyr::unnest_wider(df, d)
#> Error: Can't convert <list> to <fake>.


# works again
x$b <- 1
df <- tibble::tibble(c = 1, d = list(x))
tidyr::unnest_wider(df, d)
#> # A tibble: 1 × 3
#>       c     a     b
#>   <dbl> <dbl> <dbl>
#> 1     1     1     1
```

<sup>Created on 2022-02-25 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

<details style="margin-bottom:10px;">
<summary>
rlang::last_error()
</summary>

``` r
rlang::last_error()
#> <error/vctrs_error_incompatible_type>
#> Can't convert <list> to <fake>.
#> Backtrace:
#>   1. tidyr::unnest_wider(df, d)
#>   2. tidyr:::df_simplify(...)
#>   3. tidyr:::col_simplify(...)
#>   4. tidyr:::list_init_empty(x, null = TRUE, typed = TRUE)
#>   5. vctrs::vec_assign(x, empty_null, replacement)
#>   7. vctrs::vec_default_cast(...)
#>   8. vctrs::stop_incompatible_cast(...)
#>   9. vctrs::stop_incompatible_type(...)
[...]

PR #1442, opened 2023-01-04
Implement `keep_empty` for `unnest_longer()`

DavisVaughan:
Closes https://github.com/tidyverse/tidyr/issues/1363
Closes https://github.com/tidyverse/tidyr/issues/1339

We are going to use `keep_empty = FALSE` as the default to be consistent with `unnest()`, `unchop()`, `separate_longer_position()`, etc. This is a breaking change as previously `NULL` elements implicitly had `keep_empty = TRUE` applied to them. However, we are now consistent with how we handle `NULL` and empty vectors like `integer()`, which I think is important.


R/unnest-longer.R
```diff
@@ -20,38 +22,41 @@
 #'   column names. The default, `NULL`, gives the output columns the same names
 #'   as the input columns.
 #' @param indices_to A string giving the column name (or names) to store the
-#'   the inner names or positions (if not named) of the values. If multiple
```
DavisVaughan, line 23: I just fixed a typo of duplicated `the the` split over these two lines. No other changes besides a reflow of the comment

R/unnest-longer.R
```diff
@@ -194,21 +205,23 @@ col_to_long <- function(col,
...
-    x <- unspecified(1L)
+    x <- unspecified()
 
     if (indices_include) {
       if (is.integer(index)) {
-        index <- 1L
+        index <- integer()
       } else {
-        index <- NA_character_
+        index <- character()
```
DavisVaughan, line 224: General idea is to have 1 standard path we always apply to `NULL`, and then if `keep_empty = TRUE` then the other path below handles both the updated `NULL` result we created here and the empty vector case

R/unnest-longer.R
```diff
@@ -220,6 +233,21 @@ elt_to_long <- function(x,
...
+
+  if (keep_empty && size == 0L) {
+    x <- vec_init(x)
+    size <- 1L
+
+    if (indices_include) {
+      if (is.integer(index)) {
+        index <- 1L
+      } else {
+        index <- NA_character_
```
DavisVaughan, line 246: I am slightly questioning our decision to use `NA` when there aren't any names over `""`, like what `names2()` would return

revdep/cran.md
```diff
@@ -1,9 +1,9 @@
 ## revdepcheck results
 
-We checked 1733 reverse dependencies (1715 from CRAN + 18 from Bioconductor), comparing R CMD check results across CRAN and dev versions of this package.
+We checked 1741 reverse dependencies (1723 from CRAN + 18 from Bioconductor), comparing R CMD check results across CRAN and dev versions of this package.
 
- * We saw 6 new problems
- * We failed to check 11 packages
+ * We saw 4 new problems
```
DavisVaughan, line 5: 2 were false positives from before, so that's good

tests/testthat/test-unnest-longer.R
```diff
@@ -173,22 +256,39 @@ test_that("can unnest multiple columns (#740)", {
...
+  df <- tibble(a = list(NULL, 3:4), b = list(1:2, 4L))
+
+  expect_snapshot(error = TRUE, {
+    unnest_longer(df, c(a, b))
+  })
+
+  out <- unnest_longer(df, c(a, b), keep_empty = TRUE, indices_include = TRUE)
+
+  expect_identical(out$a, c(NA, NA, 3L, 4L))
+  expect_identical(out$a_id, c(1L, 1L, 1L, 2L))
```
hadley, line 275: These values confused me for a while because I though they must be out indexes; but they're inner indexes, some of which point to elements that don't exist. (`NULL[[1]]` is ok by coincidence, but `double()[[1]]` is not).

hadley (approved):

R/unnest-longer.R
DavisVaughan: I'm really referring to this

``` r
tibble(
  x = list(1:2, c(a = 3, b = 4))
) %>%
  unnest_longer(x)
#> # A tibble: 4 × 2
#>       x x_id 
#>   <dbl> <chr>
#> 1     1 <NA> 
#> 2     2 <NA> 
#> 3     3 a    
#> 4     4 b
```

DavisVaughan: Maybe we could change ^ to use `""`, and then consistently use `NA_integer_` or `NA_character_` when we promote empty vectors to size 1? So its like:

No names anywhere:
- Positional names for existing values
- `NA_integer_` if a zero length vector was promoted to size 1

At least one name somewhere:
- Elements with names use their names
- Elements without names use `""`
- `NA_character_` if a zero length vector was promoted to size 1

DavisVaughan: Using `""` for values without names is probably more consistent in this case

``` r
tibble(
  x = list(c(a = 1, 2), c(3, 4))
) %>%
  unnest_longer(x)
#> # A tibble: 4 × 2
#>       x x_id 
#>   <dbl> <chr>
#> 1     1 "a"  
#> 2     2 ""   
#> 3     3 <NA> 
#> 4     4 <NA>
```

hadley: I kind of like using `""` for unnamed elements and `NA` for elements that didn't exist. But I can see how that might be slicing things too fine.

DavisVaughan: Isn't that what I'm suggesting?

hadley: Like this case:

``` r
library(tidyr)
tibble(x = list(c(a = 1, 2), c(3, 4), NULL)) %>%
  unnest_longer(x, keep_empty = TRUE)
#> # A tibble: 5 × 2
#>       x x_id 
#>   <dbl> <chr>
#> 1     1 "a"  
#> 2     2 ""   
#> 3     3 <NA>  # should be ""
#> 4     4 <NA>  # should be ""
#> 5    NA <NA> # should be NA
```

<sup>Created on 2023-01-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

tests/testthat/test-unnest-longer.R
DavisVaughan: These are now `NA` instead of `1L`

DavisVaughan:
This PR grew a bit, but I realized that I could vectorize `col_to_long()` a bit more by:
- Moving all names handling to `collect_indices_info()`
- Adding `list_init_null()` to handle converting `NULL`s to either size 0 or size 1 vectors
- Adding `list_init_typed()` to follow that up and convert size 0 vectors to size 1 vectors

We already have `list_init_empty()` that tries to do both of these at once, and we use it in a few other places. It isn't quite right for our needs here, so I had to break it apart into its two functions.

After this PR I'll go back and remove `list_init_empty()` in favor of using these two helpers.

I'm also confident that when I do that I will be able to fix https://github.com/tidyverse/tidyr/issues/1327 based on the way I've implemented these two helpers.

---

I've also implemented the changes to the name handling we discussed here https://github.com/tidyverse/tidyr/pull/1442#discussion_r1062645387

R/unnest-longer.R
```diff
@@ -141,97 +149,89 @@ col_to_long <- function(col,
...
-      name = name,
-      values_to = values_to,
-      indices_to = indices_to,
-      indices_include = indices_include,
-      error_call = error_call
+  if (!list_all_vectors2(col)) {
+    cli::cli_abort(
+      "List-column {.var {name}} must contain only vectors or `NULL`.",
+      call = error_call
     )
```
DavisVaughan, line 169: Nicely able to pull this check out of the iteration helper

R/unnest-longer.R
```diff
@@ -141,97 +149,89 @@ col_to_long <- function(col,
...
-    }
+  info <- list_init_null(col, sizes, ptype = ptype, size = size_null)
+  col <- info$x
+  sizes <- info$sizes
+
+  if (keep_empty) {
+    info <- list_init_typed(col, sizes, ptype = ptype, size = 1L)
+    col <- info$x
+    sizes <- info$sizes
   }
```
DavisVaughan, line 190: These two vectorized helpers now handle most of what `elt_to_long()` was doing

R/unnest-longer.R
```diff
@@ -141,97 +149,89 @@ col_to_long <- function(col,
...
   } else {
-    out <- list(x)
-    names(out) <- values_to
+    name <- values_to
+    ptype <- new_long_frame(ptype, 0L, name)
+    col <- map2(
+      col,
+      sizes,
+      function(elt, size) new_long_frame(elt, size, name)
+    )
```
DavisVaughan, line 212: The `map2()` and `pmap()` paths are now the most expensive part of `unnest_longer()`. Even though the functions are fairly simple (wrap into a data frame), if `col` is long the repeated calls to `.f` just naturally take awhile.

R/unnest-longer.R
```diff
@@ -261,10 +259,20 @@ collect_indices_info <- function(col, indices_include) {
   } else {
     # Indices are requested, and some elements are named.
     # For any unnamed elements, generate `NA` indices.
-    indices[unnamed] <- map(sizes[unnamed], vec_rep, x = NA_character_)
+    indices[unnamed] <- map(sizes[unnamed], vec_rep, x = "")
```
DavisVaughan, line 262: Here is the change to `""` for fully unnamed vectors

R/unnest-longer.R
```diff
@@ -261,10 +259,20 @@ collect_indices_info <- function(col, indices_include) {
...
     index_ptype <- character()
   }
 
+  if (keep_empty) {
+    empty <- sizes == 0L
+
+    if (any(empty)) {
+      # `NULL` or typed empty elements get an `NA` index of the right type
+      index_empty <- vec_init(index_ptype)
+      indices[empty] <- list(index_empty)
```
DavisVaughan, line 272: And here is the change to consistently use `NA` of the right type for `NULL` or typed empty elements that we "keep" (rather than using a mix of `1L` and `NA_character_` somewhat inconsistently)

R/utils.R
```diff
@@ -171,6 +172,122 @@ list_init_empty <- function(x,
...
+#' @noRd
+list_init_null <- function(x, sizes, ..., ptype = NULL, size = 1L) {
+  check_dots_empty0(...)
+
+  if (!vec_is_list(x)) {
+    cli::cli_abort("`x` must be a list.", .internal = TRUE)
+  }
+  if (is_list_of(x)) {
+    cli::cli_abort("`x` can't be a list-of. Unclass first and provide `ptype`.", .internal = TRUE)
+  }
```
DavisVaughan, line 203: The restriction of "no list-of" isn't technically necessary, in particular the previous `list_init_empty()` helper tried to allow list-ofs and use their information. But this is part of what caused the failure in https://github.com/tidyverse/tidyr/issues/1327, and it is cleaner to extract out the `ptype` of list-ofs ahead of time, strip them down to bare lists, and pass all that through instead.

R/utils.R
```diff
@@ -171,6 +172,122 @@ list_init_empty <- function(x,
...
+#'
+#' @param size The size of the replacement value to fill with. Commonly either
+#'   `0L` or `1L`.
+#'
+#' @returns
+#' A named list holding `x` with `NULL`s replaced, and `sizes` updated to
+#' have the correct size where `NULL` elements were replaced.
+#'
+#' @noRd
+list_init_null <- function(x, sizes, ..., ptype = NULL, size = 1L) {
```
hadley, line 195: The name made me thing think would initialise a list of `NULL`s. Maybe `list_replace_null()`?

R/utils.R
hadley: This makes sense to me because `list_of()` can't contain `NULL`s, right?

R/utils.R
```diff
@@ -171,6 +172,122 @@ list_init_empty <- function(x,
...
+  if (!vec_is_list(x)) {
+    cli::cli_abort("`x` must be a list.", .internal = TRUE)
+  }
+  if (is_list_of(x)) {
+    cli::cli_abort("`x` can't be a list-of. Unclass first and provide `ptype`.", .internal = TRUE)
+  }
+
+  if (vec_any_missing(x)) {
+    null <- vec_detect_missing(x)
+    null <- which(null)
```
hadley, line 207: Can't you assume that `vec_assign()` will do this?

hadley (approved):

R/utils.R
DavisVaughan: They can actually (that's the missing value) but can't contain `unspecified()` (although that branch probably wouldn't be hit for list-ofs because you'd typically also be supplying `ptype`)

R/utils.R
DavisVaughan: It definitely will. But since I use `null` in two separate assignments below, it seemed slightly more efficient to `which()` it once here

Issue #943, opened 2020-04-24
Restore data frames generically


hadley:
See existing work in #812, and see below for a list of functions that we needed to consider, and some thoughts on what form of genericity is needed. Goal is to make sure that data frame extensions return reasonable results in the absence of specific methods (and to make sure all needed functions are generic so that they can be extended when needed).

```
chop, unchop
pack, unpack
nest, unnest

separate, extract = append_df
hoist = append_df

complete = full_join + replace_na
drop_na = dplyr_row_slice
separate_rows = str_split + unchop
uncount = dplyr_row_slice + optional column removal
replace_na = dplyr_col_modify
expand = dplyr_reconstruct

pivot_longer = dplyr_reconstruct
pivot_wider = dplyr_reconstruct

# don't need to update superseded functions
gather, spread 
nest_legacy, unnest_legacy
```

DavisVaughan:
Need to consider the sticky column case, like panelr.

Ideally we'd be like dplyr, and just forcibly make the assumption that `[` with 1 argument `i` is going to return a data frame with length `length(i)`. That would break packages like this

``` r
library(tidyr)
library(panelr)

data("WageData")
wages <- panel_data(WageData, id = id, wave = t)

wages
#> # Panel data:    4,165 × 14
#> # entities:      id [595]
#> # wave variable: t [1, 2, 3, ... (7 waves)]
#>    id        t   exp   wks   occ   ind south  smsa    ms   fem union    ed   blk
#>    <fct> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
#>  1 1         1     3    32     0     0     1     0     1     0     0     9     0
#>  2 1         2     4    43     0     0     1     0     1     0     0     9     0
#>  3 1         3     5    40     0     0     1     0     1     0     0     9     0
#>  4 1         4     6    39     0     0     1     0     1     0     0     9     0
#>  5 1         5     7    42     0     1     1     0     1     0     0     9     0
#>  6 1         6     8    35     0     1     1     0     1     0     0     9     0
#>  7 1         7     9    32     0     1     1     0     1     0     0     9     0
#>  8 2         1    30    34     1     0     0     0     1     0     0    11     0
#>  9 2         2    31    27     1     0     0     0     1     0     0    11     0
#> 10 2         3    32    33     1     1     0     0     1     0     1    11     0
[...]

PR #1445, opened 2023-01-05
Rewrite `df_append()` from first principles

### LINKED ISSUE #1316: Should `hoist()` be converting to a bare data frame, and then calling `reconstruct_tibble()`?
You can currently generate a corrupt grouped-df because of the way `append_df()` works here:

``` r
library(tidyverse)

df <- tibble(
  id = c("x", "y", "z"),
  data = list(
    tibble(a = 1:2, b = 2:3),
    tibble(a = 2:3, b = 3:4),
    tibble(a = 3:4, b = 4:5)
  )
)

gdf <- group_by(df, id)

hoist(df, data, "a")
#> # A tibble: 3 × 3
#>   id    a         data            
#>   <chr> <list>    <list>          
#> 1 x     <int [2]> <tibble [2 × 1]>
#> 2 y     <int [2]> <tibble [2 × 1]>
#> 3 z     <int [2]> <tibble [2 × 1]>

hoist(gdf, data, "a")
#> Error in `validate_grouped_df()`:
#> ! The `groups` attribute is not a data frame with its last column called `.rows`.
```

This is not a new bug, as tidyr 1.2.0 and 1.1.4 both show this issue.

(Side note that hoisting out a column followed by an `unchop()` or `unnest()` is a really neat pattern if you don't need everything from `data`)

### LINKED ISSUE #1444: `append_df()` is broken with duplicated names
Used by things like `hoist()`, `extract()`, and `separate()`

``` r
library(tidyr)

df <- tibble(
  x = 1:3, 
  y = list(list(x = 3, a = 2), list(x = 4, a = 3), list(x = 5, a = 4)), 
  z = 1:3
)

# Typically hoisted column is placed right before column it comes from
hoist(df, y, "a")
#> # A tibble: 3 × 4
#>       x     a y                    z
#>   <int> <dbl> <list>           <int>
#> 1     1     2 <named list [1]>     1
#> 2     2     3 <named list [1]>     2
#> 3     3     4 <named list [1]>     3

# It gets a little weird if the column already existed
hoist(df, y, "x")
#> # A tibble: 3 × 3
#>   y                    x     z
#>   <list>           <dbl> <int>
#> 1 <named list [1]>     3     1
#> 2 <named list [1]>     4     2
#> 3 <named list [1]>     5     3
```

``` r
library(tidyr)

df <- tibble(x = 1:3, y = c("a-b", "c-d", "e-f"), z = 1:3, w = 1:3)

# Typically new columns replace `col` or go right after it
separate(df, y, into = c("new1", "new2"))
#> # A tibble: 3 × 5
#>       x new1  new2      z     w
#>   <int> <chr> <chr> <int> <int>
#> 1     1 a     b         1     1
#> 2     2 c     d         2     2
#> 3     3 e     f         3     3
separate(df, y, into = c("new1", "new2"), remove = FALSE)
#> # A tibble: 3 × 6
#>       x y     new1  new2      z     w
#>   <int> <chr> <chr> <chr> <int> <int>
#> 1     1 a-b   a     b         1     1
#> 2     2 c-d   c     d         2     2
#> 3     3 e-f   e     f         3     3

# Things get weird if you are replacing a name that already exists
separate(df, y, into = c("x", "new"))
#> # A tibble: 3 × 4
#>       z x     new       w
#>   <int> <chr> <chr> <int>
#> 1     1 a     b         1
#> 2     2 c     d         2
#> 3     3 e     f         3
separate(df, y, into = c("x", "new"), remove = FALSE)
#> # A tibble: 3 × 5
#>   y         z x     new       w
#>   <chr> <int> <chr> <chr> <int>
#> 1 a-b       1 a     b         1
#> 2 c-d       2 c     d         2
#> 3 e-f       3 e     f         3
```

DavisVaughan:
Closes #1444 
Closes #1316

`"be829164-2ec7-49e6-a2b8-79c64ea128b4"`

### DIFF
--- a/R/hoist.R
+++ b/R/hoist.R
@@ -114,27 +114,27 @@ hoist <- function(.data,
   )
 
   # Place new columns before old column
-  out <- append_df(.data, cols, after = match(.col, names(.data)) - 1L)
+  out <- df_append(.data, cols, after = match(.col, names(.data)) - 1L)
 
-  if (!.remove) {
-    return(out)
-  }
+  if (.remove) {
+    x <- map(x, function(x) {
+      # rev() is sneaky hack assuming that most people will remove in
+      # numeric order, so this should avoid most order problems. A full
+      # resolution will be considerably more work.
+      for (plucker in rev(pluckers)) {
+        x <- strike(x, plucker)
+      }
+      x
+    })
 
-  x <- map(x, function(x) {
-    # rev() is sneaky hack assuming that most people will remove in
-    # numeric order, so this should avoid most order problems. A full
-    # resolution will be considerably more work.
-    for (plucker in rev(pluckers)) {
-      x <- strike(x, plucker)
+    if (every(x, is_empty)) {
+      x <- NULL
     }
-    x
-  })
-  if (every(x, is_empty)) {
-    x <- NULL
-  }
-  out[[.col]] <- x
 
-  out
+    out[[.col]] <- x
+  }
+
+  reconstruct_tibble(.data, out)
 }
 
 check_pluckers <- function(..., .call = caller_env()) {
--- a/R/append.R
+++ b/R/append.R
@@ -56,26 +56,3 @@ df_append <- function(x, y, after = NULL, remove = FALSE) {
 
   out
 }
-
-append_df <- function(x, y, after = length(x), remove = FALSE) {
-  if (is.character(after)) {
-    after <- match(after, dplyr::tbl_vars(x))
-  } else if (!is.integer(after)) {
-    cli::cli_abort("{.arg after} must be character or integer", .internal = TRUE)
-  }
-
-  # Replace duplicated variables
-  x_vars <- setdiff(names(x), names(y))
-  if (remove) {
-    x_vars <- setdiff(x_vars, names(x)[[after]])
-    after <- after - 1L
-  }
-
-  y <- append(x[x_vars], y, after = after)
-  structure(y, class = class(x), row.names = .row_names_info(x, 0L))
-}
-
-append_col <- function(x, col, name, after = length(x)) {
-  name <- enc2utf8(name)
-  append_df(x, set_names(list(col), name), after = after)
-}
--- a/R/append.R
+++ b/R/append.R
@@ -1,3 +1,62 @@
+#' Append new columns (`y`) to an existing data frame (`x`)
+#'
+#' @details
+#' If columns are duplicated between `x` and `y`, then `y` columns are
+#' silently preferred.
+#'
+#' @param x A data frame.
+#' @param y A named list of columns to append. Each column must be the same size
+#'   as `x`.
+#' @param after One of:
+#'   - `NULL` to place `y` at the end.
+#'   - A single column name from `x` to place `y` after.
+#'   - A single integer position (including `0L`) to place `y` after.
+#' @param remove Whether or not to remove the column corresponding to `after`
+#'   from `x`.
+#'
+#' @returns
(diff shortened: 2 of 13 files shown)


DavisVaughan:
I think there are 2 failures related to these changes, mentioned below. I still think we should proceed though.

---

mapme.biodiversity was relying on buggy behavior of the previous `append_df()` function with sf objects. Somehow if you have a tibble backed sf object and you pass it to `extract()` then you get duplicates of the column you extracted from? Very weird. This now returns a bare tibble with the right number of columns in the dev version.

```r
# - You have an sf that is backed by a tibble (not a data.frame!)
# - You `extract()` from it
# - Weirdness occurs

library(sf)
#> Linking to GEOS 3.10.2, GDAL 3.4.2, PROJ 8.2.1; sf_use_s2() is TRUE
library(tidyr)

pt1 = st_point(c(0,1))
pt2 = st_point(c(1,1))
d = tibble(a = 1:2, geom = st_sfc(pt1, pt2))
df = st_as_sf(d)
df
#> Simple feature collection with 2 features and 1 field
#> Geometry type: POINT
#> Dimension:     XY
#> Bounding box:  xmin: 0 ymin: 1 xmax: 1 ymax: 1
#> CRS:           NA
#> # A tibble: 2 × 2
#>       a    geom
#>   <int> <POINT>
#> 1     1   (0 1)
#> 2     2   (1 1)

# `geom` was supposed to be `remove = TRUE`d, but now you get two??
tidyr::extract(df, "geom", into = c("longitude", "latitude"), "\\((.*),(.*)\\)")
#> # A tibble: 2 × 5
#>       a    geom longitude latitude    geom
#>   <int> <POINT> <chr>     <chr>    <POINT>
#> 1     1   (0 1) 0         " 1"       (0 1)
#> 2     2   (1 1) 1         " 1"       (1 1)
```

---

One thing that is different about this new `df_append()` vs the original `append_df()` is that the new one makes no attempt to restore the type of the input. It always returns a bare data frame and expects our standard `reconstruct_tibble()` function to handle restoration afterwards.

`append_df()` tried to use `structure()` to add the class back on before it returned, but this is fairly fragile so I've removed that behavior.

`reconstruct_tibble()` can restore:
- To a grouped-df
- To a tibble
- Otherwise returns the output as is if the input wasn't one of the above types

This means `reconstruct_tibble()` in itself isn't fully generic in how it restores data frames (see #943).

All of that results in a failure in wpa, where they expect a `data.table` object passed through `extract()` (I think) to retain its `data.table` class. But since `df_append()` now returns a bare data frame and `reconstruct_tibble()` doesn't restore data.tables, we now end up with a bare data.frame on the way out of `extract()`

```r
df <- data.table::data.table(x = c("x", "y"))
extract(df, "x", "foo")
#>   foo
#> 1   x
#> 2   y
```

I'm not sure there is much we can robustly do about this until #943 is fully handled.

R/hoist.R
```diff
@@ -114,27 +114,27 @@ hoist <- function(.data,
...
-  })
-  if (every(x, is_empty)) {
-    x <- NULL
+
+    out[[.col]] <- x
   }
-  out[[.col]] <- x
 
-  out
+  reconstruct_tibble(.data, out)
```
DavisVaughan, line 137: This is new for `hoist()` but allows us to correctly restore grouped data frames. It is also now necessary because we use `df_append()` which will always return a bare data frame which likely needs to be restored to a tibble.

hadley (approved):
Seems reasonable to me.

Issue #1367, opened 2022-06-06
Should `unnest_wider()` default to some separator for unnamed entries?


hadley:
I can't imagine that this behaviour is ever useful:

``` r
library(tidyr)

df <- tribble(
  ~x, ~y,
  1, c(11, 12, 13),
  2, 21
)
df |> unnest_wider(y)
#> New names:
#> New names:
#> • `` -> `...1`
#> • `` -> `...2`
#> • `` -> `...3`
#> # A tibble: 2 × 4
#>       x  ...1  ...2  ...3
#>   <dbl> <dbl> <dbl> <dbl>
#> 1     1    11    12    13
#> 2     2    21    NA    NA
```

<sup>Created on 2022-06-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

What if it was instead:

``` r
df |> unnest_wider(y)
#> Warning: 
#> * Using unnest_wider() with column containing unnamed data
#> i Use `names_sep = "_"` to suppress this warning and control the separator.
#> # A tibble: 2 × 4
#>       x   y_1   y_2   y_3
#>   <dbl> <dbl> <dbl> <dbl>
#> 1     1    11    12    13
#> 2     2    21    NA    NA
```

?

hadley:
And could it recommend `names_sep` instead of (as well as?) `names_repair` here?

``` r
library(tidyr)

df <- tibble(x = 1, y = list(list(x = 2)))
df |> unnest_wider(y)
#> Error in `unpack()`:
#> ! Names must be unique.
#> ✖ These names are duplicated:
#>   * "x" at locations 1 and 2.
#> ℹ Use argument `names_repair` to specify repair strategy.
```

<sup>Created on 2022-06-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

hadley:
Need to do the same thing for `separate_wider_*()`

PR #1448, opened 2023-01-05
Improve on `unpack()` column name duplication errors

### LINKED ISSUE #1425: `separate_wider_regex()`, `names_sep`, and `too_few`
Moving from Slack.

I just used `separate_wider_regex()` on a tibble created with parsermd on the section headings from a chapter from R Packages.

It turns out I needed to specify `names_sep` and `too_few`, but those user errors interact with each other in a way that makes it harder to figure out what's wrong.

``` r
library(tidyverse)

x <- tribble(
  ~ sec_h1, ~ sec_h2,
  "Intro {intro}", NA,
  "First cool thing {first-cool-thing}", NA,
  "First cool thing {first-cool-thing}", "First cool details {first-cool-details}",
  "Second cool thing", NA
)
x |> 
  separate_wider_regex(
    everything(),
    c("text" = "[^\\{]*", " [{]", label = ".*", "[}]"),
    too_few = "debug"
  )
#> Warning: Debug mode activated: adding variables `sec_h1_ok`, `sec_h1_matches`, and
#> `sec_h1_remainder`.
#> Warning: Debug mode activated: adding variables `sec_h2_ok`, `sec_h2_matches`, and
#> `sec_h2_remainder`.
#> Error in `unpack()`:
#> ! Names must be unique.
#> ✖ These names are duplicated:
#>   * "text" at locations 1 and 7.
#>   * "label" at locations 2 and 8.
#> ℹ Use argument `names_repair` to specify repair strategy.
```

I think I should be getting scolded first and foremost for not specifying `names_sep`, which is actually required since `cols = everything()`. That is the root cause of the `unpack()` error but the suggested remedy is off-base.

I'd probably also be happy with some default choice of `names_sep` (related #1367).

Until you get the naming stuff straightened out, you can't see the cool `too_few = "debug"` diagnostics.

DavisVaughan:
Closes #1425 (Unless we decide otherwise)
Related to #1367

I'd say that this resolves #1425 by giving a much better error message.

It also resolves this comment https://github.com/tidyverse/tidyr/issues/1367#issuecomment-1147998930, again with a much better error message.

It does _not_ resolve the original issue presented in #1367 about using some automatic separator for unnamed elements. I think that is related but can be thought about separately. I've kept that one open.

In general this improves errors for `unpack()`, `unnest()`, `unnest_wider()`, and the `separate_wider_*()` family.

---

There are two types of name duplication issues that can occur when unpacking, which I have called:
- inner / inner duplication (unpacking multiple columns that have duplicate inner names)
- outer / inner duplication (unpacking 1 column that duplicates an outer name)

This PR attempts to give better error messages for both of those cases when `names_sep = NULL` and `names_repair = "check_unique"`, i.e. the user has not attempted to solve these issues themselves yet.

I tried not to mention "unpacking" in the error because these bubble up from other functions.

``` r
library(tidyr)

# inner / inner
df <- tibble(
  x = list(tibble(a = 1)),
  y = list(tibble(a = 2))
)
unnest(df, c(x, y))
#> Error in `unpack()` at tidyr/R/unnest.R:181:2:
#> ! Can't duplicate names across the modified columns.
#> ✖ These names are duplicated:
#>   ℹ `a`, across `x` and `y`.
#> ℹ Use `names_sep` to disambiguate using the column name.
#> ℹ Or use `names_repair` to specify a repair strategy.

# outer / inner
df <- tibble(
  x = 1, 
  y = list(tibble(x = 2))
)
unnest(df, y)
#> Error in `unpack()` at tidyr/R/unnest.R:181:2:
#> ! Can't duplicate names between the modified columns and the original data.
#> ✖ These names are duplicated:
#>   ℹ `x`, from `y`.
#> ℹ Use `names_sep` to disambiguate using the column name.
#> ℹ Or use `names_repair` to specify a repair strategy.
```

You would get more indented bullets if there are additional problems (see tests)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 # tidyr (development version)
 
 * `unpack()` does a better job of reporting column name duplication issues and
-  gives better advise about how to resolve them using `names_sep`. This also
+  gives better advice about how to resolve them using `names_sep`. This also
   improves errors from functions that use `unpack()`, like `unnest()` and
   `unnest_wider()` (#1425, #1367).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # tidyr (development version)
 
+* `unpack()` does a better job of reporting column name duplication issues and
+  gives better advise about how to resolve them using `names_sep`. This also
+  improves errors from functions that use `unpack()`, like `unnest()` and
+  `unnest_wider()` (#1425, #1367).
+
 * `unnest_longer()` now uses `""` in the index column for fully unnamed
   vectors. It also now consistently uses `NA` in the index column for empty
   vectors that are "kept" by `keep_empty = TRUE` (#1442).
--- a/R/pack.R
+++ b/R/pack.R
@@ -112,6 +112,13 @@ unpack <- function(data, cols, names_sep = NULL, names_repair = "check_unique") 
 
   cols_names <- names(cols)
 
+  if (is.null(names_sep) && is_string(names_repair, "check_unique")) {
+    # Only perform checks if user hasn't supplied `names_sep` or `names_repair`.
+    # We let `vec_as_names()` catch any remaining problems.
+    check_inner_inner_duplicate(cols)
+    check_outer_inner_duplicate(cols, names(data))
+  }
+
   if (!is.null(names_sep)) {
     out[cols_names] <- map2(
       cols,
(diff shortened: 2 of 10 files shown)


R/pack.R
```diff
@@ -139,6 +146,87 @@ unpack <- function(data, cols, names_sep = NULL, names_repair = "check_unique")
...
+
+  bullets <- map2_chr(inners, outers, function(inner, outer) {
+    cli::format_inline("{.code {inner}}, across {.code {outer}}.")
+  })
+  bullets <- set_names(bullets, "i")
+  bullets <- cli::format_bullets_raw(bullets)
+  bullets <- set_names(bullets, " ")
+
+  message <- c(
+    "Can't duplicate names across the modified columns.",
```
hadley, line 188:
```suggestion
    "Can't duplicate names within the affected columns.", 
```
?

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # tidyr (development version)
 
+* `unpack()` does a better job of reporting column name duplication issues and
+  gives better advise about how to resolve them using `names_sep`. This also
+  improves errors from functions that use `unpack()`, like `unnest()` and
+  `unnest_wider()` (#1425, #1367).
```
hadley, line 6: I also checked what happens if you unnest a data frame with bad names:

```R
df <- tibble(
  x = list(data.frame(a = 1, a = 2, a = 3, check.names = FALSE))
)
df |> unnest(x)
```

And the error looks reasonable there too.

R/pack.R
```diff
@@ -139,6 +146,87 @@ unpack <- function(data, cols, names_sep = NULL, names_repair = "check_unique")
...
+
+  bullets <- map2_chr(problems, names(problems), function(inner, outer) {
+    cli::format_inline("{.code {inner}}, from {.code {outer}}.")
+  })
+  bullets <- set_names(bullets, "i")
+  bullets <- cli::format_bullets_raw(bullets)
+  bullets <- set_names(bullets, " ")
+
+  message <- c(
+    "Can't duplicate names between the modified columns and the original data.",
```
hadley, line 220: Is this consistent with other tidyr functions? Should the inner names replace the outer names?

hadley (approved):

NEWS.md
DavisVaughan: There is actually a test for that case https://github.com/tidyverse/tidyr/pull/1448/files#diff-2d1025444fee4618ff7bfdb6355745f5664188d6133df3abf807c4dc2413cc7aR152-R161

R/pack.R
DavisVaughan: I feel like that might be more surprising than useful? i.e. the inner/outer collision is the reason for the first error you see here https://tidyr.tidyverse.org/articles/rectangle.html#sharla-gelfands-discography, but I feel like knowing there was a collision is pretty useful. And having `id` replaced "automatically" by the inner one might be surprising, especially if you aren't exactly sure what the names of those inner columns are.

It seems like `pivot_longer/wider()` also error, and they don't use `unpack()`.

``` r
tibble(x = c(1, 2), a = c(3, 4), b = c(5, 6)) %>%
  pivot_longer(c(a, b), names_to = "x")
#> Error in `pivot_longer()`:
#> ! Names must be unique.
#> ✖ These names are duplicated:
#>   * "x" at locations 1 and 2.
#> ℹ Use argument `names_repair` to specify repair strategy.

#> Backtrace:
#>      ▆
#>   1. ├─tibble(x = c(1, 2), a = c(3, 4), b = c(5, 6)) %>% ...
#>   2. ├─tidyr::pivot_longer(., c(a, b), names_to = "x")
#>   3. └─tidyr:::pivot_longer.data.frame(., c(a, b), names_to = "x") at tidyr/R/pivot-long.R:153:2
#>   4.   └─tidyr::pivot_longer_spec(...) at tidyr/R/pivot-long.R:185:2
#>   5.     ├─tidyr:::wrap_error_names(...) at tidyr/R/pivot-long.R:328:2
#>   6.     │ └─base::withCallingHandlers(...) at tidyr/R/pivot.R:57:2
#>   7.     └─vctrs::vec_cbind(...) at tidyr/R/pivot-long.R:328:2
#>   8.       └─vctrs (local) `<fn>`()
#>   9.         └─vctrs:::validate_unique(names = names, arg = arg, call = call)
#>  10.           └─vctrs:::stop_names_must_be_unique(names, arg, call = call)
#>  11.             └─vctrs:::stop_names(...)
#>  12.               └─vctrs:::stop_vctrs(...)
#>  13.                 └─rlang::abort(message, class = c(class, "vctrs_error"), ..., call = vctrs_error_call(call))

tibble(x = 1:2, a = c("x", "y"), b = c(3, 4)) %>%
  pivot_wider(names_from = a, values_from = b)
#> Error in `pivot_wider()`:
#> ! Names must be unique.
#> ✖ These names are duplicated:
#>   * "x" at locations 1 and 2.
#> ℹ Use argument `names_repair` to specify repair strategy.
```

Things like `unnest()`, `unnest_wider()`, `separate_*()` all go through `unpack()` so those will be consistent.

hadley: Ok, sounds good.

PR #1460, opened 2023-01-12
Standardise data-raw scripts

### LINKED ISSUE #1459: Remove vroom/readr class from built in datasets
```r
> class(fish_encounters)
[1] "spec_tbl_df" "tbl_df"      "tbl"         "data.frame" 
```

hadley:
* Always produce standard tbls (not spec_tbl created by readr)
* Fix deprecated functions
* Standardise package loading
* Create csvs for all external datasets

Fixes #1459

---

I verified that all datasets had the correct type with 

```R
data <- dir("data", full.names = TRUE)
data |> lapply(function(path) {
  env <- new.env()
  name <- load(path, envir = env)
  class(env[[name]])
})
```

and that all scripts worked in a fresh session with

```R
scripts <- dir("data-raw", pattern = "\\.R$", full.names = TRUE)
for (script in scripts) {
  cat(script, "\n")
  callr::r(function(path) {
    source(path)
  }, list(path = script))
}
```


DavisVaughan (approved):
`cms_patient_experience` seems to have changed a lot in the CSV, is that worrisome? It looks like it pulls from an API

PR #1461, opened 2023-01-12
Implement `nest(.by = )` and revive `.key`

### LINKED ISSUE #1458: Can we somehow have `nest(.by = )`?
`dplyr::nest_by(data, x, y)` is somewhat useful because I like that the variables you specify are the ones that end up _outside_ the nesting. The equivalent `nest(data, -c(x, y))` has always felt very awkward to me.

But really we'd like to pick one of `nest()`, `nest_by()` and `group_nest()` and deprecate the others.

It feels like tidyr is the right place for `nest()` to live, but it would be nice if we could somehow bring that `by` behavior over.

Maybe it is something like `nest(.data, ..., .by = NULL/everything())`? Then you could do `nest(data, .by = c(x, y))` to get the `nest_by()` behavior. I still think you could use the `...` in combination with `.by`, we'd just have to think about exactly how that would work (and can you duplicate variables in both `.by` and `...`?)

We'd have to consider if this conflicts with the current behavior of `nest(data)` with no `...`, which currently nests everything (rather than nothing, as might be expected by `nest(data, .by = c(x, y))`)

Consider how this conflicts with dplyr `.by`, and if that would be confusing or not. It might not be, because they seem to be very similar concepts.

DavisVaughan:
Closes #1458 

``` r
df <- tibble(x = c(1, 1, 1, 2, 2, 3), y = 1:6, z = 6:1)

# Can now do this
nest(df, .by = x)
#> # A tibble: 3 × 2
#>       x data            
#>   <dbl> <list>          
#> 1     1 <tibble [3 × 2]>
#> 2     2 <tibble [2 × 2]>
#> 3     3 <tibble [1 × 2]>

# Rather than either of these:
# nest(df, data = -x)
# nest(df, data = c(y, z))
```

Also has these two cool new features:

``` r
# Implicit `select(df, -z)`
nest(df, data = y, .by = x)
#> # A tibble: 3 × 2
#>       x data            
#>   <dbl> <list>          
#> 1     1 <tibble [3 × 1]>
#> 2     2 <tibble [2 × 1]>
#> 3     3 <tibble [1 × 1]>

# Include `x` in the inner nesting as well
nest(df, data = everything(), .by = x)
#> # A tibble: 3 × 2
#>       x data            
#>   <dbl> <list>          
#> 1     1 <tibble [3 × 3]>
#> 2     2 <tibble [2 × 3]>
#> 3     3 <tibble [1 × 3]>
```

This also revives the `.key` argument, with a default of `NULL`. This is a bit tricky because it used to be `deprecated()`, and if S3 method authors updated their methods to also say `.key = deprecated()` as the default, then their code may break. I've handled this by silently converting `deprecated()` to `NULL`.

`.key` is used whenever `...` isn't specified. This happens in 3 ways:
- `nest(df)` where we nest everything
- `nest(df, .by = x)` where we specify columns to nest by
- `df %>% group_by(x) %>% nest()` where columns to nest by are implied by the groups

If both `...` and `.key` are specified, a warning is thrown saying `.key` is ignored.

---

The only revdep that breaks because of these changes is panelr, because they have sticky `[` methods and we now use `df[cols]` in `nest()`. I can send them a PR after we merge this.

### DIFF
--- a/R/nest.R
+++ b/R/nest.R
@@ -76,14 +76,10 @@
 #' @examples
 #' df <- tibble(x = c(1, 1, 1, 2, 2, 3), y = 1:6, z = 6:1)
 #'
+#' # Specify variables to nest using name-variable pairs.
 #' # Note that we get one row of output for each unique combination of
-#' # non-nested variables
+#' # non-nested variables.
 #' df %>% nest(data = c(y, z))
-#' # `chop()` does something similar, but retains individual columns
-#' df %>% chop(c(y, z))
-#'
-#' # Use tidyselect syntax and helpers, just like in `dplyr::select()`
-#' df %>% nest(data = any_of(c("y", "z")))
 #'
 #' # Specify variables to nest by (rather than variables to nest) using `.by`
 #' df %>% nest(.by = x)
@@ -92,6 +88,9 @@
 #' # name with `.key`
 #' df %>% nest(.by = x, .key = "cols")
 #'
+#' # Use tidyselect syntax and helpers, just like in `dplyr::select()`
+#' df %>% nest(data = any_of(c("y", "z")))
+#'
 #' # `...` and `.by` can be used together to drop columns you no longer need,
 #' # or to include the columns you are nesting by in the inner data frame too.
 #' # This drops `z`:
@@ -99,9 +98,7 @@
 #' # This includes `x` in the inner data frame:
 #' df %>% nest(data = everything(), .by = x)
 #'
-#' iris %>% nest(.by = Species)
-#' nest_vars <- names(iris)[1:4]
-#' iris %>% nest(data = any_of(nest_vars))
+#' # Multiple nesting structures can be specified at once
 #' iris %>%
 #'   nest(petal = starts_with("Petal"), sepal = starts_with("Sepal"))
 #' iris %>%
--- a/R/nest.R
+++ b/R/nest.R
@@ -125,11 +125,6 @@ nest <- function(.data,
                  .by = NULL,
                  .key = NULL,
                  .names_sep = NULL) {
-  if (identical(.key, deprecated())) {
-    # Temporary support for S3 method authors that set `.key = deprecated()`
-    return(nest(.data, ..., .by = {{ .by }}, .key = NULL, .names_sep = .names_sep))
-  }
-
   cols <- enquos(...)
   empty <- names2(cols) == ""
 
@@ -244,7 +239,7 @@ nest_info <- function(.data,
 
   key <- check_key(.key, error_call = .error_call)
 
-  if (n_cols != 0L && !is.null(.key)) {
+  if (n_cols != 0L && !is_default_key(.key)) {
     warn_unused_key(error_call = .error_call)
   }
 
@@ -302,10 +297,20 @@ warn_unused_key <- function(error_call = caller_env()) {
 }
 
 check_key <- function(key, error_call = caller_env()) {
-  if (is.null(key)) {
+  if (is_default_key(key)) {
     "data"
   } else {
     check_string(key, allow_empty = FALSE, arg = ".key", call = error_call)
     key
   }
+}
+
+is_default_key <- function(key) {
+  if (identical(maybe_missing(key), deprecated())) {
+    # Temporary support for S3 method authors that set `.key = deprecated()`.
+    # Remove this entire helper all methods have been updated.
+    key <- NULL
+  }
+
+  is.null(key)
 }
--- a/R/nest.R
+++ b/R/nest.R
@@ -9,10 +9,14 @@
 #'
 #' Learn more in `vignette("nest")`.
 #'
+#' @details
+#' If neither `...` nor `.by` are supplied, `nest()` will nest all variables,
+#' and will use the column name supplied through `.key`.
+#'
 #' @section New syntax:
 #' tidyr 1.0.0 introduced a new syntax for `nest()` and `unnest()` that's
 #' designed to be more similar to other functions. Converting to the new syntax
-#' should be straightforward (guided by the message you'll recieve) but if
+#' should be straightforward (guided by the message you'll receive) but if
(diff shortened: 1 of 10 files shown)


R/nest.R
```diff
@@ -74,20 +112,32 @@
...
-  cols <- enquos(...)
+nest <- function(.data,
+                 ...,
+                 .by = NULL,
+                 .key = NULL,
+                 .names_sep = NULL) {
+  if (identical(.key, deprecated())) {
+    # Temporary support for S3 method authors that set `.key = deprecated()`
+    return(nest(.data, ..., .by = {{ .by }}, .key = NULL, .names_sep = .names_sep))
+  }
```
DavisVaughan, line 131: My hack to hopefully catch S3 methods that set `.key = deprecated()`. I know tibbletime did that. It doesn't look like sf or tsibble do it, so hopefully no big issues.

R/nest.R
```diff
@@ -106,39 +156,52 @@ nest <- function(.data, ..., .names_sep = NULL, .key = deprecated()) {
       i = "Did you want `{(.key)} = {cols_fixed_label}`?"
     ))
 
-    return(nest(.data, !!!cols))
+    return(nest(.data, !!!cols, .by = {{ .by }}))
```
DavisVaughan, line 159: Supports `df %>% nest(y, .by = x)` where we would get the warning about forgetting `data = y` but then it would continue as expected

R/nest.R
```diff
@@ -106,39 +156,52 @@ nest <- function(.data, ..., .names_sep = NULL, .key = deprecated()) {
...
+  cols <- info$cols
+  inner <- info$inner
+  outer <- info$outer
+
+  inner <- .data[inner]
+  inner <- pack(inner, !!!cols, .names_sep = .names_sep, .error_call = error_call)
+
+  out <- .data[outer]
+  out <- vec_cbind(out, inner, .name_repair = "check_unique", .error_call = error_call)
+  out <- reconstruct_tibble(.data, out)
```
DavisVaughan, line 204: `vec_cbind()` with a grouped data frame will drop the `<grouped_df>` class and return a bare tibble. So we use `reconstruct_tibble()` after that to restore the grouped class if possible. There are tests for this.

Both `pack()` and `chop()` also do `reconstruct_tibble()` on the way out so we don't need it after them

R/nest.R
```diff
@@ -150,21 +213,99 @@ nest.tbl_df <- function(.data, ..., .names_sep = NULL, .key = deprecated()) {
...
+    cli::cli_abort("Can't supply {.arg .by} when {.arg .data} is a grouped data frame.")
+  }
+
   if (missing(...)) {
     .key <- check_key(.key)
     cols <- setdiff(names(.data), dplyr::group_vars(.data))
     nest.tbl_df(.data, !!.key := all_of(cols), .names_sep = .names_sep)
   } else {
-    NextMethod()
+    nest.tbl_df(.data, ..., .key = .key, .names_sep = .names_sep)
```
DavisVaughan, line 231: This was a bit unfortunate. Adding a tidy-select `.by` argument means that you can't call `NextMethod()` anymore, you have to recall the generic (or the method directly, as is done here).

I'm not sure if this will break existing code or not

tests/testthat/test-nest.R
```diff
@@ -90,26 +90,157 @@ test_that("can nest multiple columns", {
...
+
+test_that("catches when `...` overwrites an existing column", {
+  df <- tibble(x = 1, y = 2)
+
+  # Hardcoded as an error.
+  # Name repair would likely break internal usage of `chop()`.
+  expect_snapshot(error = TRUE, {
+    nest(df, x = y)
+  })
 })
```
DavisVaughan, line 156: This was previously not being caught, and was giving an even more obscure error from `chop()`

R/nest.R
DavisVaughan: This didn't work quite right. See https://github.com/tidyverse/tidyr/pull/1461/commits/ab2673dda0047846b5829c0481cb89477d008a4a

R/nest.R
```diff
@@ -150,21 +208,109 @@ nest.tbl_df <- function(.data, ..., .names_sep = NULL, .key = deprecated()) {
...
+    key
+  }
+}
+
+is_default_key <- function(key) {
+  if (identical(maybe_missing(key), deprecated())) {
+    # Temporary support for S3 method authors that set `.key = deprecated()`.
+    # Remove this entire helper all methods have been updated.
+    key <- NULL
   }
```
DavisVaughan, line 313: My "hack" to catch `.key = deprecated()` from other S3 methods

R/nest.R
```diff
@@ -24,44 +28,78 @@
...
-#' # chop does something similar, but retains individual columns
+#' # `chop()` does something similar, but retains individual columns
 #' df %>% chop(c(y, z))
 #'
-#' # use tidyselect syntax and helpers, just like in dplyr::select()
+#' # Use tidyselect syntax and helpers, just like in `dplyr::select()`
 #' df %>% nest(data = any_of(c("y", "z")))
 #'
-#' iris %>% nest(data = !Species)
+#' # Specify variables to nest by (rather than variables to nest) using `.by`
```
hadley, line 88: Maybe move these examples before the examples of using `data`?

R/nest.R
```diff
@@ -106,39 +151,52 @@ nest <- function(.data, ..., .names_sep = NULL, .key = deprecated()) {
       i = "Did you want `{(.key)} = {cols_fixed_label}`?"
```
hadley, line 151: Should we switch this to `lifecycle_warn(always = TRUE)`?  Or maybe even make it defunct? It was deprecated in tidyr 1.0.0 (August 2019), so it's been >3 years with a warning on every use.

hadley (approved):
Maybe we've finally gotten the nesting interface right!

R/nest.R
DavisVaughan: Hmm I'll fiddle with that in a follow up PR

PR #1464, opened 2023-01-13
Advance unnamed `nest()` specification to defunct stage

DavisVaughan:
`"afe60a50-970e-4cfc-815b-28350cfbf490"`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # tidyr (development version)
 
+* Calling `nest()` without naming `...` is now defunct. It has been deprecated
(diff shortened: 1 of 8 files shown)


DavisVaughan:
@hadley 

5 packages fail if we change to defunct. I think that is probably too many at this stage, so we can do this in the next minor release. I will do a separate PR that transitions us over to actually use `lifecycle::deprecate_warn(always = TRUE)` though

- cubble
- highcharter
- HTSSIP
- T2Qv
- unpivotr

hadley:
Agreed, lets do it next release.

PR #1465, opened 2023-01-13
Rethrow purrr index errors

DavisVaughan:
Closes #1440

- Required purrr 1.0.1
- Switched from `lapply()` to `map()` everywhere
- Added `with_indexed_errors()` to ease rethrowing with a custom message

@hadley, @lionel- what do you think of this rethrowing pattern in `with_indexed_errors()`? It seems useful to expose `message` so the caller of `with_indexed_errors()` can customize as needed, and I've fiddled with the environment so that cli interpolation works as usual AND you get access to `cnd` when the `message` is eventually interpolated.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -331,12 +331,12 @@ check_list_of_bool <- function(x, names, arg = caller_arg(x), call = caller_env(
 with_indexed_errors <- function(expr,
                                 message,
                                 error_call = caller_env(),
-                                envir = caller_env()) {
+                                env = caller_env()) {
   try_fetch(
     expr,
     purrr_error_indexed = function(cnd) {
-      envir <- new_environment(list(cnd = cnd), parent = envir)
-      cli::cli_abort(message, call = error_call, parent = cnd$parent, .envir = envir)
+      message <- message(cnd)
+      cli::cli_abort(message, call = error_call, parent = cnd$parent, .envir = env)
     }
   )
 }
--- a/R/utils.R
+++ b/R/utils.R
@@ -328,6 +328,19 @@ check_list_of_bool <- function(x, names, arg = caller_arg(x), call = caller_env(
   }
 }
 
+with_indexed_errors <- function(expr,
+                                message,
+                                error_call = caller_env(),
+                                envir = caller_env()) {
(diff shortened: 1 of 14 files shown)


hadley:
Seems reasonable to me.

R/utils.R
```diff
@@ -327,6 +328,19 @@ check_list_of_bool <- function(x, names, arg = caller_arg(x), call = caller_env(
   }
 }
 
+with_indexed_errors <- function(expr,
+                                message,
+                                error_call = caller_env(),
+                                envir = caller_env()) {
```
lionel-, line 334: Minor: Use `env` instead of `envir` in this sort of code?

R/utils.R
```diff
@@ -327,6 +328,19 @@ check_list_of_bool <- function(x, names, arg = caller_arg(x), call = caller_env(
...
+  try_fetch(
+    expr,
+    purrr_error_indexed = function(cnd) {
+      envir <- new_environment(list(cnd = cnd), parent = envir)
```
lionel-, line 338: It's surprising to use the same variable name. Should use a new one so it's immediately clear that we're not interpolating in `envir = caller_env()`.

lionel- (approved):
Looks good!

I'm not sure we should use this pattern in exported functions though because non-standard pronouns tend to be surprising. An alternative is to leave the environment as is and supply a method instead:

```r
with_indexed_errors(fn(), message = function(cnd, ...) cli::format_inline("{cnd$name}"))
```

This is more verbose but has the advantage of declaring `cnd` lexically, and the method explicitly inherits from the current environment too, so the scoping is transparent.

R/utils.R
DavisVaughan: I was just being consistent with cli here 😞

DavisVaughan:
I've switched to forcing `message` to be a function taking `cnd`. I agree with @lionel- that that is cleaner and would be easier to document if we ever exported this from rlang.

I've also tweaked the API to just use `rlang::abort()`. It now resembles something rlang could eventually export alongside https://github.com/r-lib/rlang/issues/1533

PR #1467, opened 2023-01-13
Use standard lifecycle machinery for `nest()` warning

DavisVaughan:
Closes #1464


hadley (approved):
FWIW I don't think this needs a news bullet

PR #1474, opened 2023-01-24
give attribution to previous work in nc package

tdhock:
Hi! The new `separate_wider_regex` functions seems to be inspired by previous work in my namedCapture/nc packages, so I would suggest making it clear about where that idea comes from. At least in the academic/university, attribution is important so it would be useful for people to know/cite my related R Journal papers, 
https://journal.r-project.org/archive/2019/RJ-2019-050/index.html
https://journal.r-project.org/archive/2021/RJ-2021-029/index.html

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # tidyr (development version)
 
+# tidyr 1.3.0
+
 ## New features
 
 * New family of consistent string separating functions:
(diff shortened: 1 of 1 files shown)


hadley:
I have no recollection of ever seeing that package, but the chances are I did, and I just don't remember it. So we should certainly include a nod to this package in the news file and the docs.

tdhock:
thanks!

hadley:
I tweaked the language a bit since I think it's probably an example of convergent evolution, rather than a direct inspiration.

PR #1520, opened 2023-09-27
Fix `separate_wider_*()` error messages

matthewjnield:
Fixes tidyverse#1475. The `separate_wider_position()` and `separate_wider_regex()` functions provide the `too_few` argument to control their behavior when one or more of the strings cannot be separated into as many pieces as the user specifies. When this happens, the error message that is displayed recommends specifying too_few as "start", rather than the valid value "align_start".

This pull request changes the error messages to recommend using "align_start" instead of "start".


hadley:
Thank you!

PR #1521, opened 2023-09-27
Fix typo in definition of tidy data in README

mfgeary:
The `README` (`README.md` and `README.Rmd`) files have a typo in the definition of tidy data. I changed the text from "Every column is variable" to "Every column is a variable".

It's a very minor change but important because these sentences have different meanings. Without the "a", it is more difficult to understand what tidy data is, and this is the first thing you see when you go to `tidyr`'s homepage.

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -24,7 +24,7 @@ knitr::opts_chunk$set(
 
 The goal of tidyr is to help you create __tidy data__. Tidy data is data where:
 
-1. Every column is variable.
+1. Every column is a variable.
 1. Every row is an observation.
 1. Every cell is a single value.
 
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ coverage](https://codecov.io/gh/tidyverse/tidyr/branch/main/graph/badge.svg)](ht
 The goal of tidyr is to help you create **tidy data**. Tidy data is data
 where:
 
-1.  Every column is variable.
+1.  Every column is a variable.
 2.  Every row is an observation.
 3.  Every cell is a single value.
 


matthewjnield:
@mfgeary Good catch!

I'm adding a note here as an idea for the maintainers' discussion:

Should the definition of tidy data in the `tidyr` README and website be updated to match the one that is presented in ["R for Data Science (2e)"](https://r4ds.hadley.nz/data-tidy#sec-tidy-data), or vice-versa? I think the two definitions get the same point across, but tidy data is a core concept in the tidyverse, and having a uniform definition (whichever of these two makes the concept clearer) in these two places may be a good idea.

hadley:
Thanks @mfgeary!

@matthewjnield isn't it the same definition?

matthewjnield:
@hadley they are very close; I am only wondering if the extra wording in R for Data Science 2(e) gets the point across slightly better and should be added to the tidyr README.

Here are the parts from each source I am referring to:

R for Data Science 2(e):
1. Each variable is a column; each column is a variable.
2. Each observation is a row; each row is an observation.
3. Each value is a cell; each cell is a single value.

tidyr README:
1. Every column is a variable.
2. Every row is an observation.
3. Every cell is a single value.

hadley:
Oh I see. Do you want to do a PR?

mfgeary:
@matthewjnield @hadley I chose the shorter definition to be consistent across `tidyr` as it is also used in the tidy data vignette [here](https://tidyr.tidyverse.org/articles/tidy-data.html#tidy-data). If we switch to the definition from R4DS, I suggest switching all definitions of tidy data in `tidyr` to be consistent

matthewjnield:
This is ready, in #1532.

PR #1532, opened 2023-11-02
Standardize definition of tidy data

matthewjnield:
This commit updates the definition of tidy data used in the README and the vignette to match the one presented in ["R for Data Science (2e)"](https://r4ds.hadley.nz/data-tidy#sec-tidy-data).


hadley:
Thanks!

PR #1540, opened 2024-02-05
Re-enable autocompletion for pivot_wider and pivot_longer

olivroy:
Autocompletion text in RStudio doesn't currently work for `pivot_longer()` and `pivot_wider()`. It seems to be due to the fact that the S3 method is not present in the Rd file.

This PR seems to fix that. I am not sure if this is the right fix, or it is with roxygen2 or RStudio itself..

I installed this version locally and it works.

image

With the current cran version, when tidyr is loaded, the autocompletion doesn't show any text, only the parameters
image

It seems that if the method is not in an Rd, RStudio doesn't know where to look for the help text.

If I typed F1 (help) before this PR, I would get the message:
```r
No documentation for ‘pivot_wider.data.frame’ in specified packages and libraries:
you could try ‘??pivot_wider.data.frame’
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -50,4 +50,4 @@ Config/testthat/edition: 3
 Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.3.0
+RoxygenNote: 7.3.1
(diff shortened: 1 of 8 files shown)


olivroy:
No longer needed with RStudio 2024.04.0+426 as a hack was added to fallback on generics to autocomplete methods if the generic is not documented.

PR #1550, opened 2024-04-25
Cran tidyselect + roxygen2 + ci

olivroy:
To pass checks on main. closes #1564 

I also updated gh actions to help pass check for r-devel?


DavisVaughan:
Thanks @olivroy! I ended up actually doing these individually already, but I really appreciate the PR!

PR #1557, opened 2024-07-20
Update tidy-data.Rmd

devpowerplatform:
it is not easy to try out the two data sets (tb and weather). with this changes, any people can run the code to get the two data sets. Also use one of tidyverse packages to read in data sets


DavisVaughan:
I don't think we want to commit to downloading something at each vignette render (locally, on ci, possibly on CRAN, etc). So the best we can do is show a comment saying how to do this "at home" if you're reading along

PR #1558, opened 2024-07-20
Update tidy-data.Rmd

devpowerplatform:
creating some reproducible example files and replacing superceded function map_dfr with list_rbind()


DavisVaughan:
This is run with `eval = FALSE` and is just a theoretical example, so I think doing too much to try and make it reproducible takes away from what we are trying to focus on, but thanks for the other changes!

PR #1567, opened 2024-08-15
improve error message for non data.frame inputs

### LINKED ISSUE #1510: `pivot_wider_spec()` should check if `data` is a data frame
The error message when passing in a `lazy_frame()` or `memdb_frame()` are not very helpful:

``` r
spec <- tibble::tibble(
  .name = c("x", "y", "z"),
  .value = "val",
  key = c("x", "y", "z")
)

dbplyr::memdb_frame(key = c("x", "y", "z"), val = 1:3) |>
  tidyr::pivot_wider_spec(spec)
#> Error in `data[vec_unique(c(id_cols, names_from_cols, values_from_cols, unused_cols))]`:
#> ! Can't subset columns that don't exist.
#> ✖ Columns `src` and `lazy_query` don't exist.
#> Backtrace:
#>      ▆
#>   1. ├─tidyr::pivot_wider_spec(...)
#>   2. │ ├─...[]
#>   3. │ └─tibble:::`[.tbl_df`(...)
#>   4. │   └─tibble:::vectbl_as_col_location(...)
#>   5. │     ├─tibble:::subclass_col_index_errors(...)
#>   6. │     │ └─base::withCallingHandlers(...)
#>   7. │     └─vctrs::vec_as_location(j, n, names, missing = "error", call = call)
#>   8. └─vctrs (local) `<fn>`()
#>   9.   └─vctrs:::stop_subscript_oob(...)
#>  10.     └─vctrs:::stop_subscript(...)
#>  11.       └─rlang::abort(...)
dbplyr::lazy_frame(key = c("x", "y", "z"), val = 1:3) |>
  tidyr::pivot_wider_spec(spec)
#> Error in `as.data.frame()`:
#> ! Can not coerce <tbl_lazy> to <data.frame>
#> Backtrace:
#>     ▆
#>  1. └─tidyr::pivot_wider_spec(...)
#>  2.   ├─tibble::as_tibble(data)
#>  3.   └─tibble:::as_tibble.default(data)
#>  4.     ├─tibble::as_tibble(...)
#>  5.     ├─base::as.data.frame(value, stringsAsFactors = FALSE)
#>  6.     └─dbplyr:::as.data.frame.tbl_lazy(value, stringsAsFactors = FALSE)
#>  7.       └─cli::cli_abort("Can not coerce {.cls tbl_lazy} to {.cls data.frame}")
#>  8.         └─rlang::abort(...)
```

<sup>Created on 2023-07-13 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

catalamarti:
fixes #1510

### DIFF
--- a/tests/testthat/test-pivot-wide.R
+++ b/tests/testthat/test-pivot-wide.R
@@ -1,3 +1,29 @@
+test_that("error if input is not a data.frame", {
+  spec <- tibble::tibble(
+    .name = c("x", "y", "z"),
+    .value = "val",
+    key = c("x", "y", "z")
+  )
+
+  expect_no_error(
+    tibble(key = c("x", "y", "z"), val = 1:3) |>
+      tidyr::pivot_wider_spec(spec)
+  )
+  expect_snapshot_error(
+    list(key = c("x", "y", "z"), val = 1:3) |>
+      tidyr::pivot_wider_spec(spec)
+  )
+  # consider removing if we do not want dbplyr in suggests
+  expect_snapshot_error(
+    dbplyr::memdb_frame(key = c("x", "y", "z"), val = 1:3) |>
+      tidyr::pivot_wider_spec(spec)
+  )
+  expect_snapshot_error(
+    dbplyr::lazy_frame(key = c("x", "y", "z"), val = 1:3) |>
+      tidyr::pivot_wider_spec(spec)
+  )
+})
+
 test_that("can pivot all cols to wide", {
   df <- tibble(key = c("x", "y", "z"), val = 1:3)
   pv <- pivot_wider(df, names_from = key, values_from = val)
--- a/tests/testthat/test-pivot-wide.R
+++ b/tests/testthat/test-pivot-wide.R
@@ -1,25 +1,12 @@
 test_that("error if input is not a data.frame", {
-  spec <- tibble::tibble(
+  spec <- tibble(
     .name = c("x", "y", "z"),
     .value = "val",
     key = c("x", "y", "z")
   )
 
-  expect_no_error(
-    tibble(key = c("x", "y", "z"), val = 1:3) %>%
-      tidyr::pivot_wider_spec(spec)
-  )
   expect_snapshot_error(
     list(key = c("x", "y", "z"), val = 1:3) %>%
-      tidyr::pivot_wider_spec(spec)
-  )
-  # consider removing if we do not want dbplyr in suggests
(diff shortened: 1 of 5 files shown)


tests/testthat/test-pivot-wide.R
```diff
@@ -1,3 +1,29 @@
...
+
+  expect_no_error(
+    tibble(key = c("x", "y", "z"), val = 1:3) %>%
+      tidyr::pivot_wider_spec(spec)
+  )
+  expect_snapshot_error(
+    list(key = c("x", "y", "z"), val = 1:3) %>%
+      tidyr::pivot_wider_spec(spec)
+  )
+  # consider removing if we do not want dbplyr in suggests
```
hadley, line 16: I would remove this because it doesn't really add anything to the test coverage — you've already used a list to check that we're getting an information about what the user actually supplied. In other words, you're now starting to test the behaviour of `obj_type_friendly`, which should be out of scope for this test.

tests/testthat/test-pivot-wide.R
```diff
@@ -1,3 +1,29 @@
+test_that("error if input is not a data.frame", {
+  spec <- tibble::tibble(
+    .name = c("x", "y", "z"),
+    .value = "val",
+    key = c("x", "y", "z")
+  )
+
+  expect_no_error(
```
hadley, line 8: Since this test is about erroring when it's not a data frame, I'd suggest dropping this expectation. (And I wouldn't test it elsewhere just because I know that if `pivot_wider_spec` stopped working with data frames so much of tidyr would break)

tests/testthat/test-pivot-wide.R
```diff
@@ -1,3 +1,29 @@
+test_that("error if input is not a data.frame", {
+  spec <- tibble::tibble(
```
hadley, line 2: No need to use `tibble::` here (or `tidyr::` below).

hadley (commented):
Thanks for working on this 😄

DavisVaughan:
@hadley @catalamarti we actually already have `check_data_frame()` which could be useful here!

@catalamarti if you use `check_data_frame()`, I'd encourage you to use it at the top of `pivot_wider_spec()`, right after the `check_dots_empty0(...)` call. We typically try and group all the `check_` functions together.

You could also pass the error call through, like:

```r
check_data_frame(data, call = error_call)
```

DavisVaughan:
Thanks so much @catalamarti!

PR #1571, opened 2024-08-18
Add .vary parameter to expand_grid()

### LINKED ISSUE #1543: Feature request: `.vary` in `expand_grid()`
`expand_grid()` is meant to replace the `purrr::cross()` family. However, these functions vary fastest, and `expand_grid()` only varies slowest (in the sense of `vec_expand_grid()`).

Although varying slowest might be more useful overall, there is no other designated replacement for the `cross()` functions. Their [current docs](https://purrr.tidyverse.org/reference/cross.html) say that `cross()` and `expand_grid()` can be used equivalently, but as that very example shows, the order of combinations is different. This was first noted in <https://github.com/tidyverse/purrr/issues/768>.

It would be truly equivalent if `expand_grid()` had a `.vary` parameter passed to `vec_expand_grid()` that could be set to `"fastest"` in this case. The default should still be `"slowest"`. Alternatively, `vec_expand_grid()` itself could fulfill this role, which should then be advertised in the `cross()` docs.

Some use cases do require varying fastest (although I don't know how common this is). For example, the vectors creating the background patterns in [these plots](https://lhdjung.github.io/scrutiny/articles/grim.html#visualizing-results-with-grim_plot) were made using `cross2()`, and the patterns wouldn't emerge when varying slowest.

JamesHWade:
Fixes #1543

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,8 @@
 
 * tidyr now requires dplyr >=1.1.0 (#1568, @catalamarti).
 
+* `expand_grid()` gains a new `.vary` argument, allowing users to control whether combinations vary fastest or slowest (default). (#1543, @JamesHWade).
+
 # tidyr 1.3.1
 
 * `pivot_wider` now uses `.by` and `|>` syntax for the dplyr helper message to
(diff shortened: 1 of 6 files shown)


DavisVaughan:
Thanks @JamesHWade!

lhdjung:
This is great! Thanks @DavisVaughan and @JamesHWade!

Issue #1439, opened 2023-01-03
`.by` support in `fill()`/`complete()`/`expand()`


markfairbanks:
To match the `.by` implementation in `dplyr` v1.1.0 functions.

DavisVaughan:
Will have to wait for tidyr 1.4.0 (i.e. after dplyr 1.1.0 is released)

hadley:
@DavisVaughan does it have to? We added it to the upcoming dbplyr release.

DavisVaughan:
That seems different though because dbplyr controls the whole underlying implementation. Here we just pass through to a dplyr function like `mutate()` or `summarise()`, so without dplyr 1.1.0 the arguments wouldn't do anything.

I guess we could check to see if dplyr 1.1.0 is installed if `.by` is attempted to be used and error if there isn't a new enough version of dplyr, but that didn't feel worth it when I originally thought about this

hadley:
Ok, makes sense.

mattwarkentin:
Absolutely love the addition of the per-operation `.by/by` arguments in `dplyr`. Now that `dplyr` 1.1.0 is released, will these be added to the aforementioned `tidyr` functions. Seems like `tidyr::nest()` already got the makeover.

DavisVaughan:
We will add them in the next tidyr release. It was easier with `nest()` because tidyr owns the implementation. With the others we call out to dplyr so we needed to wait on 1.1.0

joeycouse:
+1 on adding `.by` to `complete` and `expand`

statzhero:
Checking in on this as well, just in case I missed something. UPDATE: I see it in the docs.

PR #1572, opened 2024-08-18
Add `.by` to `fill()`

olivroy:
Addresses part of #1439.

### DIFF
--- a/R/fill.R
+++ b/R/fill.R
@@ -115,6 +115,11 @@ fill.data.frame <- function(data, ..., .direction = c("down", "up", "downup", "u
   }
 
   if (dplyr::is_grouped_df(data)) {
+    if (length(by) > 0) {
+      cli::cli_abort(c(
+        "Can't supply {.arg .by} when {.arg data} is a grouped data frame."
+      ))
+    }
     dplyr::mutate(data, dplyr::across(any_of(vars), .fns = fn))
   } else {
     dplyr::mutate(data, dplyr::across(any_of(vars), .fns = fn), .by = all_of(by))
--- a/R/fill.R
+++ b/R/fill.R
@@ -12,11 +12,14 @@
 #' boundaries. This can also be accomplished using the `.by` argument to
 #' `fill()`, which creates a temporary grouping for just this operation.
 #'
+#' @inheritParams dplyr::mutate
+#'
 #' @param data A data frame.
 #' @param ... <[`tidy-select`][tidyr_tidy_select]> Columns to fill.
 #' @param .direction Direction in which to fill missing values. Currently
 #'   either "down" (the default), "up", "downup" (i.e. first down and then up)
 #'   or "updown" (first up and then down).
+#'
 #' @export
 #' @examples
 #' # direction = "down" --------------------------------------------------------
@@ -97,20 +100,17 @@
 #'   dplyr::ungroup()
 fill <- function(data,
                  ...,
+                 .by = NULL,
                  .direction = c("down", "up", "downup", "updown")) {
+  check_dots_unnamed()
   UseMethod("fill")
 }
 
-#' @inheritParams dplyr::mutate
-#' @rdname fill
 #' @export
 fill.data.frame <- function(data,
                             ...,
-                            .direction = c("down", "up", "downup", "updown"),
-                            .by = NULL) {
-  # Must be here rather than in the generic due to the placement of `.by`
-  check_dots_unnamed()
-
+                            .by = NULL,
+                            .direction = c("down", "up", "downup", "updown")) {
   vars <- names(tidyselect::eval_select(
     expr = expr(c(...)),
     data = data,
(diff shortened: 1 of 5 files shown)


R/fill.R
```diff
@@ -82,22 +83,39 @@
...
 
+#' @inheritParams dplyr::mutate
+#' @rdname fill
 #' @export
-fill.data.frame <- function(data, ..., .direction = c("down", "up", "downup", "updown")) {
-  vars <- tidyselect::eval_select(expr(c(...)), data, allow_rename = FALSE)
+fill.data.frame <- function(data,
+                            ...,
+                            .direction = c("down", "up", "downup", "updown"),
+                            .by = NULL) {
```
DavisVaughan, line 110: Putting `.by` in the `.data.frame` method rather than the generic to try and be as least disruptive as possible for S3 method authors like dbplyr

R/fill.R
```diff
@@ -82,22 +83,39 @@
...
+  # Must be here rather than in the generic due to the placement of `.by`
+  check_dots_unnamed()
```
DavisVaughan, line 112: Had to move this here out of the generic. Since `.by` isn't in the generic, supplying it will cause `check_dots_unnamed()` to yell at you if its in the generic.

This means that S3 method authors should add `check_dots_unnamed()` into their methods now, if they want the same strictness.

R/fill.R
```diff
@@ -108,5 +126,9 @@ fill.data.frame <- function(data, ..., .direction = c("down", "up", "downup", "u
     vec_fill_missing(col, direction = .direction)
   }
 
-  dplyr::mutate_at(data, .vars = dplyr::vars(any_of(vars)), .funs = fn)
+  dplyr::mutate(
```
DavisVaughan, line 129: Note also switching away from `mutate_at()` here

R/fill.R
```diff
@@ -108,5 +126,9 @@ fill.data.frame <- function(data, ..., .direction = c("down", "up", "downup", "u
...
+    data,
+    dplyr::across(any_of(vars), .fns = fn),
+    .by = {{ .by }}
```
DavisVaughan, line 132: Passing straight through to dplyr means that if `.by` is wrong for some reason, then the error call shows `dplyr::mutate()` rather than `fill()`. I think that is worth it to not have to fiddle with `.by` in any way, because dplyr has some pretty unique handling for it. (See the tests for more info)

tests/testthat/test-fill.R
```diff
@@ -106,23 +106,59 @@ test_that("fill preserves attributes", {
...
   expect_identical(out$x, c(1, 1, 2))
 })
 
+test_that("errors on named `...` inputs", {
+  df <- tibble(x = c(1, NA, 2))
+
+  expect_snapshot(error = TRUE, {
+    fill(df, fooy = x)
+  })
+})
```
DavisVaughan, line 136: I noted we didn't have a test for this and since we are touching `check_dots_unnamed()` here we may as well add one

R/fill.R
DavisVaughan: TODO: We did add `.by` to the `nest()` generic, so maybe we should look at this a little more.

Maybe it isn't as disruptive as we thought due to the presence of the `...` in the generic.

DavisVaughan:
@olivroy thanks a lot! I'll take over from here. Leaving some comments for another colleague to review

R/fill.R
DavisVaughan: According to my old research on this, it doesn't look like we will cause a warning for other S3 method authors if we add `.by` to the generic:

https://github.com/DavisVaughan/methodtest/blob/0010788d798e9ff851a871db3cd35e8ee1cb4dd6/R/methods.R#L55-L67

They will just have to opt in to support of `.by` on their own over time or catch `.by` and error if its not `NULL` if they don't support it.

DavisVaughan: Ok, I've gone back to `.by` in the generic of `fill()` based on the above analysis. That means that for dbplyr we will see something like this where `.by` is captured in the `...` until dbplyr updates its S3 method to handle `.by` (possibly by just erroring).

By some weird combination of events this passes through dbplyr without error, I've opened https://github.com/tidyverse/dbplyr/issues/1536

```r
df <- data.frame(x = c(1, 2), y = c(2, 2), z = c(1, NA))
df_sqlite <- tbl_lazy(df, con = simulate_sqlite())
df_sqlite |> window_order(x) |> fill(z, .by = y)

<SQL>
SELECT `x`, `y`, MAX(`z`) OVER (PARTITION BY `..dbplyr_partition_1`) AS `z`
FROM (
  SELECT
    `df`.*,
    SUM(CASE WHEN ((`z` IS NULL)) THEN 0 ELSE 1 END) OVER (ORDER BY `x` ROWS UNBOUNDED PRECEDING) AS `..dbplyr_partition_1`,
    SUM(CASE WHEN ((`.by` IS NULL)) THEN 0 ELSE 1 END) OVER (ORDER BY `x` ROWS UNBOUNDED PRECEDING) AS `..dbplyr_partition_2`
  FROM `df`
) AS `q01`
```

DavisVaughan:
Thanks @olivroy!

PR #1579, opened 2024-10-23
Update standalone files + standardize snapshot tests

olivroy:
* update standalone with `usethis::use_standalone("r-lib/rlang")`
* Update snapshot tests to use a single `exect_snapshot(error = TRUE)` for each error
* Remove `expect_error()` inside snapshots as we no longer need the error class
* replace `expect_warning(code, regexp = NA)` by `expect_no_warning()`
* Require more recent testthat version 
* Used `testthat::set_max_fails(Inf)` for faster iteration during development


tests/testthat/test-pivot-long.R
```diff
@@ -267,19 +267,19 @@ test_that("no names doesn't generate names (#1120)", {
 test_that("multiple names requires names_sep/names_pattern", {
   df <- tibble(x_y = 1)
 
-  expect_snapshot({
-    (expect_error(build_longer_spec(df, x_y, names_to = c("a", "b"))))
+  expect_snapshot(error = TRUE, {
+    build_longer_spec(df, x_y, names_to = c("a", "b"))
```
DavisVaughan: I tend to use 1 `expect_snapshot()` per expression that errors. Otherwise you don't know if both error, you just know that at least 1 errored

The previous `expect_error()` wrappers captured and then reemitted the individual errors, so previously it ensured that each expression did indeed error

tests/testthat/test-pivot-long.R
```diff
@@ -448,20 +452,24 @@ test_that("named `cols` gives clear error (#1104)", {
...
-    (expect_error(build_longer_spec(df, x, names_to = 1)))
-    (expect_error(build_longer_spec(df, x, names_to = c("x", "y"))))
-    (expect_error(build_longer_spec(df, x, names_to = c("x", "y"), names_sep = "_", names_pattern = "x")))
-  })
+  expect_snapshot(error = TRUE, {
+    build_longer_spec(df, x, names_to = 1)
+    build_longer_spec(df, x, names_to = c("x", "y"))
+    build_longer_spec(df, x, names_to = c("x", "y"), names_sep = "_", names_pattern = "x")
+    },
+    cnd_class = TRUE
```
DavisVaughan: Split into 1 per expression

tests/testthat/test-pivot-long.R
```diff
@@ -448,20 +452,24 @@ test_that("named `cols` gives clear error (#1104)", {
...
   df <- tibble(x = 1)
 
-  expect_snapshot({
-    (expect_error(build_longer_spec(df, x, names_ptypes = 1)))
-    (expect_error(build_longer_spec(df, x, names_ptypes = list(integer()))))
-  })
+  expect_snapshot(error = TRUE, {
+    build_longer_spec(df, x, names_ptypes = 1)
+    build_longer_spec(df, x, names_ptypes = list(integer()))
+    },
```
DavisVaughan: Split into 1 per expression

tests/testthat/test-unnest-helper.R
```diff
@@ -1,30 +1,36 @@
 # df_simplify ------------------------------------------------------------
 
 test_that("`simplify` is validated", {
-  expect_snapshot({
```
DavisVaughan, line 4: Split into 1 per expression

tests/testthat/test-unnest.R
```diff
@@ -114,7 +114,7 @@ test_that("unpacks df-cols (#1112)", {
 
 test_that("unnesting column of mixed vector / data frame input is an error", {
   df <- tibble(x = list(1, tibble(a = 1)))
-  expect_snapshot((expect_error(unnest(df, x))))
+  expect_snapshot(unnest(df, x), error = TRUE, cnd_class = TRUE)
```
DavisVaughan, line 117: I took a look at the updated snapshots. In all cases I believe we can drop `cnd_class = TRUE`. 

Our advice lately is that if you _really_ care about the error class, then you do

```r
testthat::expect_error(class = )
```

and if you _also_ care about the error message, you do a separate

```r
testthat::expect_snapshot(error = )
```

But lately we have not been combining both features into a single snapshot

And in all cases here in tidyr I don't think we care about the error class

DavisVaughan (changes requested):
Thanks! A few things to tweak across the board but otherwise looks great

tests/testthat/test-gather.R
```diff
@@ -134,13 +134,11 @@ test_that("gather throws error for weird objects", {
   df <- data.frame(y = 1)
   df$x <- expression(x)
 
-  expect_snapshot(error = TRUE, cnd_class = TRUE, {
-    gather(d, key, val, -x)
+  # Can't use snapshot, it changes between versions of R
+  expect_error(gather(d, key, val, -x), "not found")
```
DavisVaughan, line 138:
```r
expect_error(gather(d, key, val, -x))
```

I think even this would be okay

olivroy: Makes sense! Should I remove the comment above?

DavisVaughan: I think the comment is nice because i probably would wonder why we aren't snapshotting this one since we do snapshot the other!

tests/testthat/test-pivot-long.R
olivroy: Thanks for the tip! Done. I ended up doing this across the board, even for snapshots I disn't touch before.

tests/testthat/test-unnest.R
olivroy: Done in b49c272d93b43b90564f8303af7624077973e0bc

DavisVaughan:
Looks awesome thanks so much!

PR #1597, opened 2025-05-18
Add `by` to `chop()`

### LINKED ISSUE #1490: Add `.by` to `chop()` too
I was thinking of this when implementing https://github.com/tidyverse/tidyr/pull/1461 but wasn't sure if it would be useful.

But now I think it can be useful because whenever I call `chop()` I often think of what i want to "chop by" when i say it out loud in my head.

``` r
library(ivs)

df <- tibble(
  id=c(rep("a",4),rep("b",2),rep("c",3)), 
  start=c(100,250,400,600,150,610,275,600,700), 
  end=c(200,300,550,650,275,640,325,675,725)
)

df <- df %>%
  mutate(range = iv(start, end), .keep = "unused")

df
#> # A tibble: 9 × 2
#>   id         range
#>   <chr>  <iv<dbl>>
#> 1 a     [100, 200)
#> 2 a     [250, 300)
#> 3 a     [400, 550)
#> 4 a     [600, 650)
#> 5 b     [150, 275)
#> 6 b     [610, 640)
#> 7 c     [275, 325)
#> 8 c     [600, 675)
#> 9 c     [700, 725)

df %>%
  tidyr::chop(range)
#> # A tibble: 3 × 2
#>   id              range
#>   <chr> <list<iv<dbl>>>
#> 1 a                 [4]
#> 2 b                 [2]
#> 3 c                 [3]
```

Like in this example I think it makes the most sense to say that I want to chop "by id".

If I say I want to "chop the `range`", you _still_ have to ask yourself what you want to chop it by, so you implicitly always need that "by" knowledge loaded in your head anyways, so we might as well let you specify that directly instead.

Notes:
- Signature change to `chop(data, cols = NULL, ..., by = NULL)`, so `cols` is now technically optional
- If one of `cols` _or_ `by` is supplied, the other is inferred as "everything else"
- We have decided in `nest()` that `nest(data)` with no other args means "nest everything" so we probably want to either be consistent with that here (i.e. "chop everything") _or_ say that you must specify at least one of `cols` or `by`, which does feel safer.
- Allow for both `cols` and `by` to be specified together, like `nest()`
- Can probably borrow a lot from `nest_info()`? Probably worth trying to harmonize that between the two functions

### LINKED ISSUE #1506: pivot_wider: unexpected behavior specifying id_cols to omit that are included in names/values_from
I was running `pivot_wider` on some data and was surprised by the inability to use `-c(col1, col2)` to choose my `id_cols`, resulting in the error:
```
Error in `pivot_wider()`:
`id_cols` can't select a column already selected by `names_from`.
Column `type` has already been selected.
```

Repro:
```
library(dplyr)
library(tidyr)

tmp <- data.frame(id1 = c("a", "a", "b", "b"),
                  unused = c(NA, NA, NA, NA),
                  type = c("c", "d", "c", "d"),
                  values = c(1, 2, 3, 4))
```

Base case:
```
tmp %>% pivot_wider(id_cols = id1, names_from = type, values_from = values)

# A tibble: 2 × 3
  id1       c     d
  <chr> <dbl> <dbl>
1 a         1     2
2 b         3     4
```

But say you had a lot of columns; it's more concise to remove a few than name them all. Neither of these work, and produce the error above:
```
tmp %>% pivot_wider(id_cols = -c(type, values, unused), names_from = type, values_from = values)
tmp %>% pivot_wider(id_cols = c(-type, -values, -unused), names_from = type, values_from = values)
```

My failure mode may be covered by this statement from [the docs](https://tidyr.tidyverse.org/reference/pivot_wider.html):
> id_cols [...] Defaults to all columns in data except for the columns specified through names_from and values_from. If a tidyselect expression is supplied, it will be evaluated on data after removing the columns specified through names_from and values_from.

This is why I included the "unused" column, as for data with many columns, one would have to think about "ok, I'm removing `type` and `values` 'for free' since they are used in other args, but I *do* need to remember to remove those other columns."
```
tmp %>% pivot_wider(id_cols = -unused, names_from = type, values_from = values)

# A tibble: 2 × 3
  id1       c     d
  <chr> <dbl> <dbl>
1 a         1     2
2 b         3     4
```

Thoughts:
- this is a bug, in that there should be no problem specifying columns to drop, even if they are implicitly dropped by being passed to `names_from` or `values_from`
- this is not a bug, but documentation could be improved. I was confused by the message: "Column `type` has already been selected"... by what/how?" It was non-intuitive to me that it's "selected" when I'm trying to explicitly not select it as an id_col
- not a bug, and the documentation is perfectly clear. I admittedly don't use pivot functions *that* often, so it could be a misunderstanding on my part.

hrryt:
Fixes #1490
Fixes #1506

Snags:

1. In the absence of both `cols` and `by`, should we 'chop everything' as `nest()` does? I have thrown an error instead.
2. Should `chop()` use group vars if `data` is grouped, as `nest()` does? Would this require a generic?
3. What should happen if a column is selected by both `cols` and `by`? For now, the name is duplicated and repaired.

An example for the latter point:
```r
df <- tibble(x = c(1, 1, 1, 2, 2), y = c(2, 1, 2, 3, 4), z = 1:5)
nest(df, data = everything(), .by = y)
#> # A tibble: 4 × 2
#>       y data            
#>   <dbl> <list>          
#> 1     2 <tibble [2 × 3]>
#> 2     1 <tibble [1 × 3]>
#> 3     3 <tibble [1 × 3]>
#> 4     4 <tibble [1 × 3]>
chop(df, everything(), by = y)
#> New names:
#> • `y` -> `y...1`
#> • `y` -> `y...3`
#> # A tibble: 4 × 4
#>   y...1           x       y...3           z
#>   <dbl> <list<dbl>> <list<dbl>> <list<int>>
#> 1     2         [2]         [2]         [2]
#> 2     1         [1]         [1]         [1]
#> 3     3         [1]         [1]         [1]
#> 4     4         [1]         [1]         [1]
```

### DIFF
--- a/R/chop.R
+++ b/R/chop.R
@@ -126,7 +126,7 @@ chop_info <- function(
   cols_is_null <- quo_is_null(cols)
   by_is_null <- quo_is_null(by)
 
-  if(cols_is_null && by_is_null) {
+  if (cols_is_null && by_is_null) {
     stop_use_cols_or_by(error_call = error_call)
   }
 
--- a/R/chop.R
+++ b/R/chop.R
@@ -124,18 +124,20 @@
 #' df |> unchop(y, keep_empty = TRUE)
 chop <- function(
   data,
-  cols = NULL,
   ...,
+  cols = NULL,
   by = NULL,
   error_call = current_env()
 ) {
-  check_dots_empty0(...)
   check_data_frame(data, call = error_call)
+
+  cols <- compat_chop_cols(cols = enquo(cols), ...)
+  by <- enquo(by)
 
   info <- chop_info(
     data,
-    cols = {{ cols }},
-    by = {{ by }},
+    cols = !!cols,
+    by = !!by,
     error_call = error_call
   )
   cols <- info$cols
@@ -236,6 +238,49 @@ col_chop <- function(x, indices) {
   out <- new_list_of(out, ptype)
 
   out
+}
+
+compat_chop_cols <- function(cols, ...) {
+  n_dots <- dots_n(...)
+
+  if (n_dots == 0L) {
+    return(cols)
+  }
+
+  # `env` and `user_env` here are fixed to always report `chop()` as `env`
+  # and the caller of `chop()` as `user_env`, regardless of the `error_call`
+  # argument. We think that makes the most sense for these errors/warnings.
+  env <- caller_env()
+  user_env <- caller_env(2)
+
+  if (n_dots != 1L) {
+    check_dots_empty0(..., call = env)
+  }
+
+  if (!quo_is_null(cols)) {
+    cli::cli_abort(
+      "Can't specify `cols` by both name and position.",
+      call = env
+    )
+  }
+
+  # Safe, we checked `n_dots == 1L` above
+  cols <- enquos(...)[[1L]]
+
+  lifecycle::deprecate_soft(
+    when = "1.4.0",
+    what = I(cli::format_inline(
+      "Specifying the {.arg cols} argument by position"
+    )),
+    details = cli::format_inline(
+      "Please explicitly name {.arg cols}, like {.code chop(data, cols = {as_label(cols)})}."
+    ),
+    env = env,
+    user_env = user_env,
+    id = "tidyr-chop-positional-cols"
+  )
+
+  cols
 }
 
 #' @export
--- a/R/chop.R
+++ b/R/chop.R
@@ -2,17 +2,16 @@
 #'
 #' @description
 #' Chopping and unchopping preserve the width of a data frame, changing its
-#' length. `chop()` makes `df` shorter by converting rows within each group
-#' into list-columns. `unchop()` makes `df` longer by expanding list-columns
+#' length. `chop()` makes `data` shorter by converting rows within each group
+#' into list-columns. `unchop()` makes `data` longer by expanding list-columns
 #' so that each element of the list-column gets its own row in the output.
+#'
 #' `chop()` and `unchop()` are building blocks for more complicated functions
-#' (like [unnest()], [unnest_longer()], and [unnest_wider()]) and are generally
-#' more suitable for programming than interactive data analysis.
+#' (like [unnest()], [unnest_longer()], and [unnest_wider()]).
 #'
 #' @details
-#' Generally, unchopping is more useful than chopping because it simplifies
-#' a complex data structure, and [nest()]ing is usually more appropriate
-#' than `chop()`ing since it better preserves the connections between
+#' When multiple columns are being chopped at once, [nest()] is usually more
+#' appropriate than `chop()` since it better preserves the connections between
 #' observations.
 #'
 #' `chop()` creates list-columns of class [vctrs::list_of()] to ensure
@@ -21,6 +20,25 @@
 #' the roundtrip chop and unchop. Because `<list_of>` keeps tracks of
 #' the type of its elements, `unchop()` is able to reconstitute the
 #' correct vector type even for empty list-columns.
+#'
+#' @section Connection to `split()`:
+#'
+#' `chop()` is the tidyverse version of [base::split()], with a few key changes:
+#'
+#' - The unique values of the columns used to chop by are preserved losslessly
+#'   as output columns, rather than being converted to character labels used as
+#'   names on the output list. This is particularly useful when chopping by
+#'   multiple columns.
+#'
+#' - Multiple columns can be chopped at once, producing one list-column per
+#'   chopped column. The closest `split()` equivalent is to split a data frame,
+#'   which produces a result more similar to [nest()] than `chop()`.
+#'
+#' - When chopping by multiple columns, the [interaction()]s of those columns
+#'   are not taken, and missing values are not dropped, which avoids a quadratic
+#'   runtime and tends to produce more expected results in practice.
+#'
+#' For an even lower-level version, see [vctrs::vec_split()].
 #'
 #' @inheritParams rlang::args_dots_empty
 #' @inheritParams rlang::args_error_context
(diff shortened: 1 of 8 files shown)


DavisVaughan:
Regarding the above snags:

> In the absence of both cols and by, should we 'chop everything' as nest() does? I have thrown an error instead.

I believe that is correct. I have kept this behavior.

> Should chop() use group vars if data is grouped, as nest() does? Would this require a generic?

I believe changing this behavior of `chop()` is out of scope for this PR. `chop()` currently does not try and respect existing grouping columns in any way (like, in theory it could chop by the grouping columns if no inputs are provided, but I don't think that feels that useful, I'd rather people be explicit).

> What should happen if a column is selected by both `cols` and `by`? For now, the name is duplicated and repaired.

This is now an informative error. It now operates the same way as `pivot_wider()` with `id_cols`, where the `id_cols` tidyselection is computed after removing `names_from` and `values_from` columns from the data. Similarly, `cols` is computed after removing any `by` cols from the data. This:

- Disallows `df |> chop(x, by = x)`, forcing an informative error
- Enables `df |> chop(everything(), by = x)` to be meaningful

I'd argue this aligns with `pivot_wider()` and with how `across()` works

R/chop.R
```diff
@@ -44,41 +70,73 @@
...
 #' # Unchopping a list-col of data frames must generate a df-col because
-#' # unchop leaves the column names unchanged
+#' # `unchop()` leaves the column names unchanged
 #' df <- tibble(x = 1:3, y = list(NULL, tibble(x = 1), tibble(y = 1:2)))
 #' df |> unchop(y)
 #' df |> unchop(y, keep_empty = TRUE)
-chop <- function(data, cols, ..., error_call = current_env()) {
+chop <- function(
+  data,
+  cols = NULL,
```
DavisVaughan: Perhaps in a perfect world I'd move `cols` after the empty `...` now that it default to `NULL`, but that would probably not be worth the effort

Let me know if this feels too awkward as is to you

tests/testthat/_snaps/pivot-wide.md
```diff
@@ -202,6 +202,24 @@
...
       i Column `value` has already been selected.
 
+---
+
+    Code
+      pivot_wider(df, id_cols = all_of(cols), names_from = name, values_from = value)
+    Condition
+      Error in `pivot_wider()`:
+      ! `id_cols` can't select a column already selected by `names_from`.
+      i Column `name` has already been selected.
```
DavisVaughan, line 212: A test for #1506

R/chop.R
```diff
@@ -44,41 +70,73 @@
...
+#' # using `by`
+#' df |> chop(by = c(x, y))
+#'
+#' # Compare to `split()`, notice how `x` and `y` are converted to character
+#' # labels
+#' df |> split(df[c("x", "y")], drop = TRUE)
+#'
+#' # `cols` and `by` can be used together to drop columns you no longer need.
+#' # This drops `y`:
+#' df |> chop(z, by = x)
```
DavisVaughan: I am particularly compelled by this use case of `by`, where it is no longer just syntax sugar, but also enables you to avoid a preemptive `select()` call. `nest()` has the same capability.

hadley:
@DavisVaughan anything in particular that you're looking for from my review?

R/chop.R
```diff
@@ -22,15 +21,42 @@
...
+#'
+#' - Multiple columns can be chopped at once, producing one list-column per
+#'   chopped column. The closest `split()` equivalent is to split a data frame,
+#'   which produces a result more similar to [nest()] than `chop()`.
+#'
+#' - When chopping by multiple columns, the [interaction()]s of those columns
+#'   are not taken, and missing values are not dropped, which avoids a quadratic
+#'   runtime and tends to produce more expected results in practice.
+#'
+#' For an even lower-level version, see [vctrs::vec_split()].
```
DavisVaughan, line 42: A read through of this would be nice

R/chop.R
```diff
@@ -94,6 +152,77 @@ chop <- function(data, cols, ..., error_call = current_env()) {
...
+    # Consistent with `pivot_wider(id_cols = )`.
+    try_fetch(
+      cols <- names(tidyselect::eval_select(
+        expr = cols,
+        data = data[setdiff(names, by)],
+        allow_rename = FALSE,
+        error_call = error_call
+      )),
+      vctrs_error_subscript_oob = function(cnd) {
+        maybe_throw_already_selected_error(
```
DavisVaughan, line 204: Do you generally agree with this? Same as `pivot_wider(id_cols` behavior where `names_from` and `values_from` are removed first (even uses the same error message)

DavisVaughan:
@hadley I have left 3 comment blocks open above. Just take a look at those, I'm confident in the implementation

R/chop.R
```diff
@@ -22,15 +21,42 @@
...
 #' correct vector type even for empty list-columns.
 #'
+#' @section Connection to `split()`:
+#'
+#' `chop()` is the tidyverse version of [base::split()], with a few key changes:
+#'
+#' - The unique values of the columns used to chop by are preserved losslessly
+#'   as output columns, rather than being converted to character labels used as
+#'   names on the output list. This is particularly useful when chopping by
+#'   multiple columns.
```
hadley:
```suggestion
#'   non-string columns or multiple columns.
```

R/chop.R
```diff
@@ -22,15 +21,42 @@
...
+#'
+#' - Multiple columns can be chopped at once, producing one list-column per
+#'   chopped column. The closest `split()` equivalent is to split a data frame,
+#'   which produces a result more similar to [nest()] than `chop()`.
+#'
+#' - When chopping by multiple columns, the [interaction()]s of those columns
```
hadley: Can you say what it does instead of what it doesn't do? 😄  Something like:

> When chopping by multiple columns, only the combinations present in the output are included in the output. This is different to `split()` which takes the [interaction()] of the columns, leading to a potential combinatorial explosion of output elements.

R/chop.R
```diff
@@ -22,15 +21,42 @@
...
+#'
+#' For an even lower-level version, see [vctrs::vec_split()].
+#'
 #' @inheritParams rlang::args_dots_empty
 #' @inheritParams rlang::args_error_context
 #'
 #' @param data A data frame.
 #' @param cols <[`tidy-select`][tidyr_tidy_select]> Columns to chop or unchop.
 #'
+#'   For `chop()`, columns specified by `by` are removed from `data` before
```
hadley: I wonder if it would make sense to document `@params cols,by` so you can more clearly delineate the three cases:

* cols only
* by only
* cols + by

R/chop.R
```diff
@@ -44,41 +70,73 @@
...
+#'
+#' # `chop()` is most useful as a tidyverse alternative to `base::split()`
+#'
+#' # Chop `z` by `x` and `y`. Note that we get one row of output for each unique
+#' # combination of non-chopped variables.
+#' df |> chop(z)
+#'
+#' # Equivalently, specify variables to chop by (rather than variables to chop)
+#' # using `by`
+#' df |> chop(by = c(x, y))
```
hadley, line 86: Maybe better to start with this since it's closer to `split()`?

R/chop.R
```diff
@@ -44,41 +70,73 @@
...
-#' df |> chop(c(y, z))
-#' # cf nest
-#' df |> nest(data = c(y, z))
+#' df <- tibble(x = c(1, 1, 1, 2, 2, 3), y = c(1, 1, 2, 3, 3, 4), z = 1:6)
+#'
+#' # `chop()` is most useful as a tidyverse alternative to `base::split()`
+#'
+#' # Chop `z` by `x` and `y`. Note that we get one row of output for each unique
+#' # combination of non-chopped variables.
+#' df |> chop(z)
```
hadley:
```suggestion
#' df |> chop(cols = z)
```

Since (I assume) we're putting `cols` and `by` on equal footing.

R/chop.R
hadley: I think it's worth considering? It's not too much effort to provide a deprecation message when `!missing(...1)`?

R/chop.R
hadley: I don't remember enough of the situation to have a meaningful opinion.

hadley (approved):
I really like filling in this little gap in the tidyverse ecosystem!

DavisVaughan:
Thanks for getting this started @hrryt!

PR #1632, opened 2026-06-05
Clarify crossing factor level behavior

### LINKED ISSUE #1526: `crossing()` adds missing factor levels (either a bug or a documentation issue)
When using `crossing()` with a factor that does not have all levels present, it inserts the missing levels.  My guess is that this is an intentional feature and the documentation should clarify it.  When I was using `crossing()` the expansion of missing levels was unexpected since it added values that were not in the combination of inputs.

The observed behavior is below, and I expected the second behavior in both cases.

I think that the documentation clarification that would have made the behavior expected would be the following:

Here, add "and the `nesting()` and `crossing()` helpers" after "`complete()`":

https://github.com/tidyverse/tidyr/blob/ad628413265c343bd7920b4be9b1ac35529741a9/R/expand.R#L36-L38

And, in the `expand_grid()` documentation, point back to the fact that it uses `expand()`.  Perhaps adding something like the following to the details section of the docs of `expand_grid()`:  "`expand_grid()` uses `expand()` to generate all combinations.

``` r
library(tidyr)

# factor levels are expanded
crossing(a = c("A", "B"), b = factor(c("a", "b"), levels = c("a", "b", "c")))
#> # A tibble: 6 × 2
#>   a     b    
#>   <chr> <fct>
#> 1 A     a    
#> 2 A     b    
#> 3 A     c    
#> 4 B     a    
#> 5 B     b    
#> 6 B     c
# factor levels are not expanded
crossing(
  a = c("A", "B"),
  data.frame(b = factor(c("a", "b"), levels = c("a", "b", "c")))
)
#> # A tibble: 4 × 2
#>   a     b    
#>   <chr> <fct>
#> 1 A     a    
#> 2 A     b    
#> 3 B     a    
#> 4 B     b
```

<sup>Created on 2023-10-28 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

If helpful, I can make the documentation PR.

LeonidasZhak:
Fixes #1526.

This clarifies that `crossing()` uses the full set of levels for factor inputs, not just the values present in the data. I kept the change scoped to `crossing()` because `nesting()` and `expand_grid()` do not independently expand factor levels in the same way.

Checks:
- Focused check confirming `crossing()` expands the missing factor level in the issue reprex
- `Rscript -e 'tools::checkRd("man/expand.Rd")'`\n- `git diff --check`\n- `R CMD build --no-build-vignettes .`\n- `R_LIBS_USER=/tmp/tune-r-lib _R_CHECK_FORCE_SUGGESTS_=false R CMD check --no-manual --ignore-vignettes --no-build-vignettes --no-tests tidyr_1.3.2.9000.tar.gz`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # tidyr (development version)
 
+* The `crossing()` documentation now clarifies that factor inputs use the full
+  set of levels, not just the values present in the data (#1526).
+
 * `chop()` gains a `by` argument for specifying grouping columns, similar to `nest(.by =)` (@hrryt, #1490).
 
 * Specifying `chop()`'s `cols` argument by position is soft-deprecated. It must instead be specified by name, which better communicates intent now that `chop()` also has a `by` argument (#1490).
--- a/R/expand.R
+++ b/R/expand.R
@@ -33,9 +33,9 @@
 #'     a row for each present school-student combination for all possible
 #'     dates.
 #'
-#'   When used with factors, [expand()] and [complete()] use the full set of
-#'   levels, not just those that appear in the data. If you want to use only the
-#'   values seen in the data, use `forcats::fct_drop()`.
+#'   When used with factors, [expand()], [complete()], and [crossing()] use the
+#'   full set of levels, not just those that appear in the data. If you want to
+#'   use only the values seen in the data, use `forcats::fct_drop()`.
 #'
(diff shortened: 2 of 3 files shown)


DavisVaughan:
Assuming this is the same as https://github.com/tidymodels/broom/pull/1281#issuecomment-4642851835

PR #1635, opened 2026-06-06
fix: full_seq() ignores infinite values (#1496)

### LINKED ISSUE #1496: `full_seq.numeric()` should set `finite = TRUE`
Right now you just get an error

```r
> tidyr::full_seq(c(1, Inf), 1L)
Error in if (any(((x - rng[1])%%period > tol) & (period - (x - rng[1])%%period >  : 
  missing value where TRUE/FALSE needed
```

You most likely don't want to create an infinite sequence anyways.

This would be somewhat useful with the Date method in particular, as a way to generate a full sequence along the following start/end intervals:

``` r
start <- as.Date(c("2019-01-05", "2019-01-10", "2019-01-11", "2019-01-14"))
end <- as.Date(c("2019-01-07", NA, "2019-01-14", NA))
end[is.na(end)] <- Inf

# `end = Inf` means "hasn't ended yet"
df <- tibble::tibble(start = start, end = end)
df
#> # A tibble: 4 × 2
#>   start      end       
#>   <date>     <date>    
#> 1 2019-01-05 2019-01-07
#> 2 2019-01-10 Inf       
#> 3 2019-01-11 2019-01-14
#> 4 2019-01-14 Inf

# Want a full sequence along range of start/end excluding infinite values
# `finite = TRUE` gives you the right range
range <- .Date(range(as.numeric(c(df$start, df$end)), finite = TRUE))
range
#> [1] "2019-01-05" "2019-01-14"

seq(range[1], range[2], by = 1)
#>  [1] "2019-01-05" "2019-01-06" "2019-01-07" "2019-01-08" "2019-01-09"
#>  [6] "2019-01-10" "2019-01-11" "2019-01-12" "2019-01-13" "2019-01-14"

# Ideally you'd just do:
# tidyr::full_seq(c(start, end), period = 1)
```

<sup>Created on 2023-04-28 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

LeonidasZhak:
## Summary

Fixes `full_seq()` to properly handle infinite values (`Inf` and `-Inf`) in the input vector.

## Problem

Currently, `full_seq()` fails with a cryptic error when the input contains infinite values:

```r
tidyr::full_seq(c(1, Inf), 1L)
#> Error in if (any(((x - rng[1])%%period > tol) & (period - (x - rng[1])%%period > : 
#>   missing value where TRUE/FALSE needed
```

This happens because `range()` returns `Inf`, and subsequent modulo operations produce `NaN`.

## Solution

Filter out infinite values before computing the range and checking periodicity. This is consistent with how `na.rm = TRUE` already handles `NA` values.

## Changes

- `R/seq.R`: Added filtering of infinite values before range computation (+4 lines)
- `tests/testthat/test-seq.R`: Added tests for infinite value handling (+14 lines)

## Validation

- All seq tests pass: `[ FAIL 0 | WARN 0 | SKIP 2 | PASS 12 ]`
- New tests verify Inf, -Inf, and combined infinite value handling
- Date method also works correctly with infinite values

## Related

- Fixes #1496
- Issue was filed by @DavisVaughan (tidyr maintainer)

### DIFF
--- a/R/seq.R
+++ b/R/seq.R
@@ -20,11 +20,15 @@ full_seq.numeric <- function(x, period, tol = 1e-6) {
   check_number_decimal(period)
   check_number_decimal(tol, min = 0)
 
-  rng <- range(x, na.rm = TRUE)
+  # Filter out infinite values before computing range
+  finite <- is.finite(x)
+  x_finite <- x[finite]
+
+  rng <- range(x_finite, na.rm = TRUE)
   if (
     any(
-      ((x - rng[1]) %% period > tol) &
-        (period - (x - rng[1]) %% period > tol)
+      ((x_finite - rng[1]) %% period > tol) &
+        (period - (x_finite - rng[1]) %% period > tol)
     )
   ) {
     cli::cli_abort("{.arg x} is not a regular sequence.")
--- a/tests/testthat/test-seq.R
+++ b/tests/testthat/test-seq.R
@@ -36,3 +36,18 @@ test_that("validates inputs", {
     full_seq(x, 1, tol = "a")
   })
 })
+
+test_that("full_seq ignores infinite values (#1496)", {
+  expect_equal(full_seq(c(1, Inf, 5), 1), 1:5)
+  expect_equal(full_seq(c(-Inf, 1, 5), 1), 1:5)
+  expect_equal(full_seq(c(-Inf, 1, 5, Inf), 1), 1:5)
+  expect_equal(full_seq(c(1, Inf), 1), 1)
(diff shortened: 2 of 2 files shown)


LeonidasZhak:
I am withdrawing this PR as part of a broader cleanup of an oversized automated contribution batch. I am sorry for the review noise and the pressure this may have put on maintainers. Going forward I will keep contributions to this project much more limited and higher-signal, ideally only one or two focused PRs at a time. Thank you for maintaining the project.

Issue #1588, opened 2025-01-29
separate_wider_position  & names_sep


hanguojun007:
Hi ,this is my example, i think it may be wrong.
when i set `names_sep = "_"` and `cols_remove = FALSE`，the `x` column become `x_x`.

```{r}
df4 <- tibble(x = c("202215TX", "202122LACC", "202325"))
df4 |>
  separate_wider_position(
    x,
    widths = c(year = 4, age = 2, state = 2),
    names_sep = "_",
    too_few = "align_start",
    too_many = "drop",
    cols_remove = FALSE
  )
```

# A tibble: 3 × 4
  x_year x_age x_state x_x       
  <chr>  <chr> <chr>   <chr>     
1 2022   15    TX      202215TX
2 2021   22    LA      202122LACC
3 2023   25    NA      202325

Edgar-Zamora:
I think this might the be the expected behavior. According to the [docs](https://tidyr.tidyverse.org/reference/separate_wider_delim.html) for the `names_sep` argument, it is a "composed of the input column name followed by the separator followed by the new column name". So in this instance it is `x` (input column) + `_` (separator) + `x` (new column name).

Issue #1499, opened 2023-05-14
`separate_wider_delim` renames original column when `col_remove=FALSE` and `names=` not specified


tszberkowitz:
When splitting a delimited character variable using the newer `separate_wider_delim()` function from the `tidyr` package (v 1.3.0), if you:
* specify the `names_sep=` argument,
* _do NOT_ specify the `names=` argument, and
* specify `cols_remove=FALSE`,

then the original variable is retained in the output data set (as expected) but:

1. the original variable name has been duplicated using the value specified in the `names_sep=` argument such that, e.g., `names_sep='_'` with `cols=varname` produces a variable named `varname_varname` in the output data, and
2. the original variable is located *after* the new separated columns, which is different from how the older `separate()` function behaves (placing the original column *before* the new columns).

Note that the first point above (variable renaming) is the major issue. The second point is just something that I was not unexpecting.

``` r
library(dplyr, warn.conflicts = FALSE)
library(tidyr)
library(reprex)

# Create test data set
## 1 character variable (`v`):
##  * semicolon-delimited values,
##  * includes NA,
##  * inconsistent/unpredictable number of delimiters per value
test <- tibble(
  v = c('a;b', 'c', NA, 'd;e;f', 'g;h')
)

# specifying `names` (not `names_sep`)
# `cols_remove` is TRUE => behaves as expected (original column name unchanged)
separate_wider_delim(
  data = test,
  cols = v,
  delim = ';',
  names = c('v_1', 'v_2', 'v_3'),
  too_few = 'align_start',
  cols_remove = FALSE
)
#> # A tibble: 5 × 4
#>   v_1   v_2   v_3   v    
#>   <chr> <chr> <chr> <chr>
#> 1 a     b     <NA>  a;b  
#> 2 c     <NA>  <NA>  c    
#> 3 <NA>  <NA>  <NA>  <NA> 
#> 4 d     e     f     d;e;f
#> 5 g     h     <NA>  g;h

# specifying `names_sep` only
# `cols_remove` is TRUE (default) => behaves as expected
separate_wider_delim(
  data = test,
  cols = v,
  delim = ';',
  names_sep = '_',
  too_few = 'align_start',
  cols_remove = TRUE
)
#> # A tibble: 5 × 3
#>   v_1   v_2   v_3  
#>   <chr> <chr> <chr>
#> 1 a     b     <NA> 
#> 2 c     <NA>  <NA> 
#> 3 <NA>  <NA>  <NA> 
[...]

PR #1637, opened 2026-06-06
fix: preserve original column name with names_sep and cols_remove = FALSE (#1499, #1588)

LeonidasZhak:
## Problem

When using `names_sep` with `cols_remove = FALSE` in `separate_wider_delim()`, `separate_wider_position()`, and `separate_wider_regex()`, the original column was incorrectly renamed. For example, a column named `x` would become `x_x` instead of remaining `x`.

This affected both:
- Issue #1499: `separate_wider_delim()` renames original column
- Issue #1588: `separate_wider_position()` & `names_sep`

## Root Cause

The `unpack()` function applies `names_sep` to rename ALL columns in the packed data frame, including the original column that was preserved with `cols_remove = FALSE`.

## Solution

Modified `map_unpack()` to detect when the original column is present in the packed data and restore it after unpacking:

1. Before unpacking, detect if any packed column contains a sub-column with the same name as the outer column
2. Save the original column data
3. After unpacking, remove the incorrectly renamed column (e.g., `x_x`) and restore the original column with its correct name

## Changes

- `R/separate-wider.R`: Modified `map_unpack()` to preserve original column name (+34 lines)
- `tests/testthat/test-separate-wider.R`: Added 3 tests for the fix (+58 lines)

## Validation

- All existing tests pass: `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 1337 ]`
- New tests verify the fix for all three `separate_wider_*()` functions

## Reprex

```r
library(tidyr)
library(tibble)

# Issue #1499: separate_wider_delim()
test <- tibble(v = c("a;b", "c", NA, "d;e;f", "g;h"))
separate_wider_delim(test, v, ";", names_sep = "_", too_few = "align_start", cols_remove = FALSE)
#> # A tibble: 5 × 4
#>   v_1   v_2   v_3   v
#>   <chr> <chr> <chr> <chr>
#> 1 a     b     <NA>  a;b
#> 2 c     <NA>  <NA>  c
#> 3 <NA>  <NA>  <NA>  <NA>
#> 4 d     e     f     d;e;f
#> 5 g     h     <NA>  g;h

# Issue #1588: separate_wider_position()
df <- tibble(x = c("202215TX", "202122LACC", "202325"))
separate_wider_position(df, x, c(year = 4, age = 2, state = 2), names_sep = "_", too_few = "align_start", too_many = "drop", cols_remove = FALSE)
#> # A tibble: 3 × 4
#>   x_year x_age x_state x
#>   <chr>  <chr> <chr>   <chr>
#> 1 2022   15    TX      202215TX
#> 2 2021   22    LA      202122LACC
#> 3 2023   25    <NA>    202325
```

### DIFF
--- a/R/complete.R
+++ b/R/complete.R
@@ -94,6 +94,24 @@ complete.data.frame <- function(data, ..., fill = list(), explicit = TRUE) {
 
 #' @export
 complete.grouped_df <- function(data, ..., fill = list(), explicit = TRUE) {
+  group_cols <- dplyr::group_vars(data)
+
+  # Check if any completion columns are grouping columns
+  dots <- enquos(...)
+  for (i in seq_along(dots)) {
+    dot_expr <- get_expr(dots[[i]])
+    if (is_symbol(dot_expr) && as_string(dot_expr) %in% group_cols) {
+      cli::cli_abort(
+        c(
+          "Can't complete on a grouping column.",
+          i = "Column {.val {as_string(dot_expr)}} is a grouping variable.",
+          i = "Use {.code dplyr::ungroup()} first, or complete on non-grouping columns."
+        ),
+        call = caller_env()
+      )
+    }
+  }
+
   out <- dplyr::reframe(
     data,
     complete(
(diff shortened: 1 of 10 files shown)


LeonidasZhak:
I am withdrawing this PR as part of a broader cleanup of an oversized automated contribution batch. I am sorry for the review noise and the pressure this may have put on maintainers. Going forward I will keep contributions to this project much more limited and higher-signal, ideally only one or two focused PRs at a time. Thank you for maintaining the project.

PR #1640, opened 2026-06-07
tests: add factor edge-case tests for extract()

LeonidasZhak:
## What

Add edge-case tests for `extract()` covering factor column inputs:
- Factor columns are converted to character before extraction
- NA values in factor columns are preserved
- Factor input works with `convert = TRUE`
- Factor columns with unused levels extract correctly

## Why

`extract()` converts factor columns via `as.character()` internally (line 71 of `R/extract.R`), but this behavior had no test coverage. These edge cases are important for data cleaning pipelines where factor columns are common.

## Testing

- All 4 new tests pass (22 total expectations, up from 11)
- No regressions in existing tests
- `devtools::test(filter = 'extract')` passes cleanly (pre-existing internal function test skipped as expected)

## Files changed

- `tests/testthat/test-extract.R` — added 35 lines of factor edge-case tests

### DIFF
--- a/tests/testthat/test-extract.R
+++ b/tests/testthat/test-extract.R
@@ -102,3 +102,38 @@ test_that("validates its inputs", {
     df |> extract(x, into = "x", convert = 1)
   })
 })
+
+test_that("extract converts factor columns to character", {
+  df <- tibble(x = factor(c("a-b", "c-d"), levels = c("a-b", "c-d", "e-f")))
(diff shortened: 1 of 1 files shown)


LeonidasZhak:
Withdrawing this small automated PR while I consolidate an oversized batch of contributions and reduce maintainer review burden. Sorry for the noise, and thank you for maintaining the project.

